PR #5, opened 2015-02-07
makepanda.py --installer fixes for dpkg confusion on Fedora and python3

JonasT:
The commit messages should describe nicely what I've changed.


JonasT:
It is done! Unless you tell me it isn't, in which case I will find ways to refine. (:

PR #11, opened 2015-03-02
Fix broken animations when multipart actor have completely overlapping parts

deflected:
Revert "Patch by cfsworks: actor: On removePart, clean up the part's commonBundleHandle."

This reverts commit a4563fb342715bccd9a63aa320395f13abaace46.

commonBundleHandle is not holding bundle for this part only. When
two or more parts shares or have similar bundles - their bundles
are merged. In these cases it is not correct to remove
the commonBundleHandle for this parts.

### DIFF
--- a/direct/src/actor/Actor.py
+++ b/direct/src/actor/Actor.py
@@ -1019,11 +1019,6 @@ class Actor(DirectObject, NodePath):
         if (partName in partDict):
             del(partDict[partName])
 
-        # remove the bundle handle, in case this part is ever
-        # loaded again in the future
-        if partName in self.__commonBundleHandles:
-            del self.__commonBundleHandles[partName]
-
     def hidePart(self, partName, lodName="lodRoot"):
         """
         Make the given part of the optionally given lod not render,


rdb:
@cfsworks comments?

CFSworks:
@rdb The original change was so that calling loadModel in the future to load a new version of that part would properly use the new bundle as well, as without it I was getting animation bugs (the old bundle was animating the new mesh and it looked unpleasant).

I'm not entirely sure what @deflected is saying here; a revert will most definitely result in a regression of the bug I was seeing, but it looks like the problem arises when you load two parts, their bundles get merged [(as seen here)](https://github.com/deflected/panda3d/blob/d74d6cb58774a536d75ff2c41303d1e8d6279d92/direct/src/actor/Actor.py#L1987) and you try to delete one of them -- the proper behavior in this case, I believe, is to unmerge the bundles rather than delete them.

That said the Panda3D skeletal animation system is just about the one thing I'm the **least** familiar with, so take my interpretation with a grain of salt. :)

deflected:
@CFSworks: The commonBundle merges all bundles for given part on all LODs. Removing 1 part from specific LOD will remove bundleHandle for all LODs ... this is not correct.
Maybe the proper way for cleaning and deleting this is when all models for given part from all LODs are removed. Or just to re-do the commonBundleHandle (like postFlatten do it).

CFSworks:
@deflected I would agree with either, since we need to empty the commonBundleHandles entry when the last LOD model is gone, and either of those (rebuilding on every model unload or purging on the last LOD getting unloaded) will achieve that.

deflected:
@CFSworks @rdb Here is an example for the problem. It uses very similar and basic logic to my understanding of what parts are. https://drive.google.com/file/d/0B5oSLdFG_y2cTUlsa2RhOEUwUkE/view
I still think that cleaning of commonBundleHandle this way is wrong and creates more problems than it solves.
@CFSworks could you please send me an example with your case?

rdb:
Since I think it's better to err on the side of "no regressions", I'm tempted to accept this PR for the time being (reverting the original fix by @CFSworks ) until a solution is found that fixes both issues.  Any objections?

CFSworks:
@rdb Since either way this is a bug with the mergeLODBundles behavior, the default value for merge-lod-bundles should probably be switched to false until then.

deflected:
@CFSworks You can use Actor(mergeLODBundles=False) if you have problems with this. Maybe you have a regression, because you have different hierarchy of your bundles in the new models.

deflected:
@rdb: Do you have intention to merge the change?

rdb:
Yes, I will merge this, since I'm going to release 1.9.0 very soon, and I'd prefer no regressions.  It'd still be great if a permanent fix could be found, but we can deal with that after 1.9.0.

PR #13, opened 2015-03-16
Consolidate distance factor for 3D audio

nikolmiv:
The semantics of distance factor is now same for both OpenAL and Fmod.
The value is given in units per meter, and the default is 1.0.
This means that Panda units are understood to be meters by default.
For example, for feet one would have to set this value to 3.28.

### DIFF
--- a/panda/src/audiotraits/openalAudioSound.cxx
+++ b/panda/src/audiotraits/openalAudioSound.cxx
@@ -53,7 +53,7 @@ OpenALAudioSound(OpenALAudioManager* manager,
   _balance(0),
   _play_rate(1.0),
   _positional(positional),
-  _min_dist(3.28f),
+  _min_dist(1.0f),
   _max_dist(1000000000.0f),
   _drop_off_factor(1.0f),
   _length(0.0),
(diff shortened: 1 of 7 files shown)


panda/src/audiotraits/openalAudioSound.cxx
```diff
@@ -745,7 +745,7 @@ set_3d_min_distance(PN_stdfloat dist) {
     _manager->make_current();
 
     alGetError(); // clear errors
-    alSourcef(_source,AL_REFERENCE_DISTANCE,_min_dist*_manager->audio_3d_get_distance_factor());
+    alSourcef(_source,AL_REFERENCE_DISTANCE,_min_dist);
```
rdb: Hmm, why shouldn't this be scaled by the distance factor?

nikolmiv: The way I understand, only the speed of sound needs to conform to whatever units are assumed. So if min/max distances (and source locations) are assumed to be e.g. in feet, and the speed of sound is scaled to feet, then the sound will travel distances in proper time.

The other way around would be not to multiply the speed of sound, but to divide every distance and source location by the distance factor.

PR #14, opened 2015-03-17
Fix MeshDrawer::stream

nikolmiv:
Direction vector is computed as stop - start, but the position
in the loop is computed from stop as base, instead of from start;
this is changed to start as the base (and then offset must be added).
Multiplication of non-normalized direction vector with its length
yields square of the expected current step inside the loop;
this is changed to use the normalized direction vector.

### DIFF
--- a/panda/src/grutil/meshDrawer.cxx
+++ b/panda/src/grutil/meshDrawer.cxx
@@ -369,8 +369,9 @@ void MeshDrawer::stream(const LVector3 &start, const LVector3 &stop, const LVect
   LVector3 relative_pos = stop;
   LVector3 vec = stop - start;
   PN_stdfloat distance = vec.length();
+  vec.normalize();
   for(int i = 0; i < number; i++) {
-    relative_pos = stop + vec * ((i-offset)*(distance/PN_stdfloat(number)));
+    relative_pos = start + vec * ((i+offset)*(distance/PN_stdfloat(number)));
     billboard(relative_pos,frame,size,_color);
   }
 }


nikolmiv:
Oops, now I see there was a failure in the CI build. But it was actually killed by the system while building the installer, because the build took too long.

rdb:
Yeah, don't worry about the CI - it does that sometimes.  The important part is that there were no compile errors. :-)

Hmm, I'm afraid I don't have any code lying around to test MeshDrawer - I assume you've done the necessary testing in order to confirm that the old behaviour is demonstrably broken and the new behaviour fixes this issue?

In how far might this change break behaviour for people who are using MeshDrawer::stream()?

nikolmiv:
It doesn't actually draw a line of particles from start to stop, but it starts from stop and goes backwards distance-squared beyond start. I don't think anyone could have intended such behavior, so I conclude it is broken. With the fix, I checked that it draws exactly from start to stop (plus offset).

This fix will change very much the behavior for those who use it. But given how obviously wrong it behaves, I guess it wasn't really used.

rdb:
Whoops, it turns out there was already a bug report (with patch) about this:
https://bugs.launchpad.net/panda3d/+bug/730255

Fair enough -- enough grounds to accept your PR.  Thank you :-)

nikolmiv:
The patch in that bug report does fix the basic problem, but it counts the offset from the stop point, and not from the start point. E.g. in case of zero offset, it will put a particle at stop point and won't put a particle at start point. In my fix it is the opposite, which I think is less surprising behavior.

PR #16, opened 2015-04-17
 Mouse grab support (Windows / X11 only)

eswartz:
Add support for grabbing the mouse using WindowProperties#set_mouse_grabbed().
    
Implementation only for Windows and X11 so far.


eswartz:
Reimplementing as a mouse mode.

PR #17, opened 2015-04-17
Support M_confine mode (Windows / X11 only so far)

eswartz:
Different take on the earlier patch, making use of a third mouse mode.

This moves the X11 absolute/relative handling from mouse_mode_...() methods into set_properties_now.


eswartz:
Wrong branch...

PR #19, opened 2015-04-23
Support mouse confinement (Cocoa)

eswartz:
As for the findings yesterday about the bad behavior in the mouse-mode demo -- it seems that using #set_pointer() will result in a long delay before an absolute mouse position is updated, even though deltas are reported.

Thus, in CocoaPandaView#, the mouse deltas are sent for both relative and confined modes, so client code will be able to recenter the mouse without choppy movement.


eswartz:
Argh, this keeps getting mixed up with some commits I have been trying to delete for ages.  Closing again.

PR #20, opened 2015-04-30
Implement confined mouse mode for OS X.

eswartz:
Note: it seems that using #set_pointer() will result in
a long delay before an absolute mouse position is updated,
even though deltas are reported.

Thus, in CocoaPandaView, the mouse deltas are sent for both
relative and confined modes, so client code (e.g. the "mouse
modes" demo) will be able to recenter the mouse without
choppy movement.


rdb:
Whoops, not sure how I forgot to merge this.  Many thanks!

PR #21, opened 2015-05-01
Add -L (lighting) and -d <display> options to pview

eswartz:
In my workflow, I always need lighting (so textures show up) and prefer to use p3tinydisplay for faster startup.  This patch allows passing options to enable lighting and select the display.

### DIFF
--- a/doc/man/pview.1
+++ b/doc/man/pview.1
@@ -1,4 +1,4 @@
-.TH PVIEW 1 "27 December 2014" "" Panda3D
+.TH PVIEW 1 "1 May 2015" "" Panda3D
 .SH NAME
 pview \- quickly view a Panda model and/or animation
 .SH SYNOPSIS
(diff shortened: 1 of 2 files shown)


rdb:
This looks like a good addition.  But rather than hack the pipe selection using load_prc_file_data, I'd recommend instead using GraphicsPipeSelection::make_module_pipe("pandagl") to create the graphics pipe, and passing that to framework.open_window(pipe, NULL).  Passing NULL for pipe will make it use the default pipe.

I'd also prefer using "-P" for "graphics pipe" instead of "-d" for display (let's reserve lower-case "-p" for per-pixel in case we ever want to add that), since the first thing that came to mind was that "display" would set the display on which the window would open, but that could just be a personal thing.

Please also update the help text accordingly, and the man page in doc/man/pview.1 .

eswartz:
Wow, it took me a long time to realize I needed to use PointerTo<GraphicsPipe> instead of a plain pointer...   updated.

rdb:
LGTM, thank you. :-) FWIW, we tend to use PT(GraphicsPipe), which is a short-hand for PointerTo<GraphicsPipe>, but it doesn't really matter.

PR #22, opened 2015-05-02
Misc. fixes

CFSworks:
Fixes for a few problems that I uncovered during the last `cmake` branch update.


rdb:
Thanks!  I've cherry-picked these onto release/1.9.x and merged that back into master.

PR #24, opened 2015-05-06
Add sample showing some Panda3D uses of libRocket.

eswartz:
N.B.  This still uses the "from _rocketcore import rocket" hack.  

I noticed that 3rd-party Rocket files tend to have embedded Python in &lt;script&gt;> elements in *.rml, which expects "import rocket" to work.  And that doesn't work at all in Panda3D now, as we discussed, so both Panda3D *.py and *.rml files need the same hack.  

Thus, whatever fixes are made should probably include ensuring "import rocket" will work whether or not Panda3D's rocket module is loaded.


eswartz:
broken rendering on Windows -- working on it.

eswartz:
Should work now.  Building/testing on OS X and that should take a while.

eswartz:
OK, rebased agianst the model-path lookups.  Thanks!

rdb:
Oops; in my mind I'd already merged this, but evidently not. Thanks again for contributing this sample!

PR #28, opened 2015-05-11
Dynamically adjust camera clip range in pview

eswartz:


rdb:
This was merged, but GitHub didn't register it.

PR #29, opened 2015-05-15
Fix invalid assumptions in GravityWalker

jjkoletar:
The setCollisionRayHeight method assumed that the 0th child on the
avatar would be the av's GeomNode, and again assumed that the 2nd child of
the GeomNode would be this Walker's cRayNode. This is not always the case,
as scene graph manipulations can change the parenting order of nodes.

The fix is to simply pull the cRayNode from the Walker's own cRayNodePath
handle.

### DIFF
--- a/direct/src/controls/GravityWalker.py
+++ b/direct/src/controls/GravityWalker.py
@@ -697,8 +697,7 @@ class GravityWalker(DirectObject.DirectObject):
     # There are sometimes issues if the collision ray height is
     # so tall that it collides with multiple levels of floors.
     def setCollisionRayHeight(self, height):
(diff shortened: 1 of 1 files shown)


rdb:
I think what the code should be doing is storing the CollisionRay in a member variable and calling setOrigin on it instead of creating a new collision ray.

If you could do that, that would be great.  If you would, please make the change on the tip of "release/1.9.x" branch (and file it for merging into that) instead of master, so that it can be in 1.9.1.

If that's inconvenient, you can make it against master too, and I'll just cherry-pick the change.

jjkoletar:
redoing PR

PR #30, opened 2015-05-18
change drive of largeblobsenderconsts

encryptable:
probably wont get merged but it's worth a shot


rdb:
Hmm, looking at the code in question, it seems more clear that this code doesn't belong in Panda3D and should be removed.

PR #31, opened 2015-05-31
Make more keys available to libRocket event handlers

eswartz:
Some ASCII chars fell through and generate no events when modifier keys are held down.  E.g. pressing "Ctrl+\" emits neither a 'keydown' nor a 'textinput' event for libRocket.  

libRocket calls several punctuation keys "OEM" in its KeyIdentifier enum, so I assume those were overlooked.  This patch fills those in from Panda's side, assuming a US English keyboard, like libRocket's comments say.  Better that than losing the events entirely...?


eswartz:
Bah, when I tested this on OS X, I see those key events *are* getting through.  That's odd.  I am about to test on Windows too...

eswartz:
OK, my patch works as expected on Windows.  Yet more weirdness only on OS X.  :-p

rdb:
Great, thanks!

PR #33, opened 2015-06-01
Add PNMImage premultiply/unpremultiply methods.

eswartz:
Ought to be straightforward, except perhaps the naming of the methods.  Feel free to change them.

(I fought so hard thinking blend_sub_image() or blend() had some bug, and was making all sorts of elaborate fixes and extensions for that, but finally realized it was all my fault, hence these methods :-p.)

### DIFF
--- a/panda/src/pnmimage/pnmImage.cxx
+++ b/panda/src/pnmimage/pnmImage.cxx
@@ -623,6 +623,63 @@ make_grayscale(float rc, float gc, float bc) {
 }
 
 ////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


rdb:
Looks useful.  I'd prefer the names premultiply_alpha and unpremultiply_alpha to be more consistent with other operations, though (but I might be convinced otherwise if you have a compelling reason).

eswartz:
I agree.  My names were inspired by a lack of sleep.  I updated them to your suggestions and clarified the comments.

rdb:
Thank you!

PR #34, opened 2015-06-01
glstuff: Fix incorectly handle of Transpose in Shader Inputs

deflected:
- Added missing recalculation of size when
	input buffer is "resized" (due to removing of
	post/pre-fixes). Now the "Transponse" part is
	correctly handled.

Signed-off-by: deflected <deflected@github>


rdb:
Committed, with a very minor tweak.  Thank you!

PR #35, opened 2015-06-02
Power-of-two texture fixes

eswartz:
(leaving out the FilterManager changes for now)

### DIFF
--- a/panda/src/rocket/rocketRenderInterface.cxx
+++ b/panda/src/rocket/rocketRenderInterface.cxx
@@ -232,6 +232,7 @@ LoadTexture(Rocket::Core::TextureHandle& texture_handle,
 
   tex->set_minfilter(SamplerState::FT_nearest);
   tex->set_magfilter(SamplerState::FT_nearest);
+  tex->rescale_texture();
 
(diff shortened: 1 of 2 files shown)


rdb:
Thanks for your work on this!  Besides your tinydisplay message, I've just checked in two changes that should address the texture scaling issues.

Firstly, due to a regression in 1.9.0, textures weren't being scaled when loaded from disk.  This affected all texture loads, I believe.  This was fixed in ecaffd5.

Having fixed that, the libRocket renderer responded inconsistently to rescaled textures.  Since it makes layout decisions based on the reported image size, we should really be reporting the original texture size, so that it behaves consistently regardless of the `textures-power-2` setting.  It also now properly implements padding by scaling the texture coordinates.  I've also added logic to prefer padding over scaling, so that pixel-perfectness is maintained.  This is all in f8ec2c1.

I've tested the sample and it seems to work in every configuration now (except when I set `textures-power-2 none` and use tinydisplay, of course, in which case it fails with your error message).

Thanks again!  We probably would have overlooked this grave regression in 1.9.0 had you not stumbled over it.

PR #36, opened 2015-06-03
Changes to makepanda to compile with Assimp

Moguri:
This includes updating header files for Assimp 3.

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -87,7 +87,7 @@ PkgListSet(["PYTHON", "DIRECT",                        # Python support
   "ODE", "PHYSX", "BULLET", "PANDAPHYSICS",            # Physics
   "SPEEDTREE",                                         # SpeedTree
   "ZLIB", "PNG", "JPEG", "TIFF", "SQUISH", "FREETYPE", # 2D Formats support
-  ] + MAYAVERSIONS + MAXVERSIONS + [ "FCOLLADA",       # 3D Formats support
+  ] + MAYAVERSIONS + MAXVERSIONS + [ "FCOLLADA", "ASSIMP", # 3D Formats support
   "VRPN", "OPENSSL",                                   # Transport
   "FFTW",                                              # Algorithm helpers
   "ARTOOLKIT", "OPENCV", "DIRECTCAM", "VISION",        # Augmented Reality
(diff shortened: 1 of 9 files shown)


Moguri:
This has only been tested with one COLLADA file.

makepanda/makepanda.py
```diff
@@ -737,6 +737,7 @@ def parseopts(args):
         SmartPkgEnable("EIGEN",     "eigen3",    (), ("Eigen/Dense",), target_pkg = 'ALWAYS')
         SmartPkgEnable("ARTOOLKIT", "",          ("AR"), "AR/ar.h")
         SmartPkgEnable("FCOLLADA",  "",          ChooseLib(fcollada_libs, "FCOLLADA"), ("FCollada", "FCollada/FCollada.h"))
+        SmartPkgEnable("ASSIMP",    "libassimp", ("assimp"), "assimp")
```
rdb: I think this should be "assimp", as there is an assimp.pc, not a libassimp.pc (based on a quick google for either term).  If there is no .pc file shipped at all, this should be the empty string.

makepanda/makepanda.py
```diff
@@ -5208,6 +5209,16 @@ def CreatePandaVersionFiles():
   TargetAdd('libp3daeegg.lib', opts=['FCOLLADA', 'CARBON'])
 
 #
+# DIRECTORY: pandatool/src/assimp
+#
+if (PkgSkip("PANDATOOL") == 0 and PkgSkip("ASSIMP")==0):
+  OPTS=['DIR:pandatool/src/assimp', 'ASSIMP']
```
rdb: Since this is a plug-in, please add 'MODULE' to OPTS in order to prevent it from appending a soname suffix such as libp3assimp.so.1.9.0

rdb:
You should also add an appropriate LibName line for assimp in the COMPILER=="MSVC" case, or add a PkgDisable("ASSIMP") to that case until someone builds assimp on Windows and adds the correct lines.

makepanda/makepanda.py
Moguri: I believe you are correct, I have the following from my Assimp package:
/usr/lib/pkgconfig/assimp.pc

PR #39, opened 2015-06-15
Propagate OS X window size change

eswartz:


mfwass:
:-1: 

This can make some windows *super* off scaled.

eswartz:
Could you provide a sample test case or suggestion for doing it correctly?

mfwass:
For projects that manually control the display size and scale, it can really mess it up:



I do not have any suggestions on how to fix it, maybe @rdb does.

eswartz:
I see, sort of.  Do you have an example of the actual code involved here?  E.g.  that sets up WindowProperties, calls requestProperties(), etc?  Also, is this program using custom DisplayRegions?  

And what do you mean by "changing the display scale"?  It looks like the same screen resolution in both cases.  

(Without this information, we can only guess what might be going wrong, or whether a fix will solve your bug.)

FWIW, I submitted the PR because code like the following, to change an existing window's size, resulted in all the content not being resized to fit:

```
            props = WindowProperties()
            props.setFixedSize(True)

            sz = options.resolution.value
            props.setSize(sz[0], sz[1])

            if options.windowed.value:
                # center the window -- the position doesn't seem to be set properly
                x, y = self._system.fetchCurrentResolution()
                x = (x - sz[0]) / 2
                y = (y - sz[1]) / 2
                props.setOrigin(x, y)

            props.setFullscreen(not options.windowed.value)
            props.setUndecorated(not options.windowed.value)
            self.win.requestProperties(props)
```

I verified against the latest 1.9 branch, reverting my change, and the behavior still occurs.  This is on 10.8, btw.

So, your program is probably exposing another case that needs to be addressed.  But we need to know how it's doing it :)

mfwass:
I tried out the build on a separate program, seems to work fine. This may be a bug specific to the project I am working on, disregard all I said above.

mgracer48:
Legend of pirates online?  Oh how I'm excited for play that game oh I hope u release soon

rdb:
I looked into it and it did seem to leave things off scale, since calling set_size_and_recalc bypasses the mechanism that sends the window-event (which ShowBase uses for determining the aspect ratio).

A better fix seems to be not to change _properties, but instead call handle_resize_event(), which will call system_changed_properties(), which will throw both the window-event and set _properties to the changed size.  I've checked this in.

mfwass:
@rdb I'll let you know after I compile Panda again.

PR #40, opened 2015-06-17
Add back default model loading

DenialMC:
How does this bother anybody? This is a very useful feature.. especially when you already have a project with tons of DirectDialogs that you haven't set a model for.

Why was this removed?

### DIFF
--- a/direct/src/gui/DirectDialog.py
+++ b/direct/src/gui/DirectDialog.py
@@ -94,7 +94,7 @@ class DirectDialog(DirectFrame):
             ('text',              '',            None),
             ('text_align',        TextNode.ALeft,   None),
             ('text_scale',        0.06,          None),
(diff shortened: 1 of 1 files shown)


rdb:
This change was submitted by Thaumaturge, who (rightly) felt that DirectDialog should not load an ugly texture and have external dependencies by default.

But you're right, it would be preferable if the default dialog geom setting would still function.  We should instead default the default dialog geom to None rather than circumvent this mechanism entirely.  We might also need to have a setting for default relief, or something, then, or automatically use a relief in absence of a default geom.

rdb:
Sorry for taking a while to get to this.  I've just checked in a fix that allows setting the default geom, using something like:

    DGG.setDefaultDialogGeom(loader.loadModel('models/gui/dialog_box_gui')

Thanks!

PR #43, opened 2015-06-30
makepanda.py: added more refs for PackageMaker

mfwass:
PackageMaker isn't as easy to find and download anymore, whatever versions online that are available for download sometimes install the program into Applications/ folder or Developer/Tools/ with the name PackageMaker.app.

This pull request adds two new file paths to check for PackageMaker.app.


mfwass:
Thanks :)

PR #48, opened 2015-08-11
Add 'openal-device' configuration variable for selecting OpenAL device

eswartz:


panda/src/audiotraits/openalAudioManager.cxx
```diff
@@ -117,10 +118,11 @@ OpenALAudioManager() {
 
   // Initialization
   if (_active_managers == 0 || !_openal_active) {
-    _device = alcOpenDevice(NULL); // select the "preferred device"
+    string dev_name = get_audio_device();
+    _device = alcOpenDevice(dev_name.c_str()); // select the user or preferred device
```
rdb: Instead of `NULL` you're now passing the empty string if the "default device" is selected.

panda/src/audiotraits/openalAudioManager.cxx
```diff
@@ -213,6 +215,65 @@ is_valid() {
...
+  } else if (alcIsExtensionPresent(NULL, "ALC_ENUMERATION_EXT") == AL_TRUE) {
+    audio_cat->debug() << "Using ALC_ENUMERATION_EXT" << endl;
+    deviceList = (const char*) alcGetString(NULL, ALC_DEVICE_SPECIFIER);
+  }
+
+  if (deviceList) {
+    while (*deviceList) {
+      string dev = deviceList;
+      devList.push_back(dev);
+      audio_cat->debug() << "\t" << dev << endl;
```
rdb: Let's not use tab characters in the command-line output.

panda/src/audiotraits/config_openalAudio.h
```diff
@@ -27,4 +27,6 @@ NotifyCategoryDecl(openalAudio, EXPCL_OPENAL_AUDIO, EXPTP_OPENAL_AUDIO);
 extern EXPCL_OPENAL_AUDIO void init_libOpenALAudio();
 extern "C" EXPCL_OPENAL_AUDIO Create_AudioManager_proc *get_audio_manager_func_openal_audio();
 
+extern ConfigVariableString device_cfg;
```
rdb: Config variables should be named consistently, in this case `openal_device` instead of `device_cfg`.

panda/src/audiotraits/openalAudioManager.cxx
eswartz: Yes, this works fine in practice.

panda/src/audiotraits/config_openalAudio.h
eswartz: Will fix.

panda/src/audiotraits/openalAudioManager.cxx
eswartz: Will fix.

panda/src/audiotraits/openalAudioManager.cxx
rdb: I'm sure it does.  However, there is more than one OpenAL implementation out there, and the OpenAL specification states that NULL should be passed to get the default device.  I'd prefer not risking a change in default behaviour in a minor release of Panda3D.

PR #51, opened 2015-09-17
Current directory is not the root of the panda tree. ( Mac Installing )

CommanderBanana:
So I'm trying to set up Panda3D and I'm confused on what to do. I've tried to get the exact directory and do the command and terminal and all I come up is with this:

Elapsed Time: 0 sec
Current directory is not the root of the panda tree.
Build terminated.

Could anyone help me ( Also, I'm new here so I may be confused on what to do)


rdb:
You have to use `cd` to change directory to the root of the Panda3D source tree, and run makepanda from there.

I do wish to defer you to the forums for asking questions, rather than using the pull request system, which is meant for merging changes between development branches.  I assume you are not actually intending for us to merge all development changes into the stable branch.

CommanderBanana:
No, Sorry, I apologize, just a bit new here and confused on how to work github exactly. But thank you for the advice.

PR #52, opened 2015-10-06
Assimp: Add support for skinned meshes (no animations yet).

Moguri:
Here's a first attempt at skinned meshes via Assimp. Some things to note:

  * I haven't checked for new memory leaks
  * The use of const char* as a hash key seems kind of ugly, but it also seems like the simplest approach.
  * Some NULL checks might be good. An aiBone should have a matching aiNode, but it might be good to protect ourselves if Assimp misbehaves.
  * The tbtable->set_rows() could be optimized to leave out non-animated verts

GitHub doesn't like archive files, so I've uploaded one to Dropbox: https://dl.dropboxusercontent.com/u/3431679/skel_mesh.zip

Inside you'll find:

  * animbox.blend -- The original blend file
  * animbox.dae -- The COLLADA file exported from Blender using the Better COLLADA plugin
  * main.py -- The script to test with

### DIFF
--- a/pandatool/src/assimp/assimpLoader.cxx
+++ b/pandatool/src/assimp/assimpLoader.cxx
@@ -31,11 +31,23 @@
 #include "pointLight.h"
 #include "look_at.h"
 #include "texturePool.h"
+#include "character.h"
+#include "pvector.h"
 
 #include "pandaIOSystem.h"
 #include "pandaLogger.h"
 
 #include "assimp/postprocess.h"
+
+struct BoneWeight {
+  CPT(JointVertexTransform) joint_vertex_xform;
+  float weight;
+
+  BoneWeight(CPT(JointVertexTransform) joint_vertex_xform, float weight)
+    : joint_vertex_xform(joint_vertex_xform), weight(weight)
+  {}
+};
+typedef pvector<BoneWeight> BoneWeightList;
 
 ////////////////////////////////////////////////////////////////////
 //     Function: AssimpLoader::Constructor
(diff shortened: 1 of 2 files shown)


rdb:
This looks great!  Thank you!

I am a bit worried about char_cmp.  It looks like you defined it to be an appropriate comparison function for an ordered map, which only needs a less-than comparison function, but not for a hash_map, which needs a hash function and an equality comparison function.  The way you defined it causes it to use the pointer value for a hash, and your less-than function as the equality comparison function.
You can take a look at dtool/src/dtoolbase/stl_compares.h to see how a hash comparison function is defined.  Or you can switch to a regular pmap, which will happily work with your char_cmp.

In any case, I recommend just making the key a regular string.  Panda provides string_hash in stl_compares.h which will allow you to use an std::string as hash key.

Also, and this is not criticism but more of an FYI, but you only ever need to pass PT() into a function argument when you're transferring ownership into that function (with std::move()).  In the case of create_joint(), you know that the caller holds a reference to the character and bundle its duration, so you could safely get away with making those arguments regular pointers.  Anyway, this doesn't really matter, and I'm happy to accept either approach.

Moguri:
Looks like I might have been using pmap since I don't think HAVE_STL_HASH was defined. Which makes me wonder how well behaved phash_map is. If I define an equals compare, can it still gracefully fallback to pmap if HAVE_STL_HASH is undefined? For now, I've just switched back to pmap since there isn't a performance concern (you'd need quite a lot of characters or bones for phash_map to beat out pmap in any meaningful way).

I've also removed the PT() macros from the create_joint() function declaration.

rdb:
Ah, good thinking!  Indeed, it seems that hash_map isn't actually used on GCC.  C++11 now offers unordered_map anyway, so perhaps we should think about defining it based on that.

It does seem that it's designed to allow defining a hash function with both equality/hash operators and a less-than operator to allow falling back to pmap.

Anyway, this looks great!  Thanks a lot!

PR #53, opened 2015-10-09
why isnt this pandac.PandaModules import *

mgracer48:


rdb:
Hmm, direct/src/extensions is deprecated.  I believe it is no longer used.  I will accept a pull request that deletes that directory.

mgracer48:
wow never knew! lol because when i compile my game with panda3d i always get cant import direct.extensions

PR #55, opened 2015-10-09
remove extensions folder

mgracer48:


mgracer48:
according to @rdb this is deprecated so i removed the folder :)

mgracer48:
when you deprecated  this i wonder why you guys forgot to delete this folder?

mgracer48:
and what about extensions_native is that deprecated too?

rdb:
The extensions folder was replaced by extensions_native.  I think it was left around at the time for people who still relied on the old method of building the Panda bindings, but nobody has done that in many years now.

extensions_native is still in use today.  Ideally, we'd like to phase it out, by moving the relevant code into C++ instead.

mgracer48:
another question, did you work for disney?

PR #64, opened 2015-11-29
openal: Grab volume from manager instead of assuming max volume.

xHarv14:
Apologies for the whitespace cleanup (editor does so automatically).

This fixes the assumption that the manager has a volume level of 1.0 when the sound is constructed, and instead uses the volume within the manager.

### DIFF
--- a/panda/src/audiotraits/openalAudioSound.cxx
+++ b/panda/src/audiotraits/openalAudioSound.cxx
@@ -34,7 +34,7 @@ TypeHandle OpenALAudioSound::_type_handle;
 ////////////////////////////////////////////////////////////////////
 //     Function: OpenALAudioSound::Constructor
 //       Access: Private
-//  Description: 
+//  Description:
 ////////////////////////////////////////////////////////////////////
 
 OpenALAudioSound::
@@ -49,7 +49,7 @@ OpenALAudioSound(OpenALAudioManager* manager,
   _loops_completed(0),
   _source(0),
   _manager(manager),
-  _volume(1.0f),
+  _volume(manager->get_volume()),
   _balance(0),
   _play_rate(1.0),
   _positional(positional),
@@ -92,7 +92,7 @@ OpenALAudioSound(OpenALAudioManager* manager,
 ////////////////////////////////////////////////////////////////////
 //     Function: OpenALAudioSound::Destructor
 //       Access: public
-//  Description: 
+//  Description:
 ////////////////////////////////////////////////////////////////////
 OpenALAudioSound::
 ~OpenALAudioSound() {
@@ -133,12 +133,12 @@ play() {
   if (_manager == 0) return;
 
   PN_stdfloat px,py,pz,vx,vy,vz;
-  
+
   if (!_active) {
     _paused = true;
     return;
   }
-  
+
   stop();
 
   require_sound_data();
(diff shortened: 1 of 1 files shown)


mgracer48:
Sup Harv gib doodles

kamgha:
This PR needs to be reverted: `AudioSound._volume` is separate from `AudioManager._volume`. The net volume of the `AS` is capped by `AM._volume`, since `AS.play()` will call `AS.set_volume()` which multiplies `AS._volume` by `AM._volume` already: https://github.com/panda3d/panda3d/blob/d4eccee3590213b21ca1696fd75cf259e7ad1e88/panda/src/audiotraits/openalAudioSound.cxx#L600

This PR causes a new `AudioSound` to have a net volume of `AM._volume^2`, which would be wrong! Example: `AM.set_volume(0,1)` , after playing the `AS`, a volume of `0.1*0.1=0.01` would come out of the loudspeaker. Correct would be `0.1*1.0=0.1`, like before this "fix". While playing, I can `AS.set_volume(1.0)` and you'll hear that the sound becomes louder on the loudspeaker, to the level on what you'd expect.

@CFSworks

CFSworks:
I'll have to look into this; I distinctly remember this change fixing the volume levels in the music box sample. Before applying, the music box would play at full volume when first opened, but adjust to the correct level upon moving the slider.

If this PR isn't addressing the correct bug, then the bug with the volume slider may very well still be in the codebase, and I'd like to fix it or provide the commit where it was fixed as justification for removing this commit.

kamgha:
I've just reverted this PR here on my computer and tested the music-box sample and can not confirm that the music plays at full volume. I can move the slider slowly and make the music louder. Looking at the code of the sample, the AudioSound is initialized (loaded), directly followed by a `.setVolume(0.5)` overwriting what has been set in the constructor. Additionally, since it's using the AudioManager initialized in ShowBase, the volume of that AM would/should be its default value of 1.0 anyway, resulting this PR not to have any noticeable effect on the music-box sample.

It may be a different story when using Miles or FMOD.

hsinon:
Oh huh, that's rather interesting...

I do recall this fixing the music box when we landed this change, but the presented conclusion also makes sense. The only thing I can think of without digging deeper into this is that during construction, that multiplier isn't applied (perhaps `_source` is null at this point). If this is the case, perhaps some logic changed and now the multiplier is actually being applied, whereas it wasn't at the time of this commit.

kamgha:
I've just tested this with Panda 1.9.0, which was released before this PR. I still can not confirm the described bug with the music-box sample, it's all the same between all tested versions.

However, I tried this on a second computer (a laptop) as well, and found that the volume-effect between 0.5 and 1.0 is much less noticeable, almost barely, and I had to increase the volume for the loudspeaker inside the OS a lot to make it more clear. Is this related to the sound driver? I do not have a meter here to actually visualize the signal. I have replaced the music with a tone (see attached file), it would probably be impossible to find this with the music.
[musicbox.zip](https://github.com/panda3d/panda3d/files/2372242/musicbox.zip)

rdb:
@hsinon: Would you have time to make a test case to demonstrate the original issue?  Otherwise I'm inclined to back this out for now for the 1.10 release.

CFSworks:
Given that the API clearly intends for `AudioManager::set_volume` not to affect `AudioSound::get_volume`, I think we need to back this out regardless. The bug may not reappear (I did change up the OpenAL backend quite a lot since then) but if it does we can at least be better prepared to study it (and can also add a proper regression test to the test suite).

rdb:
I've reverted this in fb8a2f2af0b3b695885197616c514eef15ecbec8; we can open a new issue with a test case to discuss the original issue that @hsinon ran into.

PR #65, opened 2015-12-04
Use `is` operator for comparing with `None` (Pep8)

mgracer48:
[Review on Reviewable](https://reviewable.io/reviews/panda3d/panda3d/65)

### DIFF
--- a/contrib/src/sceneeditor/collisionWindow.py
+++ b/contrib/src/sceneeditor/collisionWindow.py
@@ -294,7 +294,7 @@ class collisionWindow(AppShell):
         widget.configure(entry_state = initialState)
         widget.pack(side=LEFT)
         self.widgetDict[catagory+id] = widget
-        if defaultButton and (defaultFunction!=None):
+        if defaultButton and (defaultFunctionis notNone):
             widget = Button(frame, text=buttonText, font=('MSSansSerif', 10), command = defaultFunction)
             widget.pack(side=LEFT, padx=3)
             self.widgetDict[catagory+id+'-'+'DefaultButton']=widget
--- a/contrib/src/sceneeditor/collisionWindow.py
+++ b/contrib/src/sceneeditor/collisionWindow.py
@@ -294,7 +294,7 @@ class collisionWindow(AppShell):
         widget.configure(entry_state = initialState)
         widget.pack(side=LEFT)
         self.widgetDict[catagory+id] = widget
-        if defaultButton and (defaultFunctionis notNone):
+        if defaultButton and (defaultFunction is not None):
             widget = Button(frame, text=buttonText, font=('MSSansSerif', 10), command = defaultFunction)
             widget.pack(side=LEFT, padx=3)
             self.widgetDict[catagory+id+'-'+'DefaultButton']=widget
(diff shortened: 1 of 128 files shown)


contrib/src/sceneeditor/collisionWindow.py
```diff
@@ -294,7 +294,7 @@ def createEntryField(self, parent, catagory, id, value,
         widget.configure(entry_state = initialState)
         widget.pack(side=LEFT)
         self.widgetDict[catagory+id] = widget
-        if defaultButton and (defaultFunction!=None):
+        if defaultButton and (defaultFunctionis notNone):
```
rdb: This is invalid syntax.  I presume you did not test the code?

mgracer48: that damn bot can you fix it for me

tobspr: It looks like you just replaced == with is , and != with is None, however that fails because most of the time there are no spaces, how did you replace those files, and did you have a look at them before committing them?

mfwass: @mgracer48 ```!= None``` and ```is not None``` are the same exact thing, what's the point in changing it to begin with?

rdb: It's apparently a popular style guideline to use an identity comparison instead of an equivalence comparison for the constants None, True and False.

tobspr: http://stackoverflow.com/questions/3257919/is-none-vs-none as reference ^

mgracer48: i didnt replace it a bot did :) because it is used in a lot of places

mgracer48:
fixed the issue guys

contrib/src/sceneeditor/dataHolder.py
```diff
@@ -737,7 +737,7 @@ def addCollisionObject(self, collisionObj, nodePath, pointA=None, pointB=None, p
         node.addSolid(collisionObj)
         self.collisionDict[name] = nodePath.attachNewNode(node)
 
-        if pointA!=None:
+        if pointAis notNone:
```
rdb: Here's an example of an invalid syntax.  Wouldn't it be quicker to just run a find and replace through the codebase?

mgracer48:
did i fix them now?

mgracer48:
@rdb what still makes u not want to merge?

contrib/src/sceneeditor/propertyWindow.py
```diff
@@ -375,7 +375,7 @@ def redrawCurveProperty(self, nodePath, curveList):
         else:
             self.ignore('curveRemovedFromNode')
             
-        if curveList!= None:
+        if curveListis not None:
```
loblao: Syntax error.

rdb:
Regardless of all the syntax errors, this is a bit of a silly issue.  While it is certainly good style to use "is" for comparing against constants in Python, changing existing code to follow that gains us absolutely nothing at the risk of breaking code that happens to deal with None in its comparison operator.

Closing this for good.

PR #66, opened 2015-12-04
Use dict comprehension instead of dict([...])

mgracer48:
[Review on Reviewable](https://reviewable.io/reviews/panda3d/panda3d/66)

### DIFF
--- a/direct/src/directutil/DirectMySQLdbConnection.py
+++ b/direct/src/directutil/DirectMySQLdbConnection.py
@@ -15,8 +15,8 @@ class DirectMySQLdbConnection(Connection):
 
         conv = kwargs.get('conv', conversions)
 
-        kwargs2['conv'] = dict([ (k, v) for k, v in conv.items()
-                                 if type(k) is int ])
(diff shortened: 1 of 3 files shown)


rdb:
I'm not opposed to this type of style change in principle, but we still target Python 2.6 as a supported version.  This change would break support for Python 2.6.  Perhaps we'll make a change like this sometime in the future.

pennomi:
For what it's worth, Python 2.6 is way past its shelf life (official support ended in 2013). Any reason we're not targeting 2.7+?

rdb:
Because Mac OS X Snow Leopard only ships with Python 2.6, and there are a number of our users who have indicated that they would like Panda3D to keep supporting Snow Leopard for now.  We do use and recommend Python 2.7+ everywhere else.

In general, I don't like breaking things that people are still using unless there was a compelling reason to do so.  A minor point in a style guide doesn't register as particularly compelling to me.

pennomi:
Perfect. That's the reasoning I was looking for. :smile:

PR #68, opened 2016-01-01
Assimp: Add initial support for skeletal mesh animations

Moguri:
[skel_anim.zip](https://github.com/panda3d/panda3d/files/76065/skel_anim.zip) contains:
* animbox.blend - Original blend file
* animbox.dae - COLLADA file exported from Blender via Godot's Better COLLADA addon
* main.py - Script used to test the changes

Known issues:
* A model will fail to load if all channels in an animation do not have the same number of keyframes
* Any animations on the skeleton/armature itself are not converted since it is not a bone (and thus not in the bonemap)
* Animations from COLLADA files exported by Godot's Better COLLADA addon do not have a name
  * Not sure if this is an issue with the addon or with Assimp
  * For the time being, all AnimBundleNodes are given the name "anim" since the animations mName is empty

### DIFF
--- a/pandatool/src/assimp/assimpLoader.cxx
+++ b/pandatool/src/assimp/assimpLoader.cxx
@@ -32,6 +32,9 @@
 #include "look_at.h"
 #include "texturePool.h"
 #include "character.h"
+#include "animBundle.h"
+#include "animBundleNode.h"
+#include "animChannelMatrixXfmTable.h"
 #include "pvector.h"
 
 #include "pandaIOSystem.h"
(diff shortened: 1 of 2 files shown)


rdb:
Many thanks! :-) I'm happy to approve the changes so far, it looks good.  The model I tried doesn't quite work yet as I indicated, but given that this isn't a public feature yet it shouldn't be a problem to have the code in there.

I'm not lying awake over a single spacing mismatch, which can be fixed in a future commit (as well as the dangling { at the top of the create_anim_channel method).

PR #69, opened 2016-01-02
Assimp: Support skeletons with unconnected bones

Moguri:
We needed to detect islands of bones and create anims and parts for them.


Moguri:
Use the goblin file you provided me to test this. It now imports without crashing or throwing an assertion. However, it's still a mess.

PR #71, opened 2016-01-12
add makepanda.sh

mgracer48:
there is a makepanda.bat time for a unix one
works for linux and mac
@rdb thoughts?

[Review on Reviewable](https://reviewable.io/reviews/panda3d/panda3d/71)


rdb:
Why?  makepanda.py already has a shebang, so you can run it directly on UNIX.

I don't see much added value from this sh file.

mgracer48:
i get ur point if only sh files could be clicked on like bats that be cool

PR #72, opened 2016-01-18
Allow setting width and numLines of DirectEntry without clearing text

assertivist:
I didn't even test it.


assertivist:
Tested this to work with Schwarzbaer's console code (the use case that brought about the change in the first place :smiley:). Changing 

`self.entry.guiItem.set_max_width(wrap_width)` 

to 

`self.entry['width'] = wrap_width`

behaves as intended; similarly with the numLines property.

rdb:
Thanks!

PR #80, opened 2016-02-18
Add new terrain implementation (ShaderTerrainMesh)

tobspr:
This PR adds a new ShaderTerrainMesh class which is capable of rendering huge terrains on the GPU, using shaders.
Further information can be found here: <a href="http://www.panda3d.org/manual/index.php/ShaderTerrainMesh">ShaderTerrainMesh</a>

Additionally a new sample (`samples/shader-terrain`) was added to show the basic usage of the terrain.

The terrain has several advantages over the existing terrain implementations in Panda3D (HeightfieldTesselator, GeoMipTerrain), mainly being much faster and capable of handling huge heightfields.


tobspr:
@wezu thats true, I'll change that

PR #83, opened 2016-02-26
Add a --wheel flag for generating `.whl` distributions.

pennomi:
Here's the first revision of the wheel generator, ready for criticism. I've only tested on Linux, but I don't see any reason it shouldn't work on all platforms as-is.

For Windows and Mac, we can upload these archives to PyPI, but for Linux, we'll need to host a URL for them to live at, since PyPI doesn't allow binary distributions for Linux.

### DIFF
--- a/makepanda/makewheel.py
+++ b/makepanda/makewheel.py
@@ -26,14 +26,14 @@ def get_platform():
     return p
 
 # Other global parameters
-PY_VERSION = "py3" if sys.version_info >= (3, 0) else "py2"
+PY_VERSION = "cp{}{}".format(sys.version_info.major, sys.version_info.minor)
+ABI_TAG = "none"
 PLATFORM_VERSION = get_platform()
 
 WHEEL_DATA = """Wheel-Version: 1.0
 Generator: bdist_wheel 1.0
 Root-Is-Purelib: false
-Tag: py2-none-any
-Tag: py3-none-any
+Tag: {}-{}-{}
 Build: 1
 """
 
@@ -130,8 +130,8 @@ def makewheel(version, output_dir):
     direct_target = join(wheel_target, "direct")
     data_target = join(wheel_target, "panda3d-{}.data".format(version))
     info_target = join(wheel_target, "panda3d-{}.dist-info".format(version))
-    wheel_name = "panda3d-{}-{}-none-{}.whl".format(
-        version, PY_VERSION, PLATFORM_VERSION)
+    wheel_name = "panda3d-{}-{}-{}-{}.whl".format(
+        version, PY_VERSION, ABI_TAG, PLATFORM_VERSION)
 
     # Update relevant METADATA entries
     METADATA['version'] = version
@@ -182,7 +182,7 @@ def makewheel(version, output_dir):
         # TODO: Populate the RECORD file with all the proper hashes
         pass
     with open(join(info_target, 'WHEEL'), 'w') as outfile:
-        outfile.write(WHEEL_DATA)
+        outfile.write(WHEEL_DATA.format(PY_VERSION, ABI_TAG, PLATFORM_VERSION))
     write_metadata(info_target)
 
     # Zip it up and name it the right thing
--- a/makepanda/makewheel.py
+++ b/makepanda/makewheel.py
@@ -25,23 +25,6 @@ def get_platform():
         return "any"
     return p
 
-
-# Global filepaths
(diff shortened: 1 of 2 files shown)


makepanda/makewheel.py
```diff
@@ -0,0 +1,195 @@
...
+    if "linux" in p:
+        print(ColorText("red", "WARNING:") +
+              " Linux-specific wheel files are not supported."
+              " We will generate this wheel as a generic package instead.")
+        return "any"
+    return p
+
+
+# Global filepaths
+LIB_SRC = join("built", "lib")  # TODO: A special wheel lib dir
```
rdb: The "built" directory should be a parameter to the makewheel function.  On 64-bit Windows it is called "built_x64" and we actually allow overriding this (which I do a lot, and so does the buildbot) via the --outputdir parameter.

See makepandacore.GetOutputDir()

makepanda/makewheel.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    for f in os.listdir(src):
+        shutil.copy2(join(src, f), join(target, f))
+
+
+def makewheel(version):
+    # Global parameters that depend on version
+    data_target = join(WHEEL_TARGET, "panda3d-{}.data".format(version))
+    info_target = join(WHEEL_TARGET, "panda3d-{}.dist-info".format(version))
+    wheel_name = "panda3d-{}-{}-none-{}.whl".format(
```
rdb: abi-tag should not be "none" but "cp27", etc.

rdb: See https://www.python.org/dev/peps/pep-0425/ for more information about python tags and ABI tags - we need to indicate that we're fixed to CPython, and a particular version of it.

makepanda/makewheel.py
```diff
@@ -0,0 +1,195 @@
...
+
+# Other global parameters
+PY_VERSION = "py3" if sys.version_info >= (3, 0) else "py2"
+PLATFORM_VERSION = get_platform()
+
+WHEEL_DATA = """Wheel-Version: 1.0
+Generator: bdist_wheel 1.0
+Root-Is-Purelib: false
+Tag: py2-none-any
+Tag: py3-none-any
```
rdb: Should this tag not match what is indicated by the filename, ie. dependence on a specific CPython version?

makepanda/makewheel.py
pennomi: Note here that the wheel wouldn't install unless the ABI tag was "none". However, I did adjust the python version to explicitly require cpython.

auvipy:
thats awesome move!!

makepanda/makewheel.py
```diff
@@ -0,0 +1,296 @@
...
+    whl.write_file_data(info_dir + '/metadata.json', json.dumps(METADATA, indent=4, separators=(',', ': ')))
+    whl.write_file_data(info_dir + '/METADATA', metadata)
+    whl.write_file_data(info_dir + '/WHEEL', WHEEL_DATA.format(PY_VERSION, ABI_TAG, PLATFORM_TAG))
+    whl.write_file(info_dir + '/LICENSE.txt', license_src)
+    whl.write_file(info_dir + '/README.md', readme_src)
+    whl.write_file_data(info_dir + '/top_level.txt', 'direct\npanda3d\npandac\n')
+
+    whl.close()
+
+# makewheel("1.10.0", "built_x64")
```
auvipy: add a new line here

rdb:
OK, binary .whl packages for Panda3D 1.9.3 for Windows, macOS and Linux have been uploaded to PyPI.  `pip install panda3d` should now work on those platforms.  I'd welcome people to help test to make sure that every aspect of Panda3D still works correctly when installed via pip.

Thanks again!

auvipy:
eagerly waiting for 1.10 stable release with py3k support <3

pennomi:
I've done some very simple tests on Linux and can confirm it's working when installed via pip on Python 2. All of my bigger projects are on Python 3 however, so I can't test much more than a couple of tech demo programs.

auvipy:
I guess pip3 install from github master branch should also work fine?

pennomi:
Installing from source almost certainly won't work. Unlike most installable python packages, the wheel file is directly compiled into a binary format, bypassing the normal `setup.py` process.

Basically, what we're doing is very atypical for a python package, but is necessary due to the complex compilation process.

wezu:
Tested the pip version on windows 10 and python 2.7.13. All the samples and bullet samples worked.

PR #88, opened 2016-03-27
Handle Python 2 vs Python 3 stdlib differences when building the morepy package

Moguri:
See PEP3108 for more details on stdlib differences. This has only been tested on Python 3.


rdb:
Many thanks!

PR #89, opened 2016-03-27
Add portable installers for Linux and Windows

Moguri:
NOTE: Windows is untested


auvipy:
kool

Moguri:
TODO before inclusion into master:
* Ensure Windows portable installer works
* Use tar.gz on Linux instead of zip

Control over which installers to build and what compression to use will be added later (possibly with a pdef-style file).

Moguri:
This PR has been replace by #112

PR #90, opened 2016-03-29
do not mutate dict we are iterating over

brakhane:
Risky at best, and actually raises an error on modern Python versions.


rdb:
Thanks, missed this during the port to Python 3.

PR #91, opened 2016-04-02
Polygon polygon collision

amras0000:
The surface point is imperfect, but the collision is able to consistently detect when two polygons intersect, and provide an accurate normal for the collision.
I haven't had performance issues, but there are some optimizations that could be made if we want to compare large polysets.

### DIFF
--- a/panda/src/collide/collisionPolygon.cxx
+++ b/panda/src/collide/collisionPolygon.cxx
@@ -966,6 +966,11 @@ test_intersection_from_polygon(const CollisionEntry &entry) const {
   const CollisionPolygon *from_polygon;
   DCAST_INTO_R(from_polygon, entry.get_from(), 0);
   
+  CPT(TransformState) wrt_space = entry.get_wrt_space();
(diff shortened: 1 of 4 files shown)


amras0000:
Optimizations:
- bitwise ^ could be replaced with !=
- average is technically unneccessary, could be replaced with intersection_segment_0[0] etc
- line_dir could be calculated as intersection_segment_0[1] - intersection_segment_0[0]
- this_normal and from_normal should be calculated after rederiving the 3d_mats, since calling get_point() rederives the mats again (switch to to_3d)
- from_3d_mat should be pre-multiplied with wrt_mat

Surface point:
Would probably be wisest to find the intersection of the two intersection segments (order seg_00 through seg_11, grab the middle two values), then average those two points and project the result onto the surface.

Might come back to this and implement them later but I want to get back to other stuff for now

Kupoman:
This PR has been inactive for quite some time, can it be closed?

rdb:
This is waiting to be picked up and continued by someone; I suppose it can be reopened if someone steps in.

PR #93, opened 2016-04-10
Change uniform warning in glShaderContext to avoid spam

tobspr:
This changes the severity of the uniform warning to info, to avoid spamming the console


rdb:
Pushed with modification.  Thanks.

PR #98, opened 2016-04-24
Fix error "TypeError: 'NoneType' object is not iterable"

darktohka:
filter(request, args) can return None, and in that case an error will be thrown.

The list() procedure is unnecessary, since if the filter returns a string, it will be wrapped into a tuple anyways.


rdb:
That looks like a regression introduced by 2to3.  Thanks for fixing it!

PR #105, opened 2016-05-09
Add DynamicHeightfield. Add properties to Bullet.

wolfgangp:
Everything except BulletWheel is backwards compatible. I doubt the change there will affect many people. Up until now the API allowed setting steering in degrees on the BulletVehicle, or in radians on its BulletWheels. Now it's degrees in both cases.

BulletVehicle basic functions before (still works!):

```
self.vehicle.set_steering_value(steering, 0)
self.vehicle.set_brake(brake_force, 2)
self.vehicle.apply_engine_force(engine_force, 3)
```

Works now and may be future recommendation:

```
self.vehicle.wheels[0].steering = steering
self.vehicle.wheels[2].brake = brake_force
self.vehicle.wheels[3].engine_force = engine_force
```


rdb:
Is it intended that _y_size and _x_size are swapped around in some code?

wolfgangp:
No, swapping these around was not intended, and must have be a remnant of me trying to keep something straight in my head about the Bullet loops.
In any case, I amended that and added a method to rotate PfmFiles (with the side-effect of making this PR yet more confusing scope-wise :-/ )

wolfgangp:
Sample for DynamicHeightfield (requires Bullet and models from bullet-samples):

[https://github.com/wolfgangp/panda3d_bullet_samples/blob/master/27_HeightfieldVehicle.py](27_HeightfieldVehicle.py)

rdb:
These changes are already being addressed by #140 and #141.

PR #106, opened 2016-05-15
Fix Python3 support

Moguri:


Moguri:
I'm not sure how well this meshes with Direct's style, but this is probably the more "Pythonic" way of checking for strings.

PR #110, opened 2016-05-23
Fixed an issue in ServerRepository, where clients can send incomplete…

SkippsDev:
… packets and crash the server.

### DIFF
--- a/direct/src/distributed/ServerRepository.py
+++ b/direct/src/distributed/ServerRepository.py
@@ -359,8 +359,11 @@ class ServerRepository:
             #datagram.dumpHex(ostream)
 
         dgi = DatagramIterator(datagram)
-
         type = dgi.getUint16()
+        
+        if dgi.getRemainingSize() == 0 and type != CLIENT_DISCONNECT_CMU:
+            # This prevents a crash where their isn't enough data recieved to unpack corresponding to the msgType.
+            return
 
         if type == CLIENT_DISCONNECT_CMU:
             self.handleClientDisconnect(client)


direct/src/distributed/ServerRepository.py
```diff
@@ -359,8 +359,11 @@ def handleDatagram(self, datagram):
             #datagram.dumpHex(ostream)
 
         dgi = DatagramIterator(datagram)
-
         type = dgi.getUint16()
+        
+        if dgi.getRemainingSize() == 0 and type != CLIENT_DISCONNECT_CMU:
+            # This prevents a crash where their isn't enough data recieved to unpack corresponding to the msgType.
```
mfwass: Might want to correct the grammar.

their -> there

loblao:
I'd like to add that ServerRepository.py is not really "fixable", there are countless ways to exploit/crash it.

SkippsDev:
There's one less way to crash it from now on.

CFSworks:
While this fix is *functional* I don't think it's *correct*. It only fixes one manifestation of the "unexpected end of datagram" crash bug (a non-disconnect packet with no payload). What happens if you only send one byte of payload to something expecting multiple bytes? The check would allow the datagram through but it would still crash upon processing.

A better fix would be to ensure that all data is read from the dgi *first*, and wrap that in a try/except to catch reads past the end of the iterator. The best fix would be to have each message-type handler check the length before attempting to process the message. (And in either case, an error should probably result in the client getting dropped for violating the protocol.)

SkippsDev:
However i was just graciously fixing an exploit because i noticed it was just something that needed to be done... I never indented to completely "fix" the server repository, just figured i would help out.

PR #113, opened 2016-06-26
OpenMayaMac.h Removed

Stellarspace:
OpenMayaMac.h has been removed as it is no longer used. Customer's legacy projects must be updated to remove the use of the header file.


rdb:
Thanks, the fix has been applied in a portable manner.

PR #114, opened 2016-06-28
Fix a TypeError in direct.fsm.FSM

mfwass:


loblao:
This is to fix an issue which happens when a class is a subclass of `NodePath` and `FSM` at the same time.

mfwass:
:)

PR #115, opened 2016-07-09
Implement make_copy() for BulletRigidBodyNode

Moguri:
This appears to be the last piece needed to load BulletRigidBodyNodes
from BAM files.

### DIFF
--- a/panda/src/bullet/bulletRigidBodyNode.cxx
+++ b/panda/src/bullet/bulletRigidBodyNode.cxx
@@ -47,6 +47,28 @@ BulletRigidBodyNode(const char *name) : BulletBodyNode(name) {
 }
 
 /**
+ * Do not call the copy constructor directly; instead, use make_copy() or
+ * copy_subgraph() to make a copy of a node.
+ */
+BulletRigidBodyNode::
+BulletRigidBodyNode(const BulletRigidBodyNode &copy) :
+  BulletBodyNode(copy)
+{
+  _motion = new MotionState(*copy._motion);
+  _rigid = new btRigidBody(*copy._rigid);
+}
+
+/**
+ * Returns a newly-allocated PandaNode that is a shallow copy of this one.  It
+ * will be a different pointer, but its internal data may or may not be shared
+ * with that of the original PandaNode.  No children will be copied.
+ */
+PandaNode *BulletRigidBodyNode::
+make_copy() const {
+  return new BulletRigidBodyNode(*this);
+}
+
+/**
  *
  */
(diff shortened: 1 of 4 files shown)


Moguri:
Adding a BAM file and Python script to test the changes.

[bulletbamtest.zip](https://github.com/panda3d/panda3d/files/355662/bulletbamtest.zip)

rdb:
This does not seem like the right solution.  These copy constructors do a shallow copy of the Bullet pointers, which means that you end up with two BulletRigidBodyNode objects that manipulate the same underlying Bullet rigid body.

Worse yet, the _shape pointer is shared between BulletBodyNode instances, but since it is not reference counted, as soon as one of the copies is destroyed, the other will have a shape pointer that points to deleted memory.

Moguri:
I have updated the bulletbam branch with your suggestions.

panda/src/bullet/bulletRigidBodyNode.cxx
```diff
@@ -47,6 +47,28 @@ BulletRigidBodyNode(const char *name) : BulletBodyNode(name) {
...
 /**
+ * Do not call the copy constructor directly; instead, use make_copy() or
+ * copy_subgraph() to make a copy of a node.
+ */
+BulletRigidBodyNode::
+BulletRigidBodyNode(const BulletRigidBodyNode &copy) :
+  BulletBodyNode(copy)
+{
+  _motion = new MotionState(*copy._motion);
+  _rigid = new btRigidBody(*copy._rigid);
```
rdb: You need to call `setUserPointer` on `_rigid` since it otherwise still points to the first object.

panda/src/bullet/bulletRigidBodyNode.cxx
```diff
@@ -47,6 +47,28 @@ BulletRigidBodyNode(const char *name) : BulletBodyNode(name) {
...
 
 /**
+ * Do not call the copy constructor directly; instead, use make_copy() or
+ * copy_subgraph() to make a copy of a node.
+ */
+BulletRigidBodyNode::
+BulletRigidBodyNode(const BulletRigidBodyNode &copy) :
+  BulletBodyNode(copy)
+{
+  _motion = new MotionState(*copy._motion);
```
rdb: Since the rigid body is initialised with the motion state, you need to call setMotionState to initialise it with the new motion state.

rdb:
In an effort to move things along I've made the requested modifications myself and committed it as 81dd4d97ad7dfc622e3a8e7e33f22a9bcaa2a182.

Thanks!

PR #116, opened 2016-07-13
Fixed exploits with unexpected data, which causes the ServerRepositor…

SkippsDev:
…y to crash.

### DIFF
--- a/direct/src/distributed/ServerRepository.py
+++ b/direct/src/distributed/ServerRepository.py
@@ -358,8 +358,10 @@ class ServerRepository:
             #datagram.dumpHex(ostream)
 
         dgi = DatagramIterator(datagram)
-
         type = dgi.getUint16()
+        
+        if dgi.getRemainingSize() == 0 and type != CLIENT_DISCONNECT_CMU:
+            return
 
         if type == CLIENT_DISCONNECT_CMU:
             self.handleClientDisconnect(client)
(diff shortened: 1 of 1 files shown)


direct/src/distributed/ServerRepository.py
```diff
@@ -372,20 +374,33 @@ def handleDatagram(self, datagram):
...
+        del self.clientsByDoIdBase[client.doIdBase]
+        self.qcr.removeConnection(connection)
 
     def handleClientCreateObject(self, datagram, dgi):
         """ client wants to create an object, so we store appropriate
         data, and then pass message along to corresponding zones """
 
         connection = datagram.getConnection()
+        
+        if dgi.getRemainingSize() < 3:
```
loblao: Still exploitable. Since `getUint32` reads 4 bytes and `getUint16` reads 2, you need 10 bytes.

loblao:
There's a major issue in SR: it doesn't inspect client updates (`dg.appendData(dgi.getRemainingBytes())`) so those bytes are sent to the clients and could be manipulated to crash them.

direct/src/distributed/ServerRepository.py
SkippsDev: Oh yeah, idek why i didn't see that... ugh....

rdb:
Closing due to inactivity despite outstanding issues.

I feel that addressing this kind of issue requires a more thorough approach to handling out-of-bounds datagram reads.  Hardcoding expected byte lengths is a rather fragile approach.

PR #118, opened 2016-08-14
Added fixed functions pipeline checking in runtime (bad)

sanguinariojoe:
The possibility of removing the fixed functions pipeline in compilation time is quite nice. However, the engine should be able to automatically detect when they are not available in runtime as well.
Otherwise the core profile OpenGL implementations will not properly work, for instance, in generic binary distributed packages of the engine

### DIFF
--- a/panda/src/display/graphicsStateGuardian.cxx
+++ b/panda/src/display/graphicsStateGuardian.cxx
@@ -245,6 +245,9 @@ GraphicsStateGuardian(CoordinateSystem internal_coordinate_system,
   _supports_geometry_instancing = false;
   _supports_indirect_draw = false;
 
+  // We are safe assuming it has luminance support
+  _supports_luminance_texture = true;
+
   // Assume a maximum of 1 render target in absence of MRT.
   _max_color_targets = 1;
   _supports_dual_source_blending = false;
--- a/panda/src/glstuff/glGraphicsStateGuardian_src.I
+++ b/panda/src/glstuff/glGraphicsStateGuardian_src.I
@@ -131,6 +131,38 @@ get_gl_version_minor() const {
 }
 
 /**
+ * Returns whether a core profile or a compatibility mode is considered.
+ */
+INLINE bool CLP(GraphicsStateGuardian)::
+has_core_profile() const {
+  return _core_profile;
+}
+
+/**
+ * Returns whether the fixed function pipeline is supported.
+ */
+INLINE bool CLP(GraphicsStateGuardian)::
+has_fixed_function_pipeline() const {
+#ifndef SUPPORT_FIXED_FUNCTION
+  return false;
+#else
+  // In OpenGL 3.1 the fixed function pipeline was declared deprecated, and
+  // theoretically all the pipeline should gone. However, some implementations
+  // kept them, "out of the law". After that, In OpenGL >= 3.2 the concept of
+  // core profile and compatibility mode where introduced. Therefore, since it
+  // is not possible to check whether those functions are implemented in GL 3.1
+  // or not, we should not try to use them.
+  if ((_gl_version_major == 3) && (_gl_version_minor == 1)) {
+    return false;
+  }
+  // Otherwise, we can just check whether we are using a core profile or a
+  // compatibility mode. The variable _core_profile is already taking into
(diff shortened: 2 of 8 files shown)


panda/src/glstuff/glGraphicsStateGuardian_src.I
```diff
@@ -131,6 +131,39 @@ get_gl_version_minor() const {
...
+#ifndef SUPPORT_FIXED_FUNCTION
+  return false;
+#else
+  // In OpenGL 3.1 the fixed function pipeline was declared deprecated, and
+  // theoretically all the pipeline should gone. However, some implementations
+  // kept them, "out of the law". After that, In OpenGL >= 3.2 the concept of
+  // core profile and compatibility mode where introduced. Therefore, since it
+  // is not possible to check whether those functions are implemented in GL 3.1
+  // or not, we should not try to use them.
+  if ((_gl_version_major == 3) && (_gl_version_minor == 1)) {
```
rdb: This looks scary.  Forcing people into core-only mode will have really bad results since a lot of out-of-the-box Panda functionality will be unavailable.  If we can't be sure whether we have the compatibility extension, we should stay on the safe side and assume we do have it.

panda/src/glstuff/glGeomMunger_src.cxx
```diff
@@ -47,9 +47,11 @@ CLP(GeomMunger)::
...
   // We need to remove this pointer from all of the geom contexts that
   // reference this object.
-  GeomContexts::iterator gci;
-  for (gci = _geom_contexts.begin(); gci != _geom_contexts.end(); ++gci) {
-    (*gci)->remove_munger(this);
+  if (((CLP(GraphicsStateGuardian)*)get_gsg())->has_fixed_function_pipeline()) {
+    GeomContexts::iterator gci;
+    for (gci = _geom_contexts.begin(); gci != _geom_contexts.end(); ++gci) {
+      (*gci)->remove_munger(this);
+    }
```
rdb: I'm not sure why this code is protected by has_fixed_pipeline check.

panda/src/glstuff/glGraphicsStateGuardian_src.I
sanguinariojoe: That's a problem of OpenGL specification itself... For some reason they let the vendors to decide whether they are implementing fixed pipeline, but without providing a way to know it prior OpenGL >= 4... A bummer...
Theoretically, from the point of view of an engine built "from scratch", the safest way is just considering it is not available. However, as you are pointing out, this is not a new engine.

Along this line, I should say that I have not saw a single 3.1 platform without fixed pipeline... But I'm not an expert

panda/src/glstuff/glGeomMunger_src.cxx
sanguinariojoe: I cannot remember... I should investigate a bit... Could you ping me next week (this week I'm too busy with a workshop)

panda/src/display/graphicsStateGuardian.cxx
```diff
@@ -230,6 +230,7 @@ GraphicsStateGuardian(CoordinateSystem internal_coordinate_system,
   _supports_generate_mipmap = false;
   _supports_depth_texture = false;
   _supports_depth_stencil = false;
+  _supports_luminance_texture = false;
```
rdb: So setting "gl-version 3 2" in Config.prc isn't sufficient to get a core-only context in Mesa?  If I recall correctly, the spec guarantees that we get a core-only context by default when requesting "gl-version 3 2" or above.

My biggest concern is that our context creation logic is already very complex.  This PR seems to add a whole lot more logic to it to the point where it's really hard to follow what it does and how it's supposed to work.

You could consider splitting out the context stuff from this pull request to make it easier to review separately.  I certainly have no objections with getting the GSG changes in.

As for D3D9, we are probably safe to assume it has luminance support.

sanguinariojoe: Sure, I'm splitting that, as well as reviewing the comments regarding glGeomMunger_src.cxx, and the luminance support.
Probably I'll fire a new pull request, so you can close this one as rejected

sanguinariojoe:
Due to the conflicts, I'm closing this PR and opening a new one

PR #120, opened 2016-08-17
make makepandacore to get Windows 10 SDK uCRT version

MU-Software:
This will find latest Windows 10 SDK universal CRT version on \Windows Kits\10\Include .


rdb:
Added a more robust checking mechanism that checks whether the version has all the required directories and also allows choosing a specific version via the command-line.

PR #121, opened 2016-08-18
Add posScaleInterval

MU-Software:
This change will add posScaleInterval which was mentioned on Panda3D manual.


direct/src/interval/LerpInterval.py
```diff
@@ -282,6 +282,37 @@ def privDoEvent(self, t, event):
             self.setupParam(self.setStartQuat, self.startQuat)
         LerpNodePathInterval.privDoEvent(self, t, event)
 
+class LerpPosScaleInterval(LerpNodePathInterval):
+    def __init__(self, nodePath, duration, pos, scale,
+                 startPos = None, startHpr = None, startQuat = None,
```
rdb: Why does LerpPosScaleInterval have startHpr and startQuat?

MU-Software:
Ugh, Branch are entangled by mistake and, to think about it, we actually don't need this.  
So i'll close this request. Sorry for bothering you :(

PR #123, opened 2016-10-28
Serializability of bullet shapes completed.

consultit:
Working on a project of mine, I noticed that not all bullet shapes implemented the TypeWritable API for serialization on bam file. So I implemented serialization for those who lacked. In particular:
- made serializable the following shapes: BulletHeightfieldShape, BulletCapsuleShape, BulletConeShape, BulletCylinderShape, BulletConvexPointCloudShape, BulletMinkowskiSumShape, BulletMultiSphereShape
- corrected the serialized parameters for BulletBoxShape (the old version led to a growth of the box size during restoring)
- also added the linear and angular velocity as serialized data for BulletRigidBodyNode, so that also the dynamic state of the rigid body could be restored.

As far as possible I tried to follow the original author's programming style (enn0x), but I had to add some variables only used for serialization, as bullet, sometimes, changes the original creation parameters, and therefore hardly obtainable by bullet underlying structures.
I attached a test script: the scene is saved on exit on 'scene.bam' file, and put this file as argument to restore it
[bullet_shapes_test.zip](https://github.com/panda3d/panda3d/files/557862/bullet_shapes_test.zip).

@rdb


rdb:
Thanks!  I've checked it in with changes, eg. to make margin setting still work correctly, make some whitespace/formatting fixes and I removed a few changes that broke the bam backward compatibility.

I've also checked in unit tests for these changes (except for a few classes that don't have accessors that are necessary to test it).

PR #124, opened 2016-10-31
Cleanup .gitignore files

Moguri:
Remove all .gitignore files in subdirectories and add a few new ignores
to the top-level directory .gitignore file.

``git status`` still shows no untracked/modified files after a build with these changes.


rdb:
This looks OK to me.  Most of these files are from ppremake, which placed generated files right in the source directories, but since we no longer support ppremake there's no compelling reason to keep those in the `.gitignore` files any more.

PR #126, opened 2016-11-01
Add functional header to parser-inc

tobspr:
This allows interrogate to parse `std::function<>` and all other types in the standard library header `functional`.  Unfortunately interrogate can not handle variadic templates at the moment, so I had to comment out some definitions.


tobspr:
Also added the `initializer_list` and `unordered_map` headers

tobspr:
Updated the headers to use variadic templates

PR #128, opened 2016-11-04
Added fixed functions pipeline checking in runtime

sanguinariojoe:
The possibility of removing the fixed functions pipeline in compilation time is quite nice. However, the engine should be able to automatically detect when they are not available in runtime as well.
Otherwise the core profile OpenGL implementations will not properly work, for instance, in generic binary distributed packages of the engine

### DIFF
--- a/panda/src/glstuff/glCgShaderContext_src.cxx
+++ b/panda/src/glstuff/glCgShaderContext_src.cxx
@@ -762,25 +762,27 @@ disable_shader_vertex_arrays() {
       _glgsg->disable_vertex_attrib_array(p);
     } else {
 #ifdef SUPPORT_FIXED_FUNCTION
-      switch (p) {
-      case CA_unknown:
-        break;
-      case CA_vertex:
-        glDisableClientState(GL_VERTEX_ARRAY);
-        break;
-      case CA_normal:
-        glDisableClientState(GL_NORMAL_ARRAY);
-        break;
-      case CA_color:
-        glDisableClientState(GL_COLOR_ARRAY);
-        break;
-      case CA_secondary_color:
-        glDisableClientState(GL_SECONDARY_COLOR_ARRAY);
-        break;
-      default:
-        _glgsg->_glClientActiveTexture(GL_TEXTURE0 + (p - CA_texcoord));
-        glDisableClientState(GL_TEXTURE_COORD_ARRAY);
-        break;
+      if(_glgsg->has_fixed_function_pipeline()) {
+        switch (p) {
+        case CA_unknown:
+          break;
+        case CA_vertex:
+          glDisableClientState(GL_VERTEX_ARRAY);
+          break;
+        case CA_normal:
+          glDisableClientState(GL_NORMAL_ARRAY);
+          break;
+        case CA_color:
+          glDisableClientState(GL_COLOR_ARRAY);
+          break;
+        case CA_secondary_color:
+          glDisableClientState(GL_SECONDARY_COLOR_ARRAY);
+          break;
+        default:
+          _glgsg->_glClientActiveTexture(GL_TEXTURE0 + (p - CA_texcoord));
+          glDisableClientState(GL_TEXTURE_COORD_ARRAY);
+          break;
+        }
       }
 #endif  // SUPPORT_FIXED_FUNCTION
     }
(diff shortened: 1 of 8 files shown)


Moguri:
_core_profile appears to be uninitialized. Also, get_supports_luminance_texture() and has_core_profile() are both unused.

panda/src/wgldisplay/wglGraphicsStateGuardian.cxx
```diff
@@ -611,8 +611,10 @@ make_context(HDC hdc) {
       attrib_list[n++] = WGL_CONTEXT_DEBUG_BIT_ARB;
     }
 #ifndef SUPPORT_FIXED_FUNCTION
-    attrib_list[n++] = WGL_CONTEXT_PROFILE_MASK_ARB;
-    attrib_list[n++] = WGL_CONTEXT_CORE_PROFILE_BIT_ARB;
+    if(!has_fixed_function_pipeline()) {
+      attrib_list[n++] = WGL_CONTEXT_PROFILE_MASK_ARB;
+      attrib_list[n++] = WGL_CONTEXT_CORE_PROFILE_BIT_ARB;
+    }
```
rdb: I think this `if` check shouldn't be here.  This is in the context creation code, so `_core_profile` hasn't been initialised yet.  Instead, I think we should check if gl_version is 3.2 or higher.

panda/src/glstuff/glCgShaderContext_src.cxx
```diff
@@ -876,48 +878,50 @@ update_shader_vertex_arrays(ShaderContext *prev, bool force) {
...
-            }
-            glEnableClientState(GL_COLOR_ARRAY);
-            break;
-
-          case CA_secondary_color:
-            _glgsg->_glSecondaryColorPointer(num_values, type,
-                                             stride, client_pointer);
-            glEnableClientState(GL_SECONDARY_COLOR_ARRAY);
-            break;
+          if(_glgsg->has_fixed_function_pipeline()) {
```
rdb: I think this check doesn't need to be here.  We never need to set it to `CA_*` if `has_fixed_function_pipeline()` is false anyway.

panda/src/glstuff/glCgShaderContext_src.cxx
```diff
@@ -762,25 +762,27 @@ disable_shader_vertex_arrays() {
...
-        glDisableClientState(GL_COLOR_ARRAY);
-        break;
-      case CA_secondary_color:
-        glDisableClientState(GL_SECONDARY_COLOR_ARRAY);
-        break;
-      default:
-        _glgsg->_glClientActiveTexture(GL_TEXTURE0 + (p - CA_texcoord));
-        glDisableClientState(GL_TEXTURE_COORD_ARRAY);
-        break;
+      if(_glgsg->has_fixed_function_pipeline()) {
```
rdb: I think this check doesn't need to be here. We never need to set it to CA_* if has_fixed_function_pipeline() is false anyway.

panda/src/wgldisplay/wglGraphicsStateGuardian.cxx
sanguinariojoe: Sure, my fault! Regarding checking gl_version, I guess we better not check nothing, not at least yet...
As far as I know, almost all the GL implementations are automatically offering the latest compatibility GL context (i.e. 3.1 in case of only core-profile systems). And we can manually enforce a newer one just simply setting gl-version in Config.prc
Maybe in the future we can think about implementing another logic to generate the GL contexts. Maybe a core profile flag at Config.prc??

panda/src/glstuff/glCgShaderContext_src.cxx
sanguinariojoe: OK! Fixed!

sanguinariojoe:
_core_profile is initialized and managed by GLGraphicsStateGuardian... I cannot find an inherited class eventually overloading reset() without calling itself.
For instance, glxGraphicsStateGuardian is just simply not overloading such method, while wglGraphicsStateGuardian is overloading it, but the first thing it is doing is calling GLGraphicsStateGuardian::reset();

Both methods,  get_supports_luminance_texture() and has_core_profile(), are conveniently used right now

panda/src/glstuff/glCgShaderContext_src.cxx
```diff
@@ -944,30 +944,32 @@ update_shader_vertex_arrays(ShaderContext *prev, bool force) {
...
-            break;
-          case CA_vertex:
-            glDisableClientState(GL_VERTEX_ARRAY);
-            break;
-          case CA_normal:
-            glDisableClientState(GL_NORMAL_ARRAY);
-            break;
-          case CA_color:
-            glDisableClientState(GL_COLOR_ARRAY);
+          if(_glgsg->has_fixed_function_pipeline()) {
```
rdb: Like before, this check can also go away, since it shouldn't get assigned to CA_anything in the core profile case anyway.

sanguinariojoe: Fixed!

Is this one OK?

```
        if (_glgsg->has_fixed_function_pipeline() && (p == 0)) {
          // NOTE: if we disable attribute 0 in compatibility profile, the
          // object will disappear.  In GLSL we fix this by forcing the vertex
          // column to be at 0, but we don't have control over that with Cg.
          // So, we work around this by just binding something silly to 0.
          // This breaks flat colors, but it's better than invisible objects?
          _glgsg->enable_vertex_attrib_array(0);
          if (bind._integer) {
            _glgsg->_glVertexAttribIPointer(0, 4, GL_INT, 0, 0);
          } else {
            _glgsg->_glVertexAttribPointer(0, 4, GL_FLOAT, GL_FALSE, 0, 0);
          }

        }
```

I have not a CG context to check

sanguinariojoe:
Rebased!

rdb:
Thanks a lot!  I've squash merged it with some changes of my own (particularly removing redundant checks).

PR #129, opened 2016-11-27
Do not throw an assertion when processing a light source shader input…

Moguri:
… that does not have a light

Instead, return an identity matrix. This allows shaders to use an
arbitrary size for p3d_LightSource instead of needing to know the number
of lights ahead of time. Extra lights should not affect the final
shading result.


rdb:
I decided to just implement this with the proper values that gl_LightSource would give in this case.  This follows the defaults specified on page 61 of the OpenGL 2.1 spec.

This makes light source 0 have a default white diffuse colour, which makes sure that a scene without lighting shows up as expected, making emulation of FFP behaviour easier.

PR #130, opened 2016-11-27
Do not call expensive BulletDebugNode::sync_b2p() if the node is not in a live scenegraph

Moguri:
I have some code [here](https://github.com/Moguri/prototype-lithium/blob/master/game/lithium/components.py#L211) that attempts to toggle Bullet debug drawing on/off. To my surprise, the very expensive BulletWorld::sync_b2p() was still being called even though the BulletDebugNode was not visible (or even in the scenegraph).

I've switched my code to use explicit show()/hide() calls, which prevents the debug code from running and killing my frame time. However, this PR may still be useful to others.

### DIFF
--- a/panda/src/bullet/bulletDebugNode.cxx
+++ b/panda/src/bullet/bulletDebugNode.cxx
@@ -180,7 +180,7 @@ draw_mask_changed() {
 void BulletDebugNode::
 sync_b2p(btDynamicsWorld *world) {
 
-  if (is_overall_hidden()) return;
+  if (is_overall_hidden() || !is_under_scene_root()) return;
 
   nassertv(get_num_geoms() == 2);
 


rdb:
Hmm, I'm not intimately familiar with `is_under_scene_root()`, but I'm aware of Panda games that have a scene graph for Bullet physics without using it for rendering.  I believe this change would cause these applications to stop working.

rdb:
Thanks a lot for the contribution!

The reason I feel a bit reluctant about this PR is because the implementation of `is_under_scene_root()` is slow and hacky: it recursively iterates through the parents, and then iterates over all the windows and their regions on the main GraphicsEngine to see whether any of them have that top node set as root, which could quickly exponentially expand for a nested scene graph in a RenderPipeline programs.  (I don't have any numbers to test the performance, though.  If it turns out not to matter much, I wouldn't object to this PR as a stopgap measure.)

I think a better way to fix this is to instead do the debug drawing via `cull_callback`, in the same way that CollisionNode does it (creating viz nodes and traversing those) or by rendering the created geoms via `add_for_draw`.  This way we are assured that Panda will only call it when it's about to render the debug node.

PR #132, opened 2017-01-13
Add hidden window mode

Moguri:
This PR implements an enumerated type for window modes. Support is added for the existing window mode properties as W_regular, W_fullscreen, W_undecorated. Fullscreen and undecorated support is handled by delegating to the existing properties. We can look into removing these to reclaim bits in the Specified bitfield and make the window modes truly mutually exclusive (right now some odd combinations can be set). However, for now I tried to keep the changes to a minimum to reduce the risk of bugs.

A new W_hidden WindowMode is implemented for X11 to hide a window from the window manager. Support for macOS and Windows is still TODO.

The changes were tested using a modified version of the asteroids sample.
[window_mode_test.zip](https://github.com/panda3d/panda3d/files/703519/window_mode_test.zip)

This PR is more of a request for review and feedback than a request for a merge at this point in time.

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -99,7 +99,7 @@ move_pointer(int device, int x, int y) {
 
   if (device == 0) {
     CGPoint point;
-    if (_properties.get_fullscreen()) {
+    if (_properties.get_window_mode() == WindowProperties::W_fullscreen) {
       point = CGPointMake(x, y + 1);
     } else {
       point = CGPointMake(x + _properties.get_x_origin(),
@@ -147,7 +147,7 @@ begin_frame(FrameMode mode, Thread *current_thread) {
   CGLLockContext((CGLContextObj) [cocoagsg->_context CGLContextObj]);
 
   // Set the drawable.
-  if (_properties.get_fullscreen()) {
+  if (_properties.get_window_mode() == WindowProperties::W_fullscreen) {
     // Fullscreen.
     CGLSetFullScreenOnDisplay((CGLContextObj) [cocoagsg->_context CGLContextObj], CGDisplayIDToOpenGLDisplayMask(_display));
   } else {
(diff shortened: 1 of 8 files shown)


Moguri:
Here is a completely untested patch for Windows support: https://hastebin.com/sopavezume.diff

Moguri:
And here is a completely untested patch for macOS/Cocoa: https://hastebin.com/kasivubemi.diff

rdb:
This is good.  I like this.

We should probably display a deprecation warning when people attempt to use `set_fullscreen()` and `set_undecorated()`.  Perhaps we should just drop the flags entirely but keep `set_fullscreen(true)`, etc. as a backward-compatible way to set the window mode.  Otherwise it will get a mess if there are two settings that do the same thing.  I don't mind breaking API compatibility as long as the only the most ubiquitous idioms still work.

I think there should probably be a `window-mode` config variable to set the default setting.

Moguri:
I have rebased this branch to reflect changing the GraphicsWindows to use window_mode instead of fullscreen and undecorated. X11 is handled by this PR. I have win32 changes in a branch (window_mode_win), but they are still untested (have not even been compiled). Cocoa is still TODO.

Moguri:
I have a window_modes_cocoa branch that compiles on build-osx, but is otherwise untested.

Moguri:
I pulled all of the branches together and rebased them on top of master. I have pushed the branch to both this repo and my fork. I can create a new PR off of the panda3d/window_modes branch.

Moguri:
This requires unit tests before being merge-ready. We need to test the new window_mode and make sure we didn't break the old API (get/set_undecorated(), get/set_fullscreen(), etc.).

Also, deprecation warnings should be added to the old API.

Moguri:
I am tempted to drop this PR. I no longer need these changes anymore, and I am not sure if/when I will put in the energy to push this PR to completion. Thoughts?

Moguri:
I am abandoning this PR. Someone else can feel free to pick it up again.

PR #133, opened 2017-01-25
Fix typo of include guard.

bluekyu:
This fixes typo of include guard in parametricCurveCollection.h file.

I found compile error when I included both nodePathCollection.h file and parametricCurveCollection.h file.


rdb:
Thanks!

PR #134, opened 2017-01-28
Fixed an exploit in ServerRepository which allowed clients to send in…

AnythingTechPro:
…complete packets which would result in the buffer raising an assertion error, because it couldn't unpack the specified values correctly.


Moguri:
Could you remove all of the white space changes to reduce the noise of this patch? Also, is this related to the issue brought up in PR116? The descriptions look similar, but I haven't analyzed the code changes.

rdb:
Yes, please keep formatting changes and functional changes separate.

PR #135, opened 2017-02-08
Directtools fixup

fireclawthefox:
Fixed a bunch of errors and made some enhancements within the direct tool set.
Also made the scene editor usable again for some small prototyping, testing and debugging models.

A short summary of what has been done:
- Enhanced scene graph explorer with new parts and better resizeability
- moved some useful features from scene editor to directtools (e.g. loading animations in the animPanel)
- Made the scene editor mostly work again
  - Cleaned up most of the old/deprecated Panda3D features from the scene editor
  - Removed some old scene editor scripts by replacing them with directtool alternatives
- Fix and handle many exceptions esp. from tk and pmw usage


direct/src/tkwidgets/EntryScale.py
```diff
@@ -483,16 +483,27 @@ def setNodePathColor(color, np = nodePath, cb = callback):
...
     menu = esg.component('menubar').component('EntryScale Group-menu')
+    menubar = vgp.component('menubar')
+    menuName = "EntryScale Group"
     # Some helper functions
     # Clear color
     menu.insert_command(index = 1, label = 'Clear Color',
                         command = lambda np = nodePath: np.clearColor())
+    menubar._menuInfo[menuName][1].append("")
+    menubar._menuInfo[menuName][1][2] = menubar._menuInfo[menuName][1][1]
+    menubar._menuInfo[menuName][1][1] = ""
```
rdb: Could you explain what this bit of code does?

fireclawthefox: As this was already addressed in irc, here the short answer just to have it documented where it should be.

This fixes a "list index out of range" exception that occurs in thePmwMenuBar code whenever the user hovers over some of the menu entries in the tools.

rdb: I'm not really happy with seeing this unexplained hack repeated all over the code.  Ideally we'd find out why this is needed and fix it elsewhere, or move it to a central location somewhere where it can be changed and documented easily (and taken out if it's no longer needed).

contrib/src/sceneeditor/seSession.py
```diff
@@ -208,36 +211,36 @@ def oobe(self):
...
+                name = 'manipulateCamera',
                 blendType = 'easeInOut')
             # When move is done, switch to oobe mode
             t.uponDeath = self.beginOOBE
 
     def beginOOBE(self, state):
         # Make sure we've reached our final destination
-        self.oobeCamera.iPosHpr(self.cameraControl.camManipRef)
-        self.camera = self.oobeCamera
+        self.oobeCamera.setPosHpr(self.cameraControl.camManipRef.getPos(), self.cameraControl.camManipRef.getHpr())
```
rdb: The correct way to replace iPosHpr in this instance is:
`self.oobeCamera.setPosHpr(self.cameraControl.camManipRef, 0, 0, 0, 0, 0, 0)`

This handles the case correctly where the two nodes in question have different parents.

contrib/src/sceneeditor/dataHolder.py
```diff
@@ -693,13 +693,17 @@ def getAnimationDictFromActor(self, actorName):
...
         ###########################################################################
-        animContorlDict = self.ActorDic[actorName].getAnimControlDict()
+        animControlDict = self.ActorDic[actorName].getAnimControlDict()
         animNameList = self.ActorDic[actorName].getAnimNames()
         if len(animNameList)==0:
             return {}
         animDict = {}
         for anim in animNameList:
-            animDict[anim] = animContorlDict['lodRoot']['modelRoot'][anim][0]
+            print animControlDict
```
rdb: This breaks compatibility with Python 3.  I assume you did not mean to leave this line in?

rdb:
Also, could you motivate the removal of some of the `se*.py` files, for the record?

fireclawthefox:
The reason behind removing some se*.py files was because they became unnecessary. Effectively they just doubled the code for no real reason anymore. At least from the comment (one of them "If we do change original directools, it will force user has to install the latest version of OUR Panda") it seemed to me it was useful back in the days when the scene editor was not part of the official Panda3D, but as that's now the case, no one would need to install/build another version of Panda just to get the SceneEditor to work properly.

As for the other requests, I'm with you on them and will change them asap as well as looking further into the _menuInfo problems to see if there's a better fix.

Moguri:
@fireclawthefox Are you still interested in pursuing this PR? If not, or if I don't here from you in two weeks, I intend to close this PR. It can always be re-opened.

fireclawthefox:
@Moguri I am still interested in bringing those tools back to a state worth using. Though I'm not sure if this stuff in here is to much outdated already seeing all the conflicting files. Maybe it's worth redoing everything that has been done if still needed and open a new PR. Or should this one better be updated?

Moguri:
@fireclawthefox Do whatever works best for you. It may be worthwhile to at least take a look at the conflicts to see how bad they are. I think either approach works about the same for us. Should we close this PR in the mean time?

fireclawthefox:
@Moguri ok, I'll check what will work best then. Not sure when I'll get to it, but in the meantime I'd prefer to keep this PR open as it will remind me to do this once. If it gets closed I probably will forget this someday.
@rdb it may be useful to know when it's a good time to do such probably rather large changes with all the cleanup and fixing stuff. Otherwise we may end up again having all this laying around for a long time not being integrated into master and become outdated again.

rdb:
That's fine, I just marked it as Draft until it's ready for review again.

Moguri:
Closing due to inactivity for now. Please feel free to reopen this PR if/when you resume work on this again.

PR #138, opened 2017-03-08
Is is faster than ==

PendulumDreams:


PendulumDreams:
This didn't do what I wanted.

PendulumDreams:
I want to merge all my files into a single commit first.

direct/src/cluster/ClusterClient.py
```diff
@@ -145,7 +145,7 @@ def controlObjectTask(self, task):
 
     def sendNamedMovementDone(self, serverList = None):
 
-        if (serverList == None):
+        if (serverList is None):
```
rdb: While you're at it making style changes, perhaps you could also remove the unnecessary parentheses in `if` statements like this one.

rdb:
For future reference, GitHub can automatically squash multiple commits when merging a pull request.

PR #139, opened 2017-03-09
Changes to help speed up the python code

PendulumDreams:
Figured out in the last couple commits how to squash the commits when editing directly on github.  The first couple were not.

Still have some other changes which I will do later.


PendulumDreams:
It appears that editing directly through github caused some strange character to get added.  Tried to remove one and after committing the character is still there.  Going to see if I can edit through a different browser with different results.

CFSworks:
Just commenting to make sure everyone reviewing is aware to use extreme caution. `is` is indeed faster than `==` because the former tests that the underlying `PyObject *`s are the same (i.e. that it's the same object) while the latter will first perform the identity test and THEN invoke `__eq__` (to test if their contents are identical - e.g. with strings)

This means `is` only offers a speedup in cases where the objects *aren't* the same. It also means `is` may return false for certain equal objects.

iirc None, True, False, and all built-in types are safe, as the only copies of these objects are created at runtime init. Strings are dangerous unless they're interned and I have no idea on the subject of ints.

I haven't reviewed this PR, and maybe I'm espousing stuff everyone knows already. Just thought this deserved a warning label before it got a look-over.

jayden-emergence:
I'm the one who submitted this.  Adding a comment from my work account.

If I counted correctly for the "is" changes there are:

Number | Expression
---------|-----------
308 | is None
214 | is not None
9 | is dict
2 | is not dict
16 | is str
8 | is not str
4 | is list
1| is tuple
2 | is int
2 | is float
1 | is types.ModuleType
5 | is types.MethodType
1 | is types.FunctionType
1 | type(self) is type(other)
1 | is types.ClassType<br /> (which is commented out)

There are some like `is type('')` or `is type(0)` but I changed them to str or int in another part of the pull request and counted it as str or int above.

And then 3 which would blow up in python 3 before and after the change because the types module does not contain them anymore.  I've got fixes for these, just didn't have time to submit them yet.

Number | Expression
---------|-----------
1 | type(classDef) is not types.ClassType and type(classDef) != types.TypeType
1 | type(classObj) is types.ClassType or type(classObj) is types.TypeType
1 | type(object) is types.InstanceType

direct/src/distributed/ServerRepository.py
```diff
@@ -266,19 +266,19 @@ def readDCFile(self, dcFileNames = None):
...
                 self.notify.debug("No class definition for %s." % (className))
             else:
-                if type(classDef) == types.ModuleType:
+                if type(classDef) is types.ModuleType:
                     if not hasattr(classDef, className):
                         self.notify.error("Module %s does not define class %s." % (className, className))
                     classDef = getattr(classDef, className)
 
-                if type(classDef) != types.ClassType and type(classDef) != types.TypeType:
+                if type(classDef) is not types.ClassType and type(classDef) != types.TypeType:
```
CFSworks: This appears to have been a missed case

direct/src/interval/IntervalManager.py
```diff
@@ -133,7 +133,7 @@ def __doPythonCallbacks(self):
     def __storeInterval(self, interval, index):
         while index >= len(self.ivals):
             self.ivals.append(None)
-        assert self.ivals[index] == None or self.ivals[index] == interval
+        assert self.ivals[index] is None or self.ivals[index] == interval
```
CFSworks: I don't know if separate intervals can be equal. This might be another good candidate for `is`

rdb: Note that any class that is wrapped by interrogate *must* use `==` since two classes with the same underlying C++ pointer may still not match via `is`.

I would suggest leaving this case alone if we're not sure.

rdb:
In addition to what @cfsworks stated, there may be cases where comparison to None could be different from an identity test, namely where the object in question has overloaded __eq__ to make None comparison meaningful.  I'm not aware of any classes for which this is the case.

That said, I am inclined to accept this PR, against my better judgment, because it's relatively harmless and good style to use 'is' for constants and built-in types even though whatever performance benefits will be obtained here are unlikely to amount to much.

Thank you for tallying up the expressions.  Most seem safe, except that I'm not 100% sure that `type(self) is type(other)`and `type(self) == type(other)` are entirely equivalent in all cases.

As I stated before, perhaps this PR would be a good opportunity to also fix the `if` expressions that unnecessarily have extra parentheses.

PendulumDreams:
You guys know the underlying C code where I do not.  Accept whatever you want, I'm only trying to help speed it up a bit.

@CFSworks 
The ClassType line you mentioned in direct/src/distributed/ServerRepository.py would actually blowup in python 3 because types does not contain ClassType, TypeType, or InstanceType like python 2 did.  There are other ways to test for these in python 3.  The classType and TypeType in python 3 can be tested with:
```if type(classDef) is not type:```  Not sure if this is compatible in 2 but I don't have a version of 2 installed at home so I can't check.

And instance type can be made 2/3 compatible with:
```
try:
  from types import InstanceType
except ImportError:
  InstanceType = object
```

It is definitely possible that I may have missed a few files or places in files.  I'm trying to write my own python code to sit on top of all the panda3d C code so I stuck panda3d's python in my own svn.  I then started going through and trying to find things that could be done in a better way and eventually just decided instead of converting panda3d code over to the format I want to end up as I would just write new files from scratch.  So I have some more changes that I haven't submitted yet but they are more specific changes to certain areas in the code.

Like this:
```(("HalfPixelShift" in self.configuration) == False)```

Can't remember where that is in the code but it should be changed to this:
```"HalfPixelShift" not in self.configuration```
It doesn't need to test that it equals False since x in x will always return true or false and you just add not to reverse it.

I'll get a new pull request for those things probably on the weekend or some point after that.  And I'll fix the types that no longer exist in a way that is 2 and 3 compatible (I'll test it out at work first).

rdb:
I apologise but I'm going to go back on my earlier comments and close this for now.  A lot of work remains before this can be merged; a lot of cases have been missed, and if we're going to make such sweeping style changes to the source code I would prefer we make other style fixes as well (such as changing unnecessary parentheses around many of the if statements).

Also, changing the `__call__` checks to `callable` is not without side-effects since it behaves differently with old-style classes.

Furthermore, without seeing measurements from real applications I have strong doubts this is going to have any impact on performance, whereas it does significantly impact mergeability.  I'm happy to entertain smaller patches to micro-optimise python code in hot loops, however, if a convincing argument is made that the performance impact is non-negligible.

PR #140, opened 2017-03-22
Add properties to Bullet classes

wolfgangp:
Note: The Panda Bullet API allows setting steering in degrees on the BulletVehicle, or in radians on its BulletWheels.

BulletVehicle basic functions before (still works!):

self.vehicle.set_steering_value(steering, 0)
self.vehicle.set_brake(brake_force, 2)
self.vehicle.apply_engine_force(engine_force, 3)

Works now and may be future recommendation:

self.vehicle.wheels[0].steering = steering
self.vehicle.wheels[2].brake = brake_force
self.vehicle.wheels[3].engine_force = engine_force


panda/src/bullet/bulletWheel.cxx
```diff
@@ -202,7 +202,7 @@ get_wheel_radius() const {
 void BulletWheel::
 set_steering(PN_stdfloat value) {
 
-  _info.m_steering = (btScalar)value;
+  _info.m_steering = (btScalar)deg_2_rad(value);
```
rdb: This seems like a functional change, one that also breaks compatibility.  Could you explain it?

rdb:
Thanks a lot!

PR #143, opened 2017-04-02
Maya 2017 support

sean5470:
Added entries to compile Maya 2017 plugins.


rdb:
Thanks, I have pushed both commits, after fixing the whitespace errors in the bam2egg one.

sean5470:
Thanks for pointing out the space/tab issue.

I've adjusted my notepad++ to accommodate so shouldn't be an issue in the future. ;)

rdb:
Addendum: the latest buildbot builds now contain Maya 2017 tools:
https://www.panda3d.org/download.php?sdk&version=devel

PR #144, opened 2017-04-15
panda: Get rid of few warnings

deflected:
- Correct missing override
        - Correct deprecated XKeycodeToKeysym
        - Correct deprecated @exec for FreeBSD pkg plist

Signed-off-by: deflected <deflected@users.noreply.github.com>

### DIFF
--- a/panda/src/egg/eggLine.h
+++ b/panda/src/egg/eggLine.h
@@ -31,7 +31,7 @@ PUBLISHED:
 
   virtual EggLine *make_copy() const OVERRIDE;
 
-  virtual void write(ostream &out, int indent_level) const;
+  virtual void write(ostream &out, int indent_level) const OVERRIDE;
 
   INLINE bool has_thick() const;
   INLINE double get_thick() const;
@@ -39,7 +39,7 @@ PUBLISHED:
   INLINE void clear_thick();
 
 protected:
-  virtual int get_num_lead_vertices() const;
+  virtual int get_num_lead_vertices() const OVERRIDE;
 
 private:
   double _thick;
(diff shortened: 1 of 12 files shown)


panda/src/x11display/x11GraphicsWindow.cxx
```diff
@@ -2051,7 +2051,7 @@ get_keyboard_map() const {
       continue;
     }
 
-    KeySym sym = XKeycodeToKeysym(_display, k, 0);
+    KeySym sym = XkbKeycodeToKeysym(_display, k, 0, 0);
```
rdb: Hmm, are you absolutely sure that this doesn't require linking with an additional library?  That's the main reason I was reluctant to make this change.

panda/src/egg/eggLine.h
```diff
@@ -31,15 +31,15 @@ class EXPCL_PANDAEGG EggLine : public EggCompositePrimitive {
 
   virtual EggLine *make_copy() const OVERRIDE;
 
-  virtual void write(ostream &out, int indent_level) const;
+  virtual void write(ostream &out, int indent_level) const OVERRIDE;
```
rdb: You can feel free to use lowercase `override` since all the compilers we target nowadays support it.

panda/src/glstuff/glShaderContext_src.h
```diff
@@ -40,7 +40,7 @@ class EXPCL_GL CLP(ShaderContext) FINAL : public ShaderContext {
   void reflect_uniform(int i, char *name_buffer, GLsizei name_buflen);
   bool get_sampler_texture_type(int &out, GLenum param_type);
 
-  INLINE bool valid(void);
+  INLINE bool valid(void) OVERRIDE;
```
rdb: Unfortunately MSVC 2010 does not support placing `override` on inline functions.  This is a compiler bug, but we do compile with MSVC 2010 at the moment.

panda/src/x11display/x11GraphicsWindow.cxx
deflected: It is part of the X11, Xkb is just an extension ... it may not present on very old X servers.

panda/src/egg/eggLine.h
deflected: Just keeping it with already existing code style.

PR #145, opened 2017-04-17
direct/showbase: fixed an issue in Audio3DManager, where node path's …

AnythingTechPro:
…that are empty aren't removed from the audio manager.

### DIFF
--- a/direct/src/showbase/Audio3DManager.py
+++ b/direct/src/showbase/Audio3DManager.py
@@ -262,6 +262,15 @@ class Audio3DManager:
             tracked_sound = 0
             while tracked_sound < len(self.sound_dict[known_object]):
                 sound = self.sound_dict[known_object][tracked_sound]
+
+                if known_object.isEmpty():
+
+                    # remove the object from the sound dict since the object's
+                    # node path is empty; and it's no longer being used.
+                    del self.sound_dict[known_object]
+
+                    return Task.cont
+
                 pos = known_object.getPos(self.root)
                 vel = self.getSoundVelocity(sound)
                 sound.set3dAttributes(pos[0], pos[1], pos[2], vel[0], vel[1], vel[2])


direct/src/showbase/Audio3DManager.py
```diff
@@ -262,6 +262,15 @@ def update(self, task=None):
...
             while tracked_sound < len(self.sound_dict[known_object]):
                 sound = self.sound_dict[known_object][tracked_sound]
+
+                if known_object.isEmpty():
+
+                    # remove the object from the sound dict since the object's
+                    # node path is empty; and it's no longer being used.
+                    del self.sound_dict[known_object]
+
+                    return Task.cont
```
rdb: I don't think it should break out of updating all of the sounds if only one of them happens to have been removed.  It should instead continue the next iteration of the for loop.

rdb:
So, I am guessing, the underlying issue here is that you called `removeNode()` on a `NodePath` that you passed into `attachSoundToObject`, and it then triggered an error in `update()`?

If so, it seems that a better solution might be not to remove it (since there may still be other NodePaths in existence pointing to that node), but rather, for `attachSoundToObject` to make a copy of the NodePath that is passed in, like so:
```python
object = NodePath(object)
```
(Since a NodePath is only a handle to a node, copying the NodePath wouldn't actually copy the underlying node.  It would just create a separate reference that wouldn't lose its reference when `removeNode` is called on the other one.)

That said, I might be wrong about how people expect `removeNode` to behave.

What do you think?

AnythingTechPro:
Well, when calling "removeNode" assuming you want to remove that object; the sound component attached to that node should be removed as well. I'm thinking maybe the user should call some sort of method maybe "destroy" on the sound component which will then destroy the NodePath it is parented to as well, this would be a more straight forward solution.

rdb:
Consider, however, that `removeNode` does nothing to delete the node; it is merely equivalent to `detachNode` followed by `clear`.  NodePaths are passed by value and as such a `clear` is equivalent to setting a variable to `None` in Python.  All other NodePaths pointing to the node will continue to exist.  As such, it is not at all intuitive to me that `clear()` should have any effect on sounds playing on a node.

I am leaning towards thinking that copying the NodePath is the right solution, but I will give an opportunity for others to weigh in.

rdb:
I checked in a fix to this.  I realise that keeping the node alive is undesirable, so I checked in a change that uses weak references and automatically removes the sound from the Audio3DManager when the underlying node is deleted.

PR #147, opened 2017-04-23
came_case functions for some direct classes

wezu:
The file changes have been made with an automated tool and then hand checked - I didn't find error but I'm also dyslexic.

### DIFF
--- a/direct/src/actor/Actor.py
+++ b/direct/src/actor/Actor.py
@@ -2560,10 +2560,10 @@ class Actor(DirectObject, NodePath):
 
     #snake_case alias:
     control_joint = controlJoint
-    set_lodanimation = setLODAnimation
+    set_lod_animation = setLODAnimation
     get_anim_control_dict = getAnimControlDict
     get_actor_info = getActorInfo
-    clear_lodanimation = clearLODAnimation
+    clear_lod_animation = clearLODAnimation
     reset_lod = resetLOD
     fix_bounds = fixBounds
     get_anim_filename = getAnimFilename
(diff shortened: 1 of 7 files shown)


darktohka:
Some of these functions are only available under __debug__, so importing these files without __debug__ is not possible.

Moguri:
Have you actually tested some of these changes? I think you're missing "self." in most (all?) of your assignments.

rdb:
These are at the class level, so use of `self.` would not be appropriate.
Adding these to the instance dict would be unnecessarily inefficient.

However, it would be great if the functions that are defined conditionally
on `__dev__` could be fixed, since this is breaking (as Travis indicates)

Moguri:
Ah, I didn't notice they are at the class level.

direct/src/actor/Actor.py
```diff
@@ -2549,3 +2557,92 @@ def renamePartBundles(self, partName, newBundleName):
...
+    get_overlapping_joints = getOverlappingJoints
+    enable_blend = enableBlend
+    face_towards_viewer = faceTowardsViewer
+    bind_anim = bindAnim
+    set_blend = setBlend
+    get_frame_time = getFrameTime
+    remove_node = removeNode
+    wait_pending = waitPending
+    expose_joint = exposeJoint
+    set_lodnode = setLODNode
```
rdb: Should be set_lod_node

direct/src/actor/Actor.py
```diff
@@ -2549,3 +2557,92 @@ def renamePartBundles(self, partName, newBundleName):
...
+    fix_bounds = fixBounds
+    get_anim_filename = getAnimFilename
+    get_subparts_complete = getSubpartsComplete
+    verify_subparts_complete = verifySubpartsComplete
+    get_play_rate = getPlayRate
+    clear_python_data = clearPythonData
+    load_anims = loadAnims
+    set_subparts_complete = setSubpartsComplete
+    draw_in_front = drawInFront
+    get_lodnode = getLODNode
```
rdb: Should be get_lod_node

rdb:
Thanks!

PR #149, opened 2017-05-09
Fix size_t type warning.

bluekyu:
Fix `size_t` to `int` type-conversion warning on `table->get_num_transforms()`.
(https://github.com/panda3d/panda3d/blob/master/panda/src/gobj/geomVertexData.I#L479)

### DIFF
--- a/panda/src/gobj/geomVertexData.h
+++ b/panda/src/gobj/geomVertexData.h
@@ -200,7 +200,7 @@ private:
                             const unsigned char *from, int from_stride,
                             int num_records);
 
-  typedef pmap<const VertexTransform *, int> TransformMap;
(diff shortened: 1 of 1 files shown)


rdb:
This PR doesn't really solve much, because the implicit conversion now happens in the return value of the function. The indices eventually get cast to int anyway, so there's no real difference whether we store it as `size_t` or `int`.

However, does your compiler not warn about the implicit conversion in the return statement?

bluekyu:
That's right, but it is disturbing in my compiler (Visual Studio 2013, 2015 and 2017.)

However, I found the reason in testing. The warning was reported in 'utility' header file, so I changed the order of header files and the warning disappeared.

PR #150, opened 2017-06-10
__file__ usage in CommonFilters

fireclawthefox:
Removed __file__ usage due to unavailability in deploy-ng builds
Moved Cg filters from sha files into separate python scripts
Removed now obsolete loadShader function from CommonFilters

### DIFF
--- a/direct/src/filter/CommonFilters.py
+++ b/direct/src/filter/CommonFilters.py
@@ -16,6 +16,13 @@ clunky approach.  - Josh
 """
 
 from .FilterManager import FilterManager
+from .filterBloomI import BLOOM_I
+from .filterBloomX import BLOOM_X
+from .filterBloomY import BLOOM_Y
+from .filterBlurX import BLUR_X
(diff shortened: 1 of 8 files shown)


rdb:
Thanks!

wezu:
If loadShader got removed then the snake case alias "load_shader=loadShader" should now also be removed

fireclawthefox:
@wezu the load_shader=loadShader alias is as far as I can see only given in the Loader.py script which doesn't correspond with the removed loadShader from the CommonFilters but instead uses the loadShader from the ShaderPool which is implemented on the C++ side.

wezu:
I think I've added in in CommonFilters.py back when in had a loadShader function, and I see it here now: https://github.com/panda3d/panda3d/blob/deploy-ng/direct/src/filter/CommonFilters.py#L556

fireclawthefox:
ah, yes you're right, that should probably be removed then.

rdb:
Fixed the `loadShader` issue in 624f28f.

PR #151, opened 2017-06-22
Remove call to missing method of AI repository

zachgates:


theclashingfritz:
It's "missing" because it was not originally in the actual Panda3D and was outside of it in the OTP. Which used it. (Or that's the most likely thing. And probably is) There is no point in removing it and you're just making things worse for people who which to use the original file.

rdb:
Panda doesn't actually ship an AI Repository implementation, I believe.

zachgates7:
Yes, it is 'missing'. And no, Panda does not include an implementation of an AI Repository.

I believe the method `startTrackRequestDeletedDO` should be called from the AI Repository implementation (which is not included in Panda, ergo it does not belong here).

darktohka:
It does belong here. Some AI Repository implementations can depend on this function being called.

AnythingTechPro:
Actually zach is right, this is Disney specific and is not used.

zachgates7:
Some AI Repository implementations may 'depend' on this function call, others *are* obstructed by it. This method call had no place in Panda.

If it remains, at the very least, there should be a dummy method with a good docstring telling users to overwrite that method in their AI Repository implementation.

rdb:
Based on the discussion above, I'm inclined to reject this PR, since it is trivial to add a dummy method called `startTrackRequestDeletedDO` to your AI Repository implementation.

Either way, since this class relies on an AI Repository and Panda doesn't provide one, it feels like the entire `DistributedObjectAI` file doesn't belong in the Panda repository and should instead be part of whatever library provides the AI Repository implementation.

rdb:
I'm closing this for now for reasons given above, but I'd be happy to accept a PR that removes the `*AI.py` files from the repository.

PR #152, opened 2017-06-30
Fix BulletHeightfieldShape not using Panda's float precision

Cheaterman:
Everything is in the title - the current code seems to break when building with double precision.

### DIFF
--- a/panda/src/bullet/bulletHeightfieldShape.I
+++ b/panda/src/bullet/bulletHeightfieldShape.I
@@ -30,8 +30,8 @@ BulletHeightfieldShape(const BulletHeightfieldShape &copy) :
   _num_rows(copy._num_rows),
   _num_cols(copy._num_cols) {
 
-  _data = new float[_num_rows * _num_cols];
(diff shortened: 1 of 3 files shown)


Cheaterman:
Actually it runs so I don't know? You tell me, buddi :-D

Le 3 juil. 2017 1:29 PM, "rdb" <notifications@github.com> a écrit :

*@rdb* requested changes on this pull request.

Did you test this? The _data member definition in the header file is still
declared as float pointer, so all this change does is allocate twice as
much memory in a double-precision build, assuming it even compiles.

Furthermore, the _data array is directly passed into
btHeightfieldTerrainShape, whose precision is defined by Bullet, not Panda.
We should rather use a btScalar so that the precision will match what
precision Bullet is compiled with (and is actually expecting in this call).

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/panda3d/panda3d/pull/152#pullrequestreview-47641066>,
or mute the thread

Cheaterman:
That should be all goodie buddi :-) you tell me.

rdb:
Thanks!

PR #153, opened 2017-07-02
Add support for compiling with VS 2017

bluekyu:
I added support for compiling with Visual Studio 2017.

I tested compiling with vs2017 and errors did not happen.
And third-party libraries can be used in vs2015 because VC Runtime of vs2017 is binary compatible with that of vs2015.
(MSDN blog: https://blogs.msdn.microsoft.com/vcblog/2017/03/07/binary-compatibility-and-pain-free-upgrade-why-moving-to-visual-studio-2017-is-almost-too-easy/)

I updated only makepanda tools to find Visual Studio 2017.
However, I tested only vs2015 and vs2017 with Windows 10 and did not test on another platforms.

Note that VS version of vs2017 is 15.0, but the compiler version is 14.1 (vc141).
So, you need to use '14.1' for --msvc-version option.

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -314,10 +314,10 @@ def parseopts(args):
     if GetTarget() == 'windows':
         if not MSVC_VERSION:
             print("No MSVC version specified. Defaulting to 10 (Visual Studio 2010).")
-            MSVC_VERSION = 10
+            MSVC_VERSION = 10.0
 
         try:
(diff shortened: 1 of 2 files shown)


rdb:
Thanks!  I made a few modifications since your changes broke compilation with cygwin and with the Windows 7.1 SDK.  I have also not tested building on a 32-bit system, which might also cause issues.

PR #154, opened 2017-07-05
Fix assert macros that cause a crash in non-debug

bluekyu:
In non-debug, the codes do not run and the deleted item is not removed.
So, a crash happens when hash map is cleared.


rdb:
Nice catch, thanks a lot!

PR #158, opened 2017-08-04
Add additional Windows SDK path to enviroment variable

bluekyu:
Windows 10 SDK binary files are installed in versioned folder from Creator Update (10.0.15063.0)
This fixes the issue that 'rc.exe' file cannot be found.


rdb:
Thanks!

PR #163, opened 2017-08-22
Fix failure to load DLL with unicode path

bluekyu:
Hello. I found the issue that graphics pipe DLL (ex, libpandagl.dll) cannot be loaded when unicode characters exist in the path of the DLL. This RP fixes this issue.

I tested if the issue is resolved on the following systems:
- Windows 10 English version (cp437), C++ & Python
- Windows 10 Korean version (cp949), C++ & Python


rdb:
That's terrific.  Thanks a lot!

PR #165, opened 2017-08-26
Add snake-case function names for 'threading2'-module

kamgha:
This PR makes the `threading2`-module safer to use as a replacement for Python 3s `threading`-module.


rdb:
Thanks.  I've also added a commit that adds `main_thread()`, `get_ident()` and `TIMEOUT_MAX`, which have been added in Python 3.

PR #166, opened 2017-08-31
Introduce png-compression-level PRC variable

kamgha:
This PR adds a _png-compression-level_ PRC variable to Panda. It allows to choose the compression level for writing out PNG images. Valid values are 0 (no compression), or 1 (best speed) to 9 (best compression). The default value is 6 and has been 6 when used in conjunction with libpng+zlib. PNG compression is lossless.

The resulting file size for PNG files could be optimized further by choosing "filters" intelligently as described on http://www.libpng.org/pub/png/book/chapter09.html and in http://www.libpng.org/pub/png/book/LCH-png-chapter.pdf . However, this is out of scope here and several programs like PNGCRUSH and OptiPNG exist that aim to accomplish this task.

For the future, a nice-to-have would be to be able to disable chroma subsampling for writing out JPEG images.


rdb:
This looks great.  Thanks a lot!

PR #168, opened 2017-09-03
Fix inheritance problem

theclashingfritz:
Interrogate likes to cause errors relating to inheritance if a class inherits a class with TypedReferenceCount it's base and another class with TypedObject as it's base.

### DIFF
--- a/panda/src/express/dcast.T
+++ b/panda/src/express/dcast.T
@@ -58,7 +58,7 @@ _dcast(WantType *, TypedObject *ptr) {
     return (WantType *)NULL;
   }
 #endif
-  return static_cast<WantType *>(ptr);
+  return dynamic_cast<WantType *>(ptr);
 }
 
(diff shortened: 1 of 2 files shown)


rdb:
Changing to virtual inheritance is problematic.  What problem is this intended to solve, exactly?

theclashingfritz:
It's supposed to solve a issue within Panda3d where as i said above. You can't inherit a class with TypedReferenceCount as it's base and a class with TypedObject as it's base without causing interrogate to cry in agony due to it trying to call a TypedObject function.

theclashingfritz:
If you have a better solution feel free to let me know

rdb:
It's a bug in interrogate, then.  It probably wouldn't be particularly hard to fix that.  Could you make a header file with such a class that fails in interrogate, and file it as a bug report?  Thanks.

theclashingfritz:
Alright, Will do.

PR #172, opened 2017-09-23
Add has_hook functions taking additional parameter

bluekyu:
Hello. This PR adds `has_hook` functions taking function pointer or callback data in EventHandler.
I needed the functions to check a specific hook, but those did not exist. So, I added them.


rdb:
LGTM, thanks!

PR #185, opened 2017-11-01
Add tests using PyTest

Moguri:
These tests were converted from the tests branch with some changes:

  * Convert from unittest to PyTest
  * Remove tests/README.md and added a section to the main README.md
  * Update code to use new API (LVector3 instead of Vec3, properties, etc.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,9 +34,16 @@ addons:
     - nvidia-cg-toolkit
     - python-dev
     - python3-dev
+    - python-virtualenv
     - zlib1g-dev
     - fakeroot
-script: $PYTHONV makepanda/makepanda.py --everything --git-commit $TRAVIS_COMMIT $FLAGS --threads 4 && LD_LIBRARY_PATH=built/lib PYTHONPATH=built $PYTHONV makepanda/test_imports.py
+install:
+    - virtualenv --python=$PYTHONV venv && source venv/bin/activate
+    - $PYTHONV -m pip install pytest
+script:
+    - $PYTHONV makepanda/makepanda.py --everything --git-commit $TRAVIS_COMMIT $FLAGS --threads 4
+    - LD_LIBRARY_PATH=built/lib PYTHONPATH=built $PYTHONV makepanda/test_imports.py
+    - LD_LIBRARY_PATH=built/lib PYTHONPATH=built $PYTHONV -m pytest tests
 notifications:
   irc:
     channels:
--- a/README.md
+++ b/README.md
@@ -156,6 +156,13 @@ python3.6 makepanda/makepanda.py --everything --installer --no-egl --no-gles --n
 If successful, this will produce a .pkg file in the root of the source
 directory which you can install using `pkg install`.
 
+Running Tests
+=============
+
+Install [PyTest](https://docs.pytest.org/en/latest/getting-started.html#installation) and run `pytest` over the tests directory.
+This can be done by running `pytest` from the tests directory or running `pytests tests` from the project's root directory.
+If `pytest` is run from the project root without specifying the tests directory, PyTest will collect some tests from the `src` directory that we do not want to run.
+
 Reporting Issues
 ================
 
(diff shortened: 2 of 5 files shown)


Moguri:
@CFSworks, I would like to hear your thoughts on this as well. This can be via comments in this PR or over IRC.

rdb:
Great.  I like the relative simplicity of pytest's test cases over unittest.

I think it may be good to use subdirectories to group the tests together.  We could do scene graph tests under `tests/pgraph/` and windowing system related tests under `tests/display/`.  What do you think?

Moguri:
I do not mind grouping these under directories. My plan was to do that as we got more tests, but we can do it now instead.

CFSworks:
I'm a little skeptical of the window tests. Can we have those gracefully skip when running on a platform that has no graphical output (e.g. Travis)? Can we have a second fixture that creates a buffer (offscreen mesa software renderer) on such platforms for testing rendering features, even on Travis?

Otherwise I'm mirroring @rdb here. The biggest challenge in verification frameworks (unit tests, server monitoring, etc.) is in making them so easy and expressive that people don't have an excuse not to use them, so using PyTest here is a very welcome change.

Now we just need to find where that HackPad with the unit testing checklist went. Should we work toward producing test coverage reports automatically?

rdb:
I agree with making the window tests optional, although I wonder if we can use a software implementation or provide a mock window implementation.

@cfsworks: Hackpad became Dropbox Paper, and apparently they botched the automatic migration of that particular document.  Fortunately they offered export to SQL from which I extracted the original text, albeit without formatting, which I inserted into the document's new location:
https://paper.dropbox.com/doc/Unit-test-categories-4fUsbipPOpCWVncD8BS4f

I don't know how easily test coverage can be determined automatically for our codebase, but I'm not opposed to the idea.

CFSworks:
We already target CI compilation with both Clang and GCC. We could ignore the coverage for GCC (since, barring really serious compiler problems, the code path for a given test case should be the same) and use [Clang's built-in instrumenting code coverage tool](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html). Then we just have to collect and print the code coverage summary after the whole unit test suite runs.

The TL;DR is: add a few flags to the Clang build, run the whole unit test suite with an environment variable set so the Clang-injected instrumenting code writes stats to a temp directory, then collect all of the stats files into a single report.

We might want a convenient command to do a coverage test outside of Travis, to help those contributing test cases visualize what they still have to cover.

We will likely not hit 100% without having the tests iterate on several different compilations with different compiler flags enabled. But if we can get upwards of 80% then we can be pretty confident any breaking changes will either be caught right away or cascade down and break a different test.

At any rate - assuming we can make the window-opening fixture skip gracefully and get the pytest things running in .travis.yml - I'd be happy to spend some time this weekend and take a first crack at adding a "coverage" button to the top of this repo's readme.

Moguri:
I have cleaned up the commits on this branch and forced pushed. Travis should work now, also we should expect failures when we get a None GraphicsPipe.

rdb:
@cfsworks: If you can set it up, go for it, thanks!  I'm not quite optimistic we'll be able to reach 80% coverage (given the size of the codebase) or that it will necessarily be representative of how thoroughly the behaviour of Panda is tested, but it can't hurt to have the information available, especially for people who wish to contribute unit tests.

@Moguri: thanks so much!  I think we now only need to set up the directory structure, and then I think we should merge it ASAP.

Given that these files are placed in the `tests` directory, though, is it quite necessary to use the `test_` prefix before all the files?  Are there going to be many Python files that aren't tests?

Moguri:
I believe PyTest auto-collection mechanism looks for the test prefix. There are ways to modify it. There may also be some other prefixes we can use.

Other than setting up the directories, I believe this is ready to merge.

rdb:
This is pushed now.  I added a setup.cfg with `[tool:pytest]` section so we can run pytest from the root.

Thanks a lot to you both!

PR #186, opened 2017-11-02
Fix compile error by missing header file

bluekyu:
Hello, I found a compilation error that occurs when building without Python on the following line:

https://github.com/panda3d/panda3d/blob/075cb14cbbb004fb167632b630d35e8d03e92e03/panda/src/pgraph/shaderInput.h#L142

This PR fixes it.


rdb:
Thanks!

PR #194, opened 2017-11-11
Fixed mirror demo

fireclawthefox:
Set "mirror" camera to always have the same roll as the mirror
-> otherwise the model in the mirror will move unexpectedly
Set y coordinate of the mirror corners to the cameras distance
-> This fixes a crash and keep the mirrored model at correct sizer


rdb:
Great, thanks!

PR #195, opened 2017-11-11
Small extensions for the mirror demo

fireclawthefox:
Add extra parameters for buffer size and clear color
Add sample code at the bottom to display usage of mirror demo


rdb:
Thanks!  I did make a typo fix and changed the order of the arguments.

PR #204, opened 2017-11-28
ShowBase: Minor fixes over aspect ratio, size calculations and want-render2dp crashes

deflected:
- Do not take into accound SBS if it is not enabled
        - Always return value for getSize()
        - Properly handle SBS for window-events
        - Redo positioning of aspect2d(p) markers exactly as
        they are created when adjusting aspect ratio
        - Fixed crash when want-render2dp is set to False


rdb:
Closed in 7a46b2ca60aacc1b8f1c87a7a101546bd6214120 and bdd53d60fc1729ba310251a78d8ae5156a1b0ee7 .  Thanks!

PR #210, opened 2017-12-13
Loader: Fixed crash when canceling async loading of model

deflected:
- Clean up the _loader after we clean up the requests,
        not before that

Signed-off-by: deflected <deflected@users.noreply.github.com>


rdb:
Whoops, not sure how I messed that up.  Thanks!

PR #214, opened 2017-12-16
MAX plugin search line moved to MAX section

sean5470:
Moved the search for MAX .ms files from the Maya section to the MAX section.


rdb:
Thanks!

PR #219, opened 2018-01-01
Fix crash by end iterator while closing a window

bluekyu:
Hello, I found a crash by end iterator that is returned when the removed window is not in `_new_windows`. This PR fixes the crash.


rdb:
Thanks!

PR #223, opened 2018-01-07
Add RenderPipeline C++ lighting modules to Panda3D codebase

rdb:
This PR makes the C++ modules of the RenderPipeline part of Panda3D (so that installing the pipeline will not require building any C++ stuff).  I've only made some minor changes to conform to Panda's coding style.

@tobspr does this seem OK to merge?


tobspr:
@rdb nice, thats pretty cool! Thanks for your effort :)

Moguri:
@rdb and @tobspr What benefits do the RP lights provide over regular Panda lights? Performance? Are there any techniques or lessons learned from this light system that can be used by regular Panda lights? Would you recommend that other render systems use these RP lights? How active is the RenderPipeline project at the moment?

I am asking these questions for my own benefit, and they are not intended to block this PR.

tobspr:
@Moguri I think I added it because of performance reasons and missing features.
First of all, I had the need for a a different shadow system (Shadow Atlas, delayed updates etc.), and also panda did not support all light types (For example, it doesn't have a rectangle light).
Besides of that, I wanted to be able to set the intensity and color temperature on lights and some further PBR based properties. Also IES Profiles for example.

Rdb already added a lot to Panda's lights (and I think he still is) making it more and more similar to the implementation in the render pipeline.

While I'm not adding any features to the render pipeline at the moment, I'm still looking at issues and pull requests regarding the Render Pipeline.

rdb:
@Moguri The main differences I can think of right now is that (1) rplight contains code to set up PSSM and shadow atlases, (2) code to write light changes to a GPU buffer for light culling on the GPU, (3) supports IES datasets.  On the flip side, RP lights have no scene graph representation.

Of course it's not ideal to have multiple lighting data structures, and the intention is for this extra functionality to be incorporated into the core lights.  But it may be a while before that is complete, so the main reason to have this included now is to make it easier for people to use the RenderPipeline without having to go through the process of acquiring the Panda SDK (if they installed from pip) and compiling the RenderPipeline C++ modules.  (In other words, the RenderPipeline becomes a pure-Python module as consequence of this PR.)

Moguri:
Thanks for the info @tobspr and @rdb! I think making popular addons/libraries easier to use is a good use of the contrib directory. This is especially the case when the library is a big part of Panda's current PBR and modern eye-candy story.

PR #236, opened 2018-01-23
Add simple way to enable overflow on DirectEntry element

fireclawthefox:
Made set_overflow_mode avail. in DirectEntry via enableOverflow option
This will enable enable the default overflow behaviour on the DirectEntry element. The position where the overflow should be triggered is set via the DirectEntry's 'width' option.


rdb:
Works like a charm, thanks so much!

PR #246, opened 2018-02-11
Give p3cull a README

CFSworks:
This is a first draft at how our component READMEs should probably look. In particular, it follows this formula:

```
p3COMPONENT: Panda3D's COMPONENT THAT DOES SOMETHING
====================================================

Overview
--------

Ten-foot overview at how the component gets its problem done. Describes interactions between parts. Write this as if someone just asked you "So how do I start coding on this component?"

Classes
-------

* **SomeClass** is a class that does one thing.
* **OtherClass** is another class that needs SomeClass for something.

Configuration
-------------

* `prc-value #t` turns on prc-value

Build-time flags
----------------

* `WANT_SOME_CLASS` ensures that SomeClass is compiled into the engine providing support for SomeClassery.
* `DO_SOME_CLASS_TRACKING` compiles in support for tracking stats on how SomeClass is used.
```

Note that I went **way** overboard with the overview section here. A lot of this belongs on the manual, under "Panda3D Rendering Process" - only the last section ("Deciding rendering order") is really written the way I'd like.

I'm also aware that a lot of the classes I documented aren't even in p3cull but, rather, in p3pgraph. We should probably fix that. Keeping our components nicely cohesive will make it far easier to find what we're looking for most of the time.

This is a pull-request (rather than a direct lob onto `master`) because I'd love to get some feedback and spark a discussion about how in-depth we want to be in our component READMEs. In particular I think they should be highly technical (the target audience is someone reading the C++ code, after all) but I'm open to suggestions about how detailed we go. Again, I went overboard with "overview," since I'm not sure if it's supposed to document _intended_ behavior or _actual_ behavior. The manual seems to go the way of _intended_ behavior - that is, what Panda3D promises to do for the application developer. Component-level READMEs should probably instead have a focus on the component's contract with the rest of the engine.

(P.S. Even though this is a _pull_ request, it's probably cleaner to cherry-pick it if everyone likes it rather than merge it. It's only one commit after all, so it's cleaner.)

### DIFF
--- a/panda/src/cull/README.md
+++ b/panda/src/cull/README.md
@@ -1,3 +1,170 @@
-This package contains the Cull Traverser.  The cull traversal collects
-all state changes specified, and removes unnecessary state change
-requests.  Also does all the depth sorting for proper alphaing.
+p3cull: Panda3D's cull traverser
+================================
+
+Overview
+--------
+
+### What is "culling"?
+
+In the 3D graphics world, "culling" refers to the process of determining
+objects that are hidden from view or are otherwise guaranteed not to contribute
+meaningfully to the scene. These objects are then skipped in the rendering step
+to save on valuable rendering time.
+
+Panda3D uses the term "cull" in a slightly different sense. In particular, it
+refers to the whole process of preparing a complex, non-linear and arbitrarily
+deep scenegraph for rendering by the graphics backend, which needs a list of
+objects to render one at a time. This involves, roughly:
+
+1. Determining all *visible* objects, given a viewpoint, a scenegraph, and a
+   bitmask indicating which objects should be shown.
+2. *Composing* (combining) the state of each object, as inherited from its
+   ancestors in the scenegraph.
+3. Deciding the order in which the objects are to be sent for rendering.
+
+Note that this processs should not be confused with "back-face culling," which
+is where triangles that face away from the screen are assumed to be on the
+opposite side of of an object and therefore not visible. Back-face culling is
+performed at render-time and therefore does not reduce the total number of
+objects being sent to the backend.
+
+### Determining visible objects
+
+Panda3D's primary culling mode is called "viewing frustum culling," where a
+given viewpoint (Camera and Lens, in Panda3D terms) is represented as a
+bounding volume, typically a frustum - a chopped-off pyramid. As the theory
+goes, whenever a volume represents completely and precisely the visible portion
+of a space, any object that does not intersect with the volume must not be
+visible and can therefore be ignored during that frame.
+
+Every object also has a bounding volume, which is a geometric approximation of
+its actual shape. This approximation is computed to be as small as possible,
+while staying computationally simple and guaranteeing that the volume
+encompasses the entire object and its children. In other words, it's better to
+err on the side of being too large.
+
+When Panda3D performs cull traversals, it walks the scenegraph in a depth-first
+search order, visiting each node, and testing for intersections between the
+view volume and the node's bounding volume. These intersections usually fall
+into one of three categories:
+1. No intersection: The node is culled, and none of its children are
+   considered.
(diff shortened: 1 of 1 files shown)


CFSworks:
Definitely most/all of my overview section here belongs in a new manual chapter. Feel free to grab it and put it where appropriate. It wasn't until halfway through that I remembered the manual was a thing.

I suppose the main shift I'm trying to instigate here is a bigger focus on cohesion - perhaps even modularity - and clearer statement of purpose for each component. The component level READMEs should probably be more of a "guided tour of the codebase" than an API reference (which belongs in the, you know, API reference) or explanation of engine concepts (which belongs in the manual - perhaps an "internals" chapter is in order)

The READMEs should probably have as much descriptiveness as necessary for someone to create a full graph (think .dot) of engine interdependencies by hand. I like the class list's emphasis on interaction and external behavior, whereas the API reference tends to have a bigger emphasis on internal behavior.

I'd say that the question I'm trying to answer here is "What class do I have to subclass if I want to add my own rendering ordering to the engine?" and not "What does CullHandler do?" - the goal here is to help a new developer, who may have lots of 3D experience but no familiarity with the engine codebase, get up to speed quickly.

I'm a bit busy today and don't think I can work on a second draft, but perhaps the 3 sections we should have are:

1) Purpose. Who uses this? What for? Is it exposed to applications or to other Panda components? (Helps maintainers identify dead components like with our cftalk confusion. Helps newcomers know where their new code belongs.)
2) Explanation. If I asked you "How does your car's camshaft work?" I'm not expecting a mere list of part numbers for each cam and/or datasheets on them. ("Each one pushes a lever at some point during one engine cycle." doesn't help me work on your car.) I'm mainly wondering about the in-situ purpose of each cam and how it fits with the others. ("These four open the fuel intakes and are timed to open the cylinder for one quarter cycle half a cycle before ignition, these four are for exhaust, ...") This helps since a newcomer or even intermediate user doesn't want to read through a class-to-purpose listing in the API reference to find the part of the pipeline that does what they want.
3) Build variables. This might belong somewhere else, I'd concede. But for thorough testing we should have some reference about how the engine changes shape when built differently, so developers who make a change to a component know which build options they need to test with.

But as you pointed out, a lot of this can be autogenerated. We might make the rule that the class list is just the first paragraph from each class's doc. That kills two birds with one stone: forces the class doc comments themselves to state their interdependencies and purpose (not just their behavior), and gives us an easy and maintainable way of keeping per-component documentation.

I like the sphinx generator pulling the README. We should still figure out what that README is. (Just the "overview" section? Maybe plus a more friendly name so it shows up as "Collison handling" rather than "p3collide"?) But yeah, assuming the generator can also grab relevant preprocessor macros, that seems like it addresses every concern.

BMaxV:
I think this deserves some mention somewhere, but not in code, is there a "to be put into the manual" section somewhere?

ArsThaumaturgis:
> I think this deserves some mention somewhere, but not in code, is there a "to be put into the manual" section somewhere?

I believe that there's a separate issue-tracker for documentation--see here:
https://github.com/panda3d/panda3d-docs/issues

Moguri:
@CFSworks should we close this PR in favor of a PR for the Manual?

rdb:
This PR is now nearly 3 years old without activity on it since, so I need to close this.  The actual content here is great, but I think it ought to be in the manual, because it describes higher-level subsystem and not specifically the organisation of code inside the panda/src/cull directory (in fact, much of the cull system is in pgraph, and to a smaller degree in display).  I think it would be a disservice to hide this information in the depths of the source tree.

Feel free to open a new issue or PR on the docs repo regarding the cull system documentation.

I'm not opposed to having README files in the tree describing the nitty-gritty code-organisation details but I don't think they should duplicate information that is in the API reference or docstrings, since doing so increases the maintenance burden.

PR #248, opened 2018-02-15
Bullet samples

Schwarzbaer:
Taken from https://github.com/wolfgangp/panda3d_bullet_samples with permission from the author.


JoelStienlet:
Hello everybody,
I'm wondering why these examples haven't been merged yet.
Best regards.

TheCheapestPixels:
Additional changes requested:
* Remove the commented out tinydisplay stuff from the tops of the files
* Naming of the samples should be lowercase

samples/bullet-samples/01_basics.py
```diff
@@ -0,0 +1,181 @@
+#from panda3d.core import load_prc_file_data
```
rdb: This file should have UNIX newlines.

Also, every file here should have a shebang:
```
#!/usr/bin/env python
```

samples/bullet-samples/27_heightfield_vehicle.py
```diff
@@ -0,0 +1,686 @@
...
+            sync-video #t
+            textures-power-2 none
+            ###gl-coordinate-system default
+            notify-level-gobj warning
+            notify-level-grutil debug
+            notify-level-shader_terrain debug
+            notify-level-bullet debug
+            ### model paths
+            model-path /usr/share/panda3d
+            model-path /home/juzzuj/code/prereq/bullet-samples/bullet-samples
```
rdb: These should not be here (most of those variables don't belong here).

samples/bullet-samples/05_filtering_extended.py
```diff
@@ -0,0 +1,190 @@
...
+        ShowBase.__init__(self)
+        base.set_background_color(0.1, 0.1, 0.8, 1)
+        base.set_frame_rate_meter(True)
+
+        base.cam.set_pos(0, -20, 4)
+        base.cam.look_at(0, 0, 0)
+
+        # Light
+        alight = AmbientLight('ambientLight')
+        alight.set_color(LVector4(0.5, 0.5, 0.5, 1))
```
rdb: We can just use a tuple to set the color, we don't need to use LVector4 for that.  We can remove the import to LVector4 too, then.

samples/bullet-samples/04_filtering.py
```diff
@@ -0,0 +1,205 @@
...
+    def toggle_debug(self):
+        if self.debugNP.is_hidden():
+            self.debugNP.show()
+        else:
+            self.debugNP.hide()
+
+    def do_screenshot(self):
+        base.screenshot('Bullet')
+
+        # ____TASK___
```
rdb: Why is this comment here?

samples/bullet-samples/02_shapes.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def cleanup(self):
+        self.world = None
+        self.worldNP.remove_node()
+
+        #def doAdded(self, node1, node2):
+        #  print 'added:', node1.getName(), node2.getName()
+
+        #def doDestroyed(self, node1, node2):
+        #  print 'destroyed:', node1.getName(), node2.getName()
```
rdb: Either we should remove these two commented out functions if they have no use, or we should fix them to conform to the coding style (such as Python 3-compatible print functions).

TheCheapestPixels: Since they're informative about the bullet API, I elect to keep and update them.

BMaxV:
This looks like a great resource, but is probably too much to review all at once. Maybe this is something we can integrate into the roadmap, reviewing file by file? Also there maybe should be a common "resource" folder, so the line count doesn't get increased by the .egg files.

Moguri:
@rdb It looks like all of your requested changes have been marked as resolved. Are there still known issues with this PR?

Moguri:
@Schwarzbaer would you mind pulling in the latest `master` (either via a merge or a rebase)? That will re-trigger the stuck/broken CI and maybe include some CI fixes from `master` as well.

rdb (approved):

PR #249, opened 2018-02-16
mathutil: Update FFTCompressor for FFTW3

CFSworks:
This has been due for a while. The last FFTW 2.x release was in 1999.

Note that this does change some of the loops; this has two benefits:
1) The halfcomplex storage order is now explained with a comment.
2) It fixed the special case "don't break a run of bytes for a zero" which
   was never triggering due to the value not being *exactly* 0.0.

I have tested these changes against older FFT-compressed animation .bams
and no noticeable decompression changes are present, so a .bam version
bump is not necessary.

----------

This is a pull request because I haven't actually touched serious Panda3D code yet, and would really appreciate some guidance and critiques on what I did right and not. If this is good, just squash/cherrypick it onto master. :)

I also noticed that `compress-channels` has been deprecated. Why is this? It seems like good code - and I have seen it reduce animation sizes by a factor of 6x in my unscientific testing. Is it simply a lack of maintenance or should we look at GPL-free alternatives to FFTW (e.g. [FFTS](https://github.com/anthonix/ffts/))? We could consider giving mathutil a `FastFourierTransform` class (used by `FFTCompressor`) which also exposes FFT planning/execution functionality to the application layer, which would prove useful in music games, networked games that use Fourier compression of player movement, etc.

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -648,8 +648,7 @@ if (COMPILER == "MSVC"):
     if (PkgSkip("HARFBUZZ")==0):
         LibName("HARFBUZZ", GetThirdpartyDir() + "harfbuzz/lib/harfbuzz.lib")
         IncDirectory("HARFBUZZ", GetThirdpartyDir() + "harfbuzz/include/harfbuzz")
-    if (PkgSkip("FFTW")==0):     LibName("FFTW",     GetThirdpartyDir() + "fftw/lib/rfftw.lib")
-    if (PkgSkip("FFTW")==0):     LibName("FFTW",     GetThirdpartyDir() + "fftw/lib/fftw.lib")
+    if (PkgSkip("FFTW")==0):     LibName("FFTW",     GetThirdpartyDir() + "fftw/lib/fftw3.lib")
     if (PkgSkip("ARTOOLKIT")==0):LibName("ARTOOLKIT",GetThirdpartyDir() + "artoolkit/lib/libAR.lib")
     if (PkgSkip("OPENCV")==0):   LibName("OPENCV",   GetThirdpartyDir() + "opencv/lib/cv.lib")
     if (PkgSkip("OPENCV")==0):   LibName("OPENCV",   GetThirdpartyDir() + "opencv/lib/highgui.lib")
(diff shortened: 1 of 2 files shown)


rdb:
We can't build the public builds with fftw because of the GPL license.  As I had understood that this was an esoteric feature that probably nobody used, so I had deprecated it on "try-it-and-see" basis to see if anyone would complain about it.  (So far, noone has, reinforcing my belief that it was a vestigial feature and could probably be removed.)

It sounds like you're saying there's a genuine use case for having this, and there is serious gain to be had from it, so I'm fine with having it un-deprecated and updated to FFTW 3.  It would be nice to have some unit/regression tests to prevent code rot, though.

The code itself looks fine assuming that it works and doesn't break backward compatibility.

As an aside, if we're interested in mesh compression we might also consider implementing existing mesh compression libraries such as https://github.com/google/draco.

theclashingfritz:
I had personally always wondered why FFTW was deprecated but hearing it was because of it's license explains everything. It does the compression quite well but i just never got around to ever bringing this up as while it was deprecated, the code was still there. So i never had a point to do so,

JoelStienlet:
- A paper about FFTS with some benchmarks:
[](https://www.cs.waikato.ac.nz/~ihw/papers/13-AMB-IHW-MJC-FastFourier.pdf)
the paper is from 2012, but the library is still maintained, and at that time its performance was better than fftw's, so my guess is that it's still efficient.
Another benchmark from 2014:
[](https://github.com/anthonix/ffts/issues/28)

- The FFTS library seems reasonably portable (x64, arm >v7 with NEON)

- Looks like the interfaces of FFTW and FFTS are quite similar, FFTS seems even simpler:

```
FFTW_EXTERN X(plan)                                                     \
FFTW_CDECL X(plan_dft_r2c_1d)(int n,R *in,C *out,unsigned flags); 

FFTW_EXTERN void                                                        \
FFTW_CDECL X(execute_dft_r2c)(const X(plan) p, R *in, C *out);    

FFTW_EXTERN X(plan)                                                     \
FFTW_CDECL X(plan_dft_c2r_1d)(int n,C *in,R *out,unsigned flags); 

FFTW_EXTERN void                                                        \
FFTW_CDECL X(execute_dft_c2r)(const X(plan) p, C *in, R *out);  
```
for FFTS the nature of the transform r->c or c-> is defined by the "sign" parameter:

```
FFTS_API ffts_plan_t*
ffts_init_1d_real(size_t N, int sign);

FFTS_API void
ffts_execute(ffts_plan_t *p, const void *input, void *output);
```

theclashingfritz:
I'd recommend implementing both and just leaving FFTW for backwards compatibility if it's scrapped in the end, and if you do leave it for that i'd recommend a method for easily converting FFTW compressed animations to FFTS ones. Just to make it easier on all of us.

rdb:
I would assume that the results created by FFTS are the same as the results created by FFTW, so I would be in favour of switching to FFTS and never looking back.

theclashingfritz:
Yeah just leave the backwards compatibility for FFTW so we can make sure the FFTW compressed stuff can be converted properly. I'd love to see FFTS just for the fact it'd be non-deperacted and the new Panda3D standard.

rdb:
You're not getting what I'm saying.  I'm saying that FFTS and FFTW implement the same transform, so there is no need for any conversion, so there is no need to keep FFTW around.

theclashingfritz:
Oh really? Well if that's the case i agree then. But i'd make sure and test it before hand, There might be edge-cases.

CFSworks:
Since this change already went through the trouble of implementing FFTW2's halfcomplex format, there's no longer any specific dependency on FFTW - I could roll my own DFT by hand and the results would be the same (just unbearably slow).

As long as FFTS has comparable-or-better performance, which Joel seems convinced of, preserving FFTW support makes no sense at all. The only thing I'm still pushing for is a specific class that exposes FFT planning and transformation via Interrogate, both because it makes testing through pytest easy and because apps could leverage the FFT to do some creative things.

PR #251, opened 2018-02-17
pgraph: Fix alignment error on NodePath::set_shader_input

lachbr:


rdb:
Sorry, I had a fix to this ready on my computer, but I'd forgotten to check it in.  I've just done that.

Thanks!

PR #252, opened 2018-02-18
Use self instead of the class name in AsyncRequest.

Nickdoge:


rdb:
Could you describe what problem this PR is intended to solve?

I imagine that the use of AsyncRequest is by design, so that specifically the AsyncRequest notify category and not a possible subclass' notify category is used, although I'm not personally sure which behaviour makes more sense in that case.

Nickdoge:
I was looking in most of Panda3D's python code (thought it would be a good idea to change it, as most of panda3d's code is using self, it would look better). However, it does not fix anything.

rdb:
Thanks for your contribution.  However, I'm going to close this as it doesn't fix anything and doesn't constitute a significant style improvement.

PR #259, opened 2018-02-26
Fix menubar issues on macOS

treamology:
On macOS, I've noticed that the menubar (including the Apple menu) is unresponsive after launching a Panda app. Switching away from the window and back fixes the unresponsiveness, but there are still no items under the application menu. Since it is common for a user to want to quit the app through that menu or through the Cmd+Q keyboard combination, I addressed that as well.

The unresponsive menubar seems to stem from where `[NSApp activateIgnoringOtherApps]` is called. From my research, Cocoa is oddly specific about where certain methods need to be called from, and behavior often differs between each OS release. I moved this call to an AppDelegate event, which is where Apple seems to recommend placing it. I also added a quit button to the application's menu, along with the corresponding keyboard combination.

In the future, we may want to add a way to customize menubar items from the application itself, rather than a developer having to modify the engine source.

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -30,6 +30,7 @@
 
 #import "cocoaPandaView.h"
 #import "cocoaPandaWindow.h"
+#import "cocoaPandaAppDelegate.h"
 
 #import <ApplicationServices/ApplicationServices.h>
 #import <Foundation/NSAutoreleasePool.h>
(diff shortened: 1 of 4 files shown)


rdb:
This looks like a good change, thank you.

I wasn't able to reproduce the menubar unresponsiveness, but I'll take your word for it that it fixes it in your case.

I normally would be a bit reluctant to bind shortcuts by default, but in this case Cmd+Q is as common on macOS as Alt+F4 is on Windows, which is also enabled by default, and pretty much every application out there has a quit button in the application menu.

Issue #262, opened 2018-02-27
Developer-friendly build options


SkyLeach:
Having developed on Windows, Linux and OSX/Darwin I can say with certainty that developers often need to have conflicting versions of tools available to them.

This issue is both a discussion and a linkbase for PRs related to small changes that help facilitate making panda's build process flexible enough to, at the least, not impede this process.

SkyLeach:
PR #263 adds the option `--use-homebrew` to instruct panda to use `brew --prefix` to find the `python Framework/<version>` instead of assuming system default paths (which may work, but break virtual environments).

SkyLeach:
Added homebrew thirdparty options for opus (opusfile, libogg), vrpn and fftw

I think it might be awesome to allow --use-homebrew to override the thirdparty tarball for packages detected using `pkg-info` (default) or `os.popen('brew --prefix pkgname')`.

This will make building on macos much easier to keep up to date with new libraries, especially the math libraries.

rdb:
Is there a way to have homebrew to write `.pc` files, which are pretty much the de facto standard for package discovery?  Then makepanda will not require any modifications in order to be able to pick up homebrew packages.

As an aside, I'm not really fond of the option being called `--use-homebrew`, since it looks too much like the `--use-`/`--no-` package options, which would lead one to expect that `--everything` and `--nothing` options should effect it as well.

Note that if you want to easily build the thirdparty packages yourself in a location where makepanda can find them, you can check out [this repository](https://github.com/rdb/panda3d-thirdparty) into the "thirdparty" directory in the source of the root, and use the provided CMake script to fetch and build the packages.  You can adjust the versions of the packages in question in the CMakeLists.txt.

SkyLeach:
##pkg-config

---
**TL;DR**

Explict calls to homebrew are more about explicit and controlled change than pkg-config.  pkg-config remains the preferred and mostly untouched method of finding third party libs.  pkg-config doesn't work for all packages, however, and I didn't want to change makepanda.py's calls to SmartPkgEnable in ways that would break either ./thirdparty/<systemarch> or existing thirdparty source builds (at least not until after testing and discussion).  In general, homebrew is much more up-to-date than other build systems, requiring extensive changes to what makepanda looks for.

---

[...]

PR #263, opened 2018-02-27
Added kwargs for using homebrew instead of system root to find python on darwin systems

SkyLeach:
This change updates the options with the flag --use-homebrew

This flag only currently take effect in makepandacore.SdkLocatePython and only for GetTarget() == 'darwin'

Assuming those conditions are met then the external command `brew --prefix` is used to retrieve the homebrew prefix location.  This location is used by homebrew to install the python Frameworks in a way that doesn't conflict with system installations of python.

#262 - OSX homebrew python linking


Moguri:
I do not like that this requires a change to every package. Instead, could we check brew as an automatic fallback if we do not find a third-party directory for the library in question and the user is on darwin?

rdb:
I appreciate the work on this, but since this makes substantial changes to makepanda and we're trying to transition to CMake, I'm going to close this for now.  I recommend people who need more flexibility with package location to use CMake.

TV4Fun:
How's that transition to CMake going?

PR #267, opened 2018-03-02
Clean up OpenAL (and fix some Apple bugs)

CFSworks:
The real meat and potatoes of this changeset is in the first two commits; those address some assumptions that the OpenAL plugin makes about the OpenAL implementation's behavior (which are, evidently, guaranteed with OpenAL Soft, but not so much on Apple OpenAL).

The remaining 4 commits are just making the OpenAL code much more robust, including adding more assertions to verify state and generally making the intended behavior of the code much clearer.

Since these are pretty substantial changes, I'm submitting them as a PR so the following can be heavily scrutinized:
* I used `assert` instead of the `nassert` variants, for consistency with the one `assert` that was already there. What's the engine core's standards on asserts? Should these be `nassertv` etc.?
* How did I do with my PRC variables? Are they named succinctly/consistently/clearly?
* The `while (true) {` construct is okay, yes? (Specifically I did *not* make it a `for` so in the common case of "the delete works the first time" there's no performance wasted querying the PRC variables, etc.)
* Did I make proper use of the .I file for inlines?
* Are my additions to the class declarations sorted neatly?
* Are my comments properly-formatted, clear, and correct?
* Does this need tests? (Or *can* this be tested?)

And otherwise please do nitpick and test heavily. I'd rather have strict standards that I have to meet to develop the right coding habits early on. :)

### DIFF
--- a/panda/src/audiotraits/config_openalAudio.cxx
+++ b/panda/src/audiotraits/config_openalAudio.cxx
@@ -30,6 +30,23 @@ ConfigVariableString openal_device
  PRC_DESC("Specify the OpenAL device string for audio playback (no quotes).  If this "
           "is not specified, the OpenAL default device is used."));
 
+ConfigVariableInt openal_buffer_delete_reattempts
+("openal-buffer-delete-reattempts", 5,
+ PRC_DESC("If deleting a buffer fails due to still being in use, the OpenAL "
+          "sound plugin will wait a moment and reattempt deletion, with an "
+          "exponentially-increasing delay for each attempt.  This number "
+          "specifies how many repeat attempts (not counting the initial attempt) "
+          "should be made before giving up and raising an error."));
+
+ConfigVariableDouble openal_buffer_delete_delay
+("openal-buffer-delete-delay", 0.001,
+ PRC_DESC("If deleting a buffer fails due to still being in use, the OpenAL "
+          "sound plugin will wait a moment and reattempt deletion, with an "
+          "exponentially-increasing delay for each attempt.  This number "
+          "specifies how long, in seconds, the OpenAL plugin will wait after "
+          "its first failed attempt.  The second attempt will be double this "
+          "delay, the third quadruple, and so on."));
+
 
 /**
  * Initializes the library.  This must be called at least once before any of
--- a/panda/src/audiotraits/config_openalAudio.cxx
+++ b/panda/src/audiotraits/config_openalAudio.cxx
@@ -30,21 +30,21 @@ ConfigVariableString openal_device
  PRC_DESC("Specify the OpenAL device string for audio playback (no quotes).  If this "
           "is not specified, the OpenAL default device is used."));
 
-ConfigVariableInt openal_buffer_delete_reattempts
-("openal-buffer-delete-reattempts", 5,
+ConfigVariableInt openal_buffer_delete_retries
+("openal-buffer-delete-retries", 5,
  PRC_DESC("If deleting a buffer fails due to still being in use, the OpenAL "
-          "sound plugin will wait a moment and reattempt deletion, with an "
-          "exponentially-increasing delay for each attempt.  This number "
-          "specifies how many repeat attempts (not counting the initial attempt) "
+          "sound plugin will wait a moment and retry deletion, with an "
+          "exponentially-increasing delay for each try.  This number "
+          "specifies how many repeat tries (not counting the initial try) "
           "should be made before giving up and raising an error."));
 
 ConfigVariableDouble openal_buffer_delete_delay
 ("openal-buffer-delete-delay", 0.001,
  PRC_DESC("If deleting a buffer fails due to still being in use, the OpenAL "
-          "sound plugin will wait a moment and reattempt deletion, with an "
-          "exponentially-increasing delay for each attempt.  This number "
+          "sound plugin will wait a moment and retry deletion, with an "
+          "exponentially-increasing delay for each try.  This number "
           "specifies how long, in seconds, the OpenAL plugin will wait after "
-          "its first failed attempt.  The second attempt will be double this "
+          "its first failed try.  The second try will be double this "
           "delay, the third quadruple, and so on."));
 
 
--- a/panda/src/audiotraits/openalAudioSound.I
+++ b/panda/src/audiotraits/openalAudioSound.I
@@ -40,7 +40,7 @@ get_calibrated_clock(double rtc) const {
  *
  * Returns true on success, false on failure.
  */
-bool OpenALAudioSound::
+INLINE bool OpenALAudioSound::
 require_sound_data() {
   if (_sd==0) {
     _sd = _manager->get_sound_data(_movie, _desired_mode);
@@ -61,7 +61,7 @@ require_sound_data() {
  * so.  The `force` argument overrides this, indicating we don't intend to
  * reacquire the sound data.
  */
-void OpenALAudioSound::
+INLINE void OpenALAudioSound::
 release_sound_data(bool force) {
   if (!has_sound_data()) return;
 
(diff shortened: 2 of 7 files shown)


CFSworks:
> I am somewhat nervous about the idea of stalling the thread for 1-32 milliseconds. That's an awful lot of time and would definitely cause noticeable chugs. Is this only going to happen rarely?

Rarely, and I'd be surprised if all 5 reattempts were necessary. I'd guess that 1 or 2 reattempts (1ms or 3ms delay, respectively) are sufficient. The 5 cutoff is just a hail mary to save the audio backend from a leak (or corruption error) in case something's really taking forever to unlock. We could probably lower the default delay if it's still a concern. In case you're thinking of tweaking the values yourself, the way to cause the bug (on master, not this branch) on Apple is to play()-then-immediately-stop() a few sounds per frame. You should get a "corruption in stream queue" error because the buffers remain on the sources due to Panda not waiting for OpenAL to be done with them, and then they get unqueued by a different sound using the same source from the pool.

> The prc variables look fine to me. If I had to nitpick I'd have called it "retries" instead of "reattempts", the former word being used in other places in Panda.

I want the nitpicks. "retries" it shall be.

> You need to use the INLINE keyword in front of the definitions as well as the declarations. Not all compilers mind, but some will complain. (The inline keyword allows the function to violate the one-definition-rule.)

Ah, okay, the lack of INLINE on \*_sound_data() threw me off.

> We shouldn't use assert(). Panda's nasserts are designed to be able to gracefully exit the function so that interrogate can detect that an assertion has happened and relay it to Python via an AssertionError.

Roger that. (When we get around to a linter I wonder if we could/should straight-up ban the use of `assert()`)

> As an aside, if the Apple version is so buggy, we should strongly consider simply using OpenAL Soft on all platforms.

It's not so much that it's *buggy* - I looked at the Creative Labs programmer's handbook, and it doesn't mention either of the behaviors I'm compensating for here as strictly against the OpenAL specification. It's more that OpenAL Soft seems to be much more careful about behaving in an intuitive way - which makes it tons more robust to use, but not super great to test against. (Put another way, I'd like to fight for Apple OpenAL support, chiefly because it'd be a good exercise in cleaning up Panda's OpenAL code, which has always made me pretty uneasy.)

That, and OpenAL Soft is LGPL, and I like not having to ship an extra dynamic library when building for Apple. ;)

CFSworks:
Oh, before I head off for the night: another option would be to make the reattempt delay constant rather than exponential. "Try up to 200 times with 0.1ms between tries." That way it never stalls for more than really needed (and again, it'd probably unlock within 2ms anyway.)

CFSworks:
@rdb I pushed a few commits addressing your feedback. Could you do another review? :)

It also occurs to me that a unit test *is* possible if we only test on OpenAL Soft, which has an extension supporting loopback devices that simply render samples into an offline buffer. (The bigger question of "do we want pytest to root around in ctypes just to grab such a buffer and skip if we're running on another vendor's OpenAL" set aside.)

panda/src/audiotraits/openalAudioSound.I
```diff
@@ -55,11 +55,48 @@ require_sound_data() {
...
     _sd = 0;
   }
 }
+
+/**
+ * Checks if the sound has NOT been cleaned up yet.
+ */
+INLINE bool OpenALAudioSound::
+is_valid() const {
+  return _manager != NULL;
```
rdb: I personally tend to use `nullptr` in all newly added code, but I don't know if you would find this to be inconsistent.

panda/src/audiotraits/openalAudioSound.I
```diff
@@ -55,11 +55,48 @@ require_sound_data() {
...
+/**
+ * Checks if the sound is playing.  This is per the OpenALAudioManager's
+ * definition of "playing" -- as in, "will be called upon every update"
+ *
+ * This is mainly intended for use in asserts.
+ */
+INLINE bool OpenALAudioSound::
+is_playing() const {
+  // Manager only gives us a _source if we need it (to talk to OpenAL), so:
+  return _source != 0;
```
rdb: Should be `NULL` or `nullptr` to make clear that this is a pointer.

panda/src/audiotraits/openalAudioSound.I
```diff
@@ -55,11 +55,48 @@ require_sound_data() {
...
+}
+
+/**
+ * Checks if the sound has its SoundData structure open at the moment.
+ *
+ * This is mainly intended for use in asserts.
+ */
+INLINE bool OpenALAudioSound::
+has_sound_data() const {
+  return _sd != 0;
```
rdb: See above.

panda/src/audiotraits/config_openalAudio.cxx
```diff
@@ -30,6 +30,23 @@ ConfigVariableString openal_device
...
+ConfigVariableInt openal_buffer_delete_retries
+("openal-buffer-delete-retries", 5,
+ PRC_DESC("If deleting a buffer fails due to still being in use, the OpenAL "
+          "sound plugin will wait a moment and retry deletion, with an "
+          "exponentially-increasing delay for each try.  This number "
+          "specifies how many repeat tries (not counting the initial try) "
+          "should be made before giving up and raising an error."));
+
+ConfigVariableDouble openal_buffer_delete_delay
+("openal-buffer-delete-delay", 0.001,
```
rdb: Have you experimented to see which value is enough here?  I don't feel strongly about this at all, but if 0.2 ms is good enough for most cases, I'd use that as a starting point.

panda/src/audiotraits/openalAudioSound.I
CFSworks: I'll stick with consistency for now. We should replace all of our NULL with nullptr en masse later on.

panda/src/audiotraits/openalAudioSound.I
CFSworks: This one's `ALuint`, not a pointer.

panda/src/audiotraits/openalAudioSound.I
CFSworks: Alright, I'll make that one a `NULL`

panda/src/audiotraits/config_openalAudio.cxx
CFSworks: I'll do some experiments before merging this down. I have a working minimal test case that demonstrates the bug. I'll have it run some timing loops to determine the average-case and worst-case buffer unlocks.

CFSworks:
> I take it that using alSourcei(source, AL_BUFFER, 0) to detach the buffer from the source is not enough in this case, then? I don't really know much about OpenAL; I'm sure you've studied this issue in more detail, so I'll defer to your opinion on this.

`alSourcei(source, AL_BUFFER, 0)` is explicitly mentioned in the official spec as the recommended way to reset the source for reuse and free all buffers it was using. The spec says nothing about _when_ this happens - though I think it's implied that it should be "immediately," which means Apple's in the wrong here.

> An alternate solution might be to instead queue up the buffers in a separate garbage vector and find some moment to periodically garbage collect old buffers, which would not stall the thread. But if this issue happens rarely as you say, it's probably not worth the trouble to implement.

The trouble is it's not just buffers that would need queueing, as the sources they're stuck on would also have to be taken out of the pool until they're cleared.

I think this only happens if you're starting a source, then immediately stop it, since at that point the source is still in the (internal/Apple-only) `kTransitionToPlay` state, and the playback thread has to give up the buffer before the app thread can safely reclaim it.

In my testing it seems like about 0.2% of the time when stopping a sound, the buffer won't unlock right away. But it does require about 16ms-21ms worth of sleeps to wait for it. (Then again this is with very large command queues and only one source; it's a synthetic loop and not representative of real-world activity like Panda would have.)

PR #273, opened 2018-03-05
direct: Fix some NameErrors

mfwass:
- Removed undeclared variable `rogerroger` in MetaInterval.
- It appears that the intended functionality of ObjectPool.containerLenStr was to print out the keys of `_len2obj`.  I've implemented this by looping through lengths.

### DIFF
--- a/direct/src/interval/MetaInterval.py
+++ b/direct/src/interval/MetaInterval.py
@@ -342,7 +342,6 @@ class MetaInterval(CMetaInterval):
     # with all of their associated Python callbacks:
 
     def setManager(self, manager):
-        rogerroger
         self.__manager = manager
         CMetaInterval.setManager(self, manager)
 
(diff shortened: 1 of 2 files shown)


rdb:
Judging by the name `containerLenStr` and the fact that it's building a string `s`, it's actually supposed to return a string rather than print it.

That said, this method has been broken since [its inception](https://github.com/panda3d/panda3d/commit/6d4def41b7542e26ab1db7ced0069c395b2e4ee2#diff-c1db2e566f4d3ec302a7ce984001be94R107), so I think the right thing to do is to remove it.  I'm assuming it's not being used by anything, right?

mfwass:
@rdb Sounds good to me.  I was unsure of its exact functionality, but that was my best guess as what it was intended to do.

We aren't making any use of it, just happened to stumble across it.

PR #274, opened 2018-03-06
direct: Fix some more NameErrors

mfwass:
Fixed a typo inside threading and PyDatagramIterator.  Both typos ahve been here for 10+ years, interesting that they've never been caught.

@rdb Do you think that we should just remove `BoundedSemaphore`?  It has been broken since it was added, so I doubt it's used anywhere.


mfwass:
👍

rdb:
Thanks for the changes!

I don't think we should remove BoundedSemaphore, as this module is supposed to mirror the threading primitives that Python's standard library provides.

mfwass:
Sounds good to me! That's why I wanted to ask, wasn't sure what you were thinking we should do with it. :)

PR #275, opened 2018-03-07
Update direct.dc sample

thetestgame:
Updated the direct.dc sample file to support the latest code base by adding in missing distributed classes and removing invalid keywords. This file should now function as is with the current Panda3D distributed code as well as Astron.

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -4494,13 +4494,14 @@ if (GetTarget() == 'windows' and PkgSkip("GL")==0):
 # DIRECTORY: panda/src/egldisplay/
 #
 
-if (PkgSkip("EGL")==0 and PkgSkip("GLES")==0 and PkgSkip("X11")==0):
+if (PkgSkip("EGL")==0 and PkgSkip("GLES")==0):
   DefSymbol('GLES', 'OPENGLES_1', '')
   OPTS=['DIR:panda/src/egldisplay', 'DIR:panda/src/glstuff', 'BUILDING:PANDAGLES',  'GLES', 'EGL']
   TargetAdd('pandagles_egldisplay_composite1.obj', opts=OPTS, input='p3egldisplay_composite1.cxx')
   OPTS=['DIR:panda/metalibs/pandagles', 'BUILDING:PANDAGLES', 'GLES', 'EGL']
   TargetAdd('pandagles_pandagles.obj', opts=OPTS, input='pandagles.cxx')
-  TargetAdd('libpandagles.dll', input='p3x11display_composite1.obj')
+  if not PkgSkip("X11"):
+    TargetAdd('libpandagles.dll', input='p3x11display_composite1.obj')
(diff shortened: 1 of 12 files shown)


rdb:
@cfsworks, you know more about distributed than I do, could you take a look at this?

theclashingfritz:
This looks fine. More up to date on the class types.

direct/src/distributed/direct.dc
```diff
@@ -80,12 +82,32 @@ dclass DistributedSmoothNode: DistributedNode {
...
+  int16/10 h;
+  int16/10 p;
+  int16/10 r;
+  string state;
+};
+
+dclass DistributedCamera : DistributedNode {
+  setCamParent(uint32) required broadcast ram ownsend airecv;
+  setFixtures(Fixture []) required broadcast ram ownsend airecv;
+};
```
CFSworks: No newline at end of file

direct/src/distributed/direct.dc
```diff
@@ -29,18 +28,21 @@ dclass DistributedObject {
   setLocation(uint32 parentId, uint32 zoneId) broadcast ram;
 };
 
+dclass DistributedObjectGlobal : DistributedObject {
+};
+
 dclass TimeManager: DistributedObject {
-  requestServerTime(uint8 context) p2p;
+  requestServerTime(uint8 context) aircv clsend;
```
CFSworks: Typo of `airecv`

Also, `TimeManager` works in a very particular way with `ServerRepository`; while airecv/clsend is a more Astronic way of doing things, I don't know if this will break CMU's server (which I don't know if people use)

direct/src/distributed/direct.dc
```diff
@@ -3,14 +3,13 @@
 // your own project (or load it from the direct source tree) and build
 // on it with your own dc file for your own classes.
 
-keyword broadcast;
-keyword ram;
-keyword p2p;
```
CFSworks: Removing the other forward-declarations of keywords is fine, however `p2p` is not a builtin keyword and carries specific meaning with CMU's `ServerRepository`.

theclashingfritz:
I don't think anybody really even uses the CMU repository anymore. Except for one or two old Toontown server fakers but that's it.

rdb:
People do, in fact, use the built-in server repository to implement networking in their Panda3D games.  We should not break support for it until we have a good replacement and upgrade path.

thetestgame:
I may be wrong but last time I tried the p2p even with the keyword definition still errored parsing the dc file inside panda. It's entirely possible it was due to something I did though

BMaxV:
Since this has been dormant for nearly two years and when it was suggested there were changes / reasoning necessary, I suggest closing this?

rdb:
Is there an intent to update this PR to address the feedback?

rdb:
@thetestgame It looks like you force-pushed this PR out of existence, feel free to open a new PR if you are willing to address the feedback.

PR #279, opened 2018-03-14
doc: WIP style guide

CFSworks:
This is a PR meant to start the ball rolling on what kind of style guide we'd like to see. Nothing in here is final - it was largely written as my "best guess" as to what everybody is most likely to agree with. (Feel free to edit!)

I tried to include the relevant parts from the guide in `panda/src/doc/`, but updated it heavily to match with what the main contributors' habits seem to be.

Note that this guide does _not_ reflect the current state of the codebase - rather, this aims to describe the _desired_ code appearance. After we agree on a guide, I'll get to work on selecting/configuring a good formatter tool (I have my eye on `clang-format`/`clang-tidy`) to help us migrate to the new style and 

Since this is a PR meant for discussion as opposed to inclusion, I've marked everybody (whom I've seen active - apologies if I've left you out) as a reviewer. That being said, I welcome feedback from the peanut gallery as well (as long as it doesn't get too bikeshed, of course).

### DIFF
--- a/doc/hacking/style_guide.md
+++ b/doc/hacking/style_guide.md
@@ -1,0 +1,187 @@
+Panda3D C++ Style Guide
+=======================
+
+Rather than try to justify every decision made, this document attempts to be as
+at-a-glance as possible.  Examples provided where necessary to illustrate best
+practices.
+
+Note that this style guide is, for the most part, not very strict, and there
+are exceptions to every rule.  However, this guide does serve to illustrate the
+preferred style for the entire codebase.
+
+Naming conventions
+------------------
+
+- **Classes**: `CamelCaseFirstUpper`
+- **Filenames**: `lowerCamelCase.h`, `lowerCamelCase.cxx`
+  - Note, every header file must have a unique filename.  They are all
+    installed into the same path.
+- **Functions**: `lower_snake_case_without_prefix`
+- **Local variables**: `lower_snake_case_without_prefix`
+- **Member variables**: `_lower_snake_case_with_initial_score`
+
+Indentation
+-----------
+
+Always use two spaces to indent.  Tabs should never appear in a C++ file.
+
+Access modifiers (`public`/`private`/`protected`/`PUBLISHED`) are never
+indented, and do not affect the indentation of the surrounding lines.
+
+Spaces
+------
+
+Always put one space after a control keyword, before the condition's
+parentheses.  Don't pad the inside of parentheses with spaces unless there are
+several nested parenthetical expressions and the extra spaces contribute
+substantially to readability.
+
+    if (x == y) {
+
+Don't place a space after a function name (declarations, definitions, or calls).
+
+    void function_name();
+    ...
+    function_name();
+
+Always place a space after each comma in an arguments list.
+
+    function_call(x, y, z);
+
+Always pad binary operators with a space on either side.
+
+    a || b
+
+Never end a line with trailing whitespace.
+
+Function definitions
+--------------------
+
+The recipe for a function definition is:
+
+1. JavaDoc comment describing the purpose of the function
+2. Function return value and class
+3. **At the beginning of its own line,** the function's name, followed by the
+   parameter list and anything else necessary for the function's signature.
+4. Opening brace on the same line, not its own line.
+5. Single empty line before next definition.
+
+For example:
+
+    /**
+     * String documenting the function, in JavaDoc style.
+     */
+    void ClassName::
+    function_name(int x, int y) const {
+      ...
+    }
+
+    /**
+     * ...
+     */
+    ...
+
+If the function is a constructor with an initializer list, the `:` goes on the
+same line as the constructor name, and the initializers are listed *one per
+line*, indented by two spaces.
+
+    ClassName::
+    ClassName() :
+      _a("a"),
+      _b("b"),
+      _c("c")
+    {
+    }
+
+Empty function bodies in a .cxx/.I file still contain a single line break (i.e.
+to put `}` on its own line).  In a .h file, an empty function can be given as
+`{}`
+
+Braces
+------
+
+Always put an opening `{` on the same line as the statement that calls for the
+brace.  There is exactly one exception to this: constructors with initializer
+lists.
+
+    SomeClass::
+    SomeClass() :
+      _one(1),
+      _two(2)
+    {
+      if (_one < _two) {
+        ...
+      }
+    }
+
+Comments
+--------
+
+Comments should be properly-formatted English, with proper capitalization and
+punctuation rules.  There are two spaces between sentences.
+
+Note that `//` is preferred over `/*` except in some special circumstances.
+
+Line limits and continuations
+-----------------------------
+
+Lines should be limited to 80 columns.  This is not a hard limit, but if a line
+does go over this length, consider a continuation.
+
+Continuation lines shall be aligned to the same column as appropriate, or
+indented two spaces otherwise:
+
+    if ((a &&
+         b) ||
+	c) {
+      ...
+    }
+
+When choosing where to break a line, these are the preferred places, from best
+to worst:
+
+1. If inside a comment, break the comment to as many lines as appropriate, but
+   only after moving the comment to its own line. (In other words, comments
+   long enough to need 
+2. After a comma in an argument/parameter list.
+3. After a binary operator (leaving the binary operator at the end of the line)
+4. In a function declaration, after the function's return type but before the
+   function's name. (Indented by two spaces.)
+5. In a log statement with `<<`, before a `<<`. (Avoid ending a line with `<<`)
+6. Around a string literal, at a natural word boundary, with a space left on
+   the previous line (e.g. `"This is a "` / `"continued string literal."`)
+7. After the `=` in an assignment operator.
+8. As a last resort, after the opening `(` in a function call.
+
+If none of these rules can apply cleanly, it is okay to have lines extend past
+the 80-column boundary.
+
+Alignment
+---------
+
+TBD, should discuss:
+
+1. Groups of declarations/definitions
+2. Groups of comments
+3. Pointers always being `obj *x;` and never `obj* x;`
+
+Grouping
+--------
+
+Try to group logically-similar lines, separating them with a single blank line.
+
+Modern language features
+------------------------
+
+Panda3D is a C++11 project.  The use of the following modern language features
+is greatly encouraged:
+
+1. `nullptr` over `NULL`
+2. `auto` (when used to store a function return value only)
+3. Range-based for loops
+
+The following modern language features are discouraged:
+
+1. C++17 features
+2. `std::atomic`
+3. Lambda functions


doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+
+    SomeClass::
+    SomeClass() :
+      _one(1),
+      _two(2)
+    {
+      if (_one < _two) {
+        ...
+      }
+    }
```
rdb: This section needs a note about whether to start `else` on its own line or not.

(We currently don't, but I think there's a good argument to be made to put it on its own line.  It tends to make code more readable since the line containing the `}` creates a visual separation between the else condition and the body of the previous if block.  I currently put a blank line at the end of the previous block to achieve the same separation).

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+the 80-column boundary.
+
+Alignment
+---------
+
+TBD, should discuss:
+
+1. Groups of declarations/definitions
+2. Groups of comments
+3. Pointers always being `obj *x;` and never `obj* x;`
```
rdb: I realise this may not be the popular opinion, but I strongly lean towards the former, because this code isn't very obvious:
```
// Is only x a pointer, or are they both pointers?
Object* x, y;
```
It makes it look like the pointer is part of the type, and that x and y are both pointers, but that's not how C++ parses instances.  Putting the decorators where they belong makes it clear that only x is a pointer:
```
Object *x, y;
```

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+
+Don't place a space after a function name (declarations, definitions, or calls).
+
+    void function_name();
+    ...
+    function_name();
+
+Always place a space after each comma in an arguments list.
+
+    function_call(x, y, z);
```
rdb: What about default argument assignments?  Do we want to follow the Python convention of using no spaces around `=` for default arguments, or be consistent with other uses of the `=` keyword and put spaces around it? eg.
```cpp
// Python convention
void function_name(double x=3.0);
// Current convention / consistency
void function_name(double x = 3.0);```

File: doc/hacking/style_guide.md [hunk 4]
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+When choosing where to break a line, these are the preferred places, from best
+to worst:
+
+1. If inside a comment, break the comment to as many lines as appropriate, but
+   only after moving the comment to its own line. (In other words, comments
+   long enough to need 
+2. After a comma in an argument/parameter list.
+3. After a binary operator (leaving the binary operator at the end of the line)
```
rdb: There are good arguments to be made for putting the binary operator at the beginning of the line, since it makes the condition more readable.  Even [Python changed this convention](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) recently.

I would suggest to allow either style, whichever makes code more readable.

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+Always place a space after each comma in an arguments list.
+
+    function_call(x, y, z);
+
+Always pad binary operators with a space on either side.
+
+    a || b
+
+Never end a line with trailing whitespace.
+
```
rdb: I would propose adding something explicitly forbidding [assignment alignment](https://www.reddit.com/r/readablecode/comments/19vkkz/do_you_guys_align_your_code_into_columns/).  While arguably being readable to align a bunch of assignments, it is a pain in the long run to constantly have to realign blocks of variables or function calls whenever something is changed.

CFSworks: That'd be a good thing to put under the "Alignment" section.

For what it's worth, I have a WIP `.clang-format` file (and a script to temporarily edit some of the Panda macros into more standard-looking C++ so clang-format can understand it) located [here](https://gist.github.com/CFSworks/42be608c5ead2251542163e3b97b66b2). It does do a very good job of aligning everything for you - and it has several editor integrations so we can tell people to use that (and perhaps even have Travis fail the PR if it doesn't pass clang-tidy/clang-format checks).

If we want to explicitly _forbid_ it, that's fine too - clang-format also does a great job of unifying everything to a single space.

The thing it doesn't seem to support is our odd habit of always breaking function _definitions_ to align the function name to column 1.

doc/hacking/style_guide.md
CFSworks: No strong opinions either way, and I even catch myself debating which one's better in the middle of breaking apart my lines. The "operator at the end of the line" choice was because this is what occurs far more often in the Panda codebase.

doc/hacking/style_guide.md
CFSworks: It's also my personal choice of pointer style (I never knew that tidbit about the C++ parser) and what's overwhelmingly (universally?) used in the existing Panda codebase.

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+Try to group logically-similar lines, separating them with a single blank line.
+
+Modern language features
+------------------------
+
+Panda3D is a C++11 project.  The use of the following modern language features
+is greatly encouraged:
+
+1. `nullptr` over `NULL`
+2. `auto` (when used to store a function return value only)
```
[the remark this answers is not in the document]
CFSworks: Which is preferred: `auto` or adding the typedef? If the latter is preferred, we shouldn't allow `auto` at all since it wouldn't encourage the use of the typedef.

If `auto` is supported by all of our target C++ compilers, might we consider _removing_ the typedefs and using `auto` instead? Or is the reminder in the C++ code about what is being iterated over especially helpful?

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+Comments should be properly-formatted English, with proper capitalization and
+punctuation rules.  There are two spaces between sentences.
+
+Note that `//` is preferred over `/*` except in some special circumstances.
+
+Line limits and continuations
+-----------------------------
+
+Lines should be limited to 80 columns.  This is not a hard limit, but if a line
+does go over this length, consider a continuation.
```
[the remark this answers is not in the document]
CFSworks: We can, although if we're making a decision over something that small (78 vs. 80) we should be discussing the strictness of the line limit first.

clang-format allows setting soft constraints on the line limit, where you can tell it that it's OK to go a little over if the alternative would be adding line breaks in awkward spots.

We should also be figuring out whether it's necessary to rewrap something that's under the limit - documentation comments wrapped at 78 do not violate the 80-column limit unless we require using as few lines as possible.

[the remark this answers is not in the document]
rdb: I don't think the line limit should be strict, especially as setting arbitrary limits can actually hurt readability.  An exception is flowing text such as documentation strings, which is easy to wrap and this is currently done at 78 columns.

doc/hacking/style_guide.md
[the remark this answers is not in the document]
rdb: I think typedef is preferred for member variables.  The typedefs help with clarity in other places as well.  However, they aren't for temporary containers used only inside the function, in which case use of `auto` seems fine to me.

I don't feel strongly about enforcing this strictly, though, and this whole thing becomes moot if we allow range-based for.

doc/hacking/style_guide.md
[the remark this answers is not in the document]
CFSworks: I see. So you're suggesting a soft limit for code and a hard limit for comments.

However I'm still not sure how raising the limit from 78 to 80 requires a rewrap. If a comment fits in 78 columns (as they all do currently), it necessarily also fits in 80 (as the current draft proposes). The difference only seems relevant if we want to set the additional rule of using all available space in the line before wrapping.

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+Naming conventions
+------------------
+
+- **Classes**: `CamelCaseFirstUpper`
+- **Filenames**: `lowerCamelCase.h`, `lowerCamelCase.cxx`
+  - Note, every header file must have a unique filename.  They are all
+    installed into the same path.
+- **Functions**: `lower_snake_case_without_prefix`
+- **Local variables**: `lower_snake_case_without_prefix`
+- **Member variables**: `_lower_snake_case_with_initial_score`
```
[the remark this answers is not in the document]
tobspr: I also would like to see enum classes. They are type safe, and less error-prone to use. I think we should refer to enum classes as the desired style (which does not mean we have to convert all old code - just to make sure new code uses this convention)

doc/hacking/style_guide.md
tobspr: From a quick look, it seems that the current code mostly uses `double x=3.0`, maybe we should then stick with that?

doc/hacking/style_guide.md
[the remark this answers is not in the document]
tobspr: More about enums: 

 1. Spacing for enum values? (Currently we seem to use 1))
```cpp
enum class MyEnum {
  my_value = 1, // 1)
  other_value=2 // or 2) ?
}
```

2. Naming for enum values. Assuming we use enum classes, how would the members be named? `unsigned_byte`, `UnsignedByte` or `unsignedByte` ? I think the current style would be `unsigned_byte` then.

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+
+Indentation
+-----------
+
+Always use two spaces to indent.  Tabs should never appear in a C++ file.
+
+Access modifiers (`public`/`private`/`protected`/`PUBLISHED`) are never
+indented, and do not affect the indentation of the surrounding lines.
```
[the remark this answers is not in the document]
tobspr: @rdb regarding goto labels, we should also probably mention somewhere that they should not be used (Although this should be clear)

doc/hacking/style_guide.md
tobspr: I think I would agree to rdb, while aligment can be useful in some cases, in many cases it is just annoying to maintain.

doc/hacking/style_guide.md
tobspr: I personally think putting else on a seperate line looks weird:
```
if (condition) {
  statement;
}
else {
  statement;
}
```

or even

```
if (condition) {
  statement;
}
else
{
  statement;
}
```


My personal preference would be similar to rdbs, so something like:

```
if (condition) {
  statement;
} else {
  statement;
}
```

And if the statements are longer, inserting a space at the end. However, this might be hard to express as a rule ..

doc/hacking/style_guide.md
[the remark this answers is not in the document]
tobspr: I think the most reasonable would be to set the limit to 80 characters, and only ensure that a line is never longer than that. In comments you will rarely hit exactly 80 characters. For example: 

```
/**
 * Replaces the current system-RAM image with the new data, converting it       // 73
 * first if necessary from the indicated component-order format.  See           // 79
 * get_ram_image_as() for specifications about the format.  This method cannot  // 78
 * support compressed image data or sub-pages; use set_ram_image() for that.    // 76
 */
```

If we would add a rule that a comment line always has to be 80 characters, this would
require having to fill the remaining space with spaces or so .. that would be really weird (And would introduce trailing spaces).

This way, its fine to have docstrings with any wrap (Currently 78) and never more than 80 characters long per line.

doc/hacking/style_guide.md
tobspr: @rdb I actually would prefer that if your variables have a different type, you declare them on seperate lines. 
```
Object *x;
Object y;
```

This makes it perfectly clear what type both variables have. I personally don't like declaring two lines in a single line at all, but I guess since we use it at many places, we should allow it.

tobspr: I would also like to discuss alignment of `*` and `&` operators. Should it be:
`int& x` or `int &x` ?  `int* x` or `int *x`?`

I think we had this discussion a while ago already. It depends on wheter you call it `x is a pointer to int` or `the type of x is an integer pointer`. I personally prefer the latter, but I think we use the former in the codebase.

doc/hacking/style_guide.md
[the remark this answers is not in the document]
tobspr: I think we should make clear where to use auto and where not. My personal opinion:
- Discourage use to declare new variables. For example: `auto x = some_func()`
- Discourage for return types: `auto myfunc() => int {` or with C++14 `auto myfunc() {`
- Encourage to use for range based for loops (once supported): `for (const auto& x : container)`
- Encourage to use for lambdas (once supported): `auto myfunc = []() { return something(); }`
- Encourage to use for iterator types, e.g. `auto begin = mycontainer.begin()` (Although, this should be rarely necessary with range based for loops`

`auto` can make the code much cleaner, but should be used with care

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+
+Modern language features
+------------------------
+
+Panda3D is a C++11 project.  The use of the following modern language features
+is greatly encouraged:
+
+1. `nullptr` over `NULL`
+2. `auto` (when used to store a function return value only)
+3. Range-based for loops
```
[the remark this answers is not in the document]
tobspr: I think we should *really* start to support modern language features. When coming from a modern development environment, its always a change to switch to pandas c++ standard. Having to declare iterator types for example is something you don't really have do in modern c++ development.

doc/hacking/style_guide.md
```diff
@@ -0,0 +1,187 @@
...
+
+The following modern language features are discouraged:
+
+1. C++17 features
+2. `std::atomic`
+3. Lambda functions
```
tobspr: I agree on 1. and 2., but I think we should support lambdas. They can be really handy, and can also be more performant (than passing a simple function pointer).

tobspr:
I have added my comments. Some topics which I think are missing:

- What about standard headers? For example `#include <vector>`. I think we should explicitely state which headers can be used, and which not. (I vaguely remember that we don't support `unsorted_map` for example)

Maybe we can also have a look at other big c++ projects (Boost for example) and see which rules they encourage in their style guide

rdb:
Maybe we could also agree on preprocessor indentation style?
```
#if ...
#  error ...
#endif
```
vs.
```
#if ...
  #error ...
#endif
```

doc/hacking/style_guide.md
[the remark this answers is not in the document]
rdb: I agree on using enum classes in the future.

I think enum values should be spaced because they look like assignments, but I can go either way on that.

As for enumerant naming style, I would suggest `unsigned_byte` or `UNSIGNED_BYTE`.  Normally I would strongly prefer the former, but it seems that [Python settled on](https://mail.python.org/pipermail/python-ideas/2016-September/042340.html) using `UNSIGNED_BYTE` for enumerants (and all other constants), so I see three options:
1. Defy the Python convention and use the lowercase style.
2. Use the UPPER_CASE style both in Python and C++ code.
3. Use lower_case in C++, but let interrogate map this to UPPER_CASE.

doc/hacking/style_guide.md
rdb: I think all of David's code uses spaces around `=`, except for the `=0` behind virtual functions.

doc/hacking/style_guide.md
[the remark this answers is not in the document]
rdb: @cfsworks: changing the limit from 78 to 80 means that running a tool that is programmed to rewrap to 80 would make lots of unnecessary changes to documentation strings.  I just see not much reason to change the rule again since we have previously explicitly run a tool to rewrap everything to 78.

…but maybe this is just my OCD-ishness talking, so if you both feel we should make it 80 characters, I'm willing to be overruled on this.

doc/hacking/style_guide.md
rdb: C++ types are best read right-to-left, for example:
```cpp
// foo is a reference to a const pointer to an integer
int *const &foo;
```

I would personally prefer aligning the `&` and `*` to the right, because it matches how C++ parses declarations, and because it matches what most of the codebase is currently doing.  That said, I think this should be a guideline, not a rule; I would not reject a PR simply because it happens to align the `*` and `&` on the other side.

I agree on declaring the variables on different lines if it could easily be confusing.  An exception is a simple case where there can be no confusion:
```cpp
int a, b, c;
```

doc/hacking/style_guide.md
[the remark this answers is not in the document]
rdb: I agree with most of that, except in the case of range-based for loops, the member type is generally simple enough to write out and can make the code easier to understand by being explicit.  `auto` could make unclear whether a reference or a local copy of the type is created.

I would also advise that anything that interrogate may need to wrap should not use `auto`, which may be relevant in the future if we want to wrap lambdas for Python code.  (Interrogate will parse `auto`, but it won't deduce the return type automatically except in very simple cases.)

doc/hacking/style_guide.md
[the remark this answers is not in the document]
rdb: I have opened #288 to gauge opinions about this.

doc/hacking/style_guide.md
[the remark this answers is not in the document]
tobspr: @rdb I think python is a special case, since you usually declare and access them like class members, so its nice to distinguish. Since in C++, one would have to write `Texture::Format::unsigned_byte` I think it would be better to stick to underscores, since `Texture::Format::UNSIGNED_BYTE` looks weird to me (Just my opinion)

For the python interface we could probably make them upper case then.

doc/hacking/style_guide.md
tobspr: We could maybe set a preference for placement, and disallow declaring variables on the same line if they have a different type ?

doc/hacking/style_guide.md
tobspr: I don't have a strong opinion on that, we can settle for whatever you prefer

tobspr:
I would *really* prefer the latter preprocessor style:

```
#if ...
  #error ...
#endif
```

I know this is against the current style, but in my opinion that looks much cleaner, especially when nesting ifs. With the hash on the beginning, it is not directly clear if this is a preprocessor if or a language if.

CFSworks:
I'm in favor of the latter preprocessor directive style as well - though we should also decide on the indentation before the # (is it always in column 1 or does it follow the surrounding scope?)

rdb:
OK, I'm willing to go with that.

Indentation before the `#` should not intermix with indentation of the surrounding C++ code, though, since that would be misleading (since it doesn't obey the same scoping rules).  So, I would prefer:
```cpp
void function() {
  if (stuff) {
#ifdef _WIN32
    CallSomeWin32Func();
#endif
  }
}
```

tobspr:
I disagree with that - while it does not have a scope, it makes it much more readable in my opinion:
```cpp
void function() {
  if (stuff) {
    #ifdef _WIN32
      CallSomeWin32Func();
    #else
      #ifdef __CLANG__
        CallSomeOtherFunc()
      #endif
    #endif
  }
}
```

Opposed to:
```cpp
void function() {
  if (stuff) {
#ifdef _WIN32
      CallSomeWin32Func();
#else
  #ifdef __CLANG__
        CallSomeOtherFunc()
  #endif
#endif
  }
}
```


Especially when nesting ifs and elses, I think this makes it more readable

rdb:
@tobspr @CFSworks @Moguri Could we settle on a convention for enum classes?  Two options I would consider:

```cpp
Texture::Format::rgba8
CoordinateSystem::zup_right
```
versus:
```cpp
Texture::Format::RGBA8
CoordinateSystem::ZUP_RIGHT
```

[Python seems to](https://mail.python.org/pipermail/python-ideas/2016-September/042340.html) prefer UPPER_CASE for [Python 3-style enums](https://docs.python.org/3/library/enum.html), matching the PEP 8 convention for constants, so that's a strong argument in favour of that. There's also plenty of precedent for constants being UPPER_CASE in other programming languages (inheriting the convention from C, where constants were written as macros). On the other hand, there are also [good arguments](https://accu.org/index.php/journals/1923) to be raised against this convention.

I think I am personally very slightly leaning towards the UPPER_CASE convention in order to just keep to the accepted Python style and keep our C++ code aligned with the same conventions, but I am happy to be overruled on this.

tobspr:
@rdb I don't have a strong opinion on this. For me, both ways would work out. I feel like the underscore one would fit better into the C++ style, but since the uppercase one seems to be the Python convention, I guess it would be a good way to stick with that

Moguri:
I do not have a strong opinion either way. I would lean toward whatever we are already doing to avoid the need to touch a ton of files.

rdb:
@Moguri we are not technically doing anything at the moment since we don't use the new C++11-style enums in the codebase yet.  Currently, the enums are defined like so:
```
enum SomeEnum {
  SE_value_one,
  SE_value_two,
};
```
The abbreviated prefix is necessary to prevent name collisions, as old-style enumerants are actually added to the parent scope.  C++11-style enums have their own scope and therefore don't need such a prefix.  (Interrogate maps C++11-style enums directly to Python's native enums.)

I suppose the fact that we are currently doing `SE_value_one` and not `SE_VALUE_ONE` is somewhat of an argument in favour of making it `SomeEnum.value_one` in the future, and not `SomeEnum.VALUE_ONE` as Guido would have us do.

I am willing to go either way on this too.  Let's see what @cfsworks thinks (or anyone else who wants to chime in).

CFSworks:
@rdb 

I have a preference for lowercase, qualified with the name of the enum class. And a lesser preference for uppercase if not.

So more preferable: `CoordinateSystem::zup_right`
Less preferable: `ZUP_RIGHT`
Never: `CoordinateSystem::ZUP_RIGHT` or `zup_right`

The reason is that both of these formats call attention to the fact that this isn't a local variable. C users are quite used to uppercase being a constant, and C++ users treat the `::` symbol similarly (it's often used to access a static method or static variable). We shouldn't do both - the article you linked had some good points about reserving uppercase for the preprocessor.

As for Python bindings, there's two options: Have Interrogate upcase everything, or leave it as-is and deviate from the Python convention. I'm leaning a little toward the former as Python has exploded in popularity since the initial public release of Panda3D, so consequently I think we should start taking Python style guides very seriously these days. If we think the consistency between C++/Python is more important, I can side with that (since uppercase enums is only a _recommendation_ rather than prescribed in PEP8 or any style guide I can readily find). What I very much want to avoid is having both (like what we do with setMethodName and set_method_name) if we can avoid it - this pollutes the class dicts and makes the binaries a little larger to boot.

So in summary:
- `CoordinateSystem::zup_right` in C++
- `CoordinateSystem.ZUP_RIGHT` (or `CoordinateSystem.zup_right` for consistency) in Python

rdb:
Oops, forgot to follow up on this.  Okay, if we're going to use lower-case for C++ enum classes, I would also prefer deviating from the Python style guide here unless there are serious objections to this.  I would value consistency between C++ and Python as being more important to avoid programmer confusion.  I also personally find it more consistent and readable to use lower-case for all variables whether they are constants or not.  We have also never used upper-case constants in the Python bindings before.

doc/hacking/style_guide.md
montreal91: Maybe we should consider [these recommendations](https://google.github.io/styleguide/cppguide.html#Functions) in general and [these](https://google.github.io/styleguide/cppguide.html#Default_Arguments) on default arguments precisely.

Moguri:
@rdb  and @CFSworks where are we at with this PR? What are the next steps? Is there a subset of this PR that we can get merged while we work out details on some of the finer (and possibly more controversial) points?

rdb:
I went ahead and updated and committed the WIP style guide so I can close this PR. Further updates welcome.

PR #285, opened 2018-03-19
controlManager: allow currentControls to setSpeeds

xseano:
This allows the local POTCO/ToonTown gravity walkers to receive updates to speed fluctuations.

### DIFF
--- a/direct/src/controls/ControlManager.py
+++ b/direct/src/controls/ControlManager.py
@@ -139,6 +139,8 @@ class ControlManager:
         for controls in self.controls.values():
             controls.setWalkSpeed(
                 forwardSpeed, jumpForce, reverseSpeed, rotateSpeed)
+            self.currentControls.setSpeeds(
+                forwardSpeed, jumpForce, reverseSpeed, rotateSpeed)
 
     def delete(self):
         assert self.notify.debugCall(id(self))
@@ -342,4 +344,3 @@ class ControlManager:
 
             inputState.set("turnLeft", False, inputSource=inputState.WASD)
             inputState.set("turnRight", False, inputSource=inputState.WASD)
-
--- a/direct/src/controls/ControlManager.py
+++ b/direct/src/controls/ControlManager.py
@@ -133,14 +133,14 @@ class ControlManager:
         else:
             assert self.notify.debug("Unkown controls: %s" % name)
 
-    def setSpeeds(self, forwardSpeed, jumpForce,
-            reverseSpeed, rotateSpeed, strafeLeft=0, strafeRight=0):
+    def setSpeeds(self, forwardSpeed, jumpForce, reverseSpeed, rotateSpeed, strafeLeft=0, strafeRight=0):
         assert self.notify.debugCall(id(self))
+
+        if self.currentControls is not None:
+            self.currentControls.setSpeeds(forwardSpeed, jumpForce, reverseSpeed, rotateSpeed)
+
         for controls in self.controls.values():
-            controls.setWalkSpeed(
-                forwardSpeed, jumpForce, reverseSpeed, rotateSpeed)
-            self.currentControls.setSpeeds(
-                forwardSpeed, jumpForce, reverseSpeed, rotateSpeed)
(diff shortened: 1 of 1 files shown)


xseano:
For instance in `PiratesGravityWalker` it would update `setSpeeds` 

```python
    def setSpeeds(self, forwardSpeed, jumpForce, reverseSpeed, rotateSpeed):
        self.avatarControlForwardSpeed = forwardSpeed
        self.avatarControlJumpForce = jumpForce
        self.avatarControlReverseSpeed = reverseSpeed
        self.avatarControlRotateSpeed = rotateSpeed
        self.notify.warning('setSpeeds: (%s, %s, %s, %s)' % (self.avatarControlForwardSpeed, self.avatarControlJumpForce, self.avatarControlReverseSpeed, self.avatarControlRotateSpeed))
```

Which sets up the av globals used in calculating control factors:
```python
...
self.speed = -(self.avatarControlReverseSpeed)
        if not reverse and slideLeft or -(self.avatarControlReverseSpeed) * 0.75:
            if not reverse and slideRight or self.avatarControlReverseSpeed * 0.75:
                if (slideLeft or -(self.avatarControlForwardSpeed) * 0.75) and slideRight:
                    pass
        self.slideSpeed = self.avatarControlForwardSpeed * 0.75
        if (turnLeft or self.avatarControlRotateSpeed) and turnRight:
            pass
        self.rotationSpeed = -(self.avatarControlRotateSpeed)
        if self.speed and self.slideSpeed:
            self.speed *= GravityWalker.DiagonalFactor
            self.slideSpeed *= GravityWalker.DiagonalFactor
...
```

thetestgame:
To prevent potential issues with other applications you should verify that currentControls is not NoneType before attempting to use the variable. The movement also should be outside the for loop.

On a side note. Having reviewed both 08 and 2013 POTCO dumps I found no references to a ``setSpeeds`` inside the source outside of the normal ``ControlManager``  ``setSpeeds`` function. Are you sure this is not local to your use case?

xseano:
Any issues with this @rdb ?

thetestgame:
I'd like to heard rdbs opinion but personally it feels use case specific. Having restored POTCO and ToonTown I have never run across a time where this change was required.

rdb:
I don't know.  I'm hoping someone more familiar with this part of DIRECT will be able to review this code.

CFSworks:
`self.currentControls` contains the currently-selected avatar control system. Games may implement their own controls if they like, but the ones included in direct.controls are:
- `BattleWalker`
- `DevWalker`
- `GhostWalker`
- `GravityWalker`
- `NonPhysicsWalker`
- `ObserverWalker`
- `PhysicsWalker`
- `SwimWalker`
- `TwoDWalker`

The interface exposed by the various control systems to the `ControlManager` appears to consist of:
- `delete`
- `deleteCollisions`
- `disableAvatarControls`
- `enableAvatarControls`
- `getIsAirborne`
- `getSpeeds`
- `placeOnFloor`
- `setAvatar`
- `setCollisionsActive`
- `setTag`
- `setWalkSpeed`

No controls implement `setSpeeds` - that's a function of `ControlManager` itself. (Indeed, a `grep -R 'def setSpeeds' .` of the entire Panda3D tree only turns up `ControlManager`.) While certain out-of-tree controls may implement a `setSpeeds` method, it's certainly not part of the API, and so making `ControlManager` rely on it for proper operation will only break existing Panda3D apps that use the in-tree control systems.

I don't fully understand the objective of this change, or I'd try to provide more useful feedback. I assume "receive updates to speed fluctuations" to mean "support analog/fractional controls rather than digital/boolean input states" but the various `setWalkSpeed` methods already respect non-binary inputs. So, I think I can't really say anything more helpful than "This gets a :-1: from me."

AnythingTechPro:
👎

Bob693:
This hurts my soul :(

rdb:
OK, yeah, adding a call to a method that does not actually exist in Panda also gets a 👎 from me.

It should be trivial to subclass ControlManager and override setSpeeds to do whatever you want.

PR #287, opened 2018-03-20
Make export/import switching more granular

CFSworks:
The basic idea behind this PR is to break apart e.g. `BUILDING_PANDA` into a bunch of granular switches to mirror `EXPCL_PANDA_COMPONENT`. This gives the buildsystem more control over how Panda is built; the Panda codebase does not assume a metalibs-like structure any longer, and the grouping/organization of component libs into larger metalibs is up to the build configuration. This eventually gives us the freedom to experiment with various metalibs configurations and possibly even deprecate metalibs.

Note that makepanda isn't affected by this; this is mostly a change to facilitate CMake, which won't use the `BUILDING_PANDA` flag at all and will instead dynamically control the `BUILDING_PANDA_COMPONENT` flags based on what components are being built into the same metalib.

We may even consider killing off pandasymbols altogether, instead putting the "#ifdef BUILDING_PANDA_WHATEVER" into config_whatever.h.

**When reviewing**: Please review the individual commits in order. Viewing this PR as a lump sum will just make following the thought process behind each change much harder. The last commit in this series (55dad22) is also optional; I'm using it to sanity-check my CMake improvements, however I figured mainlining it will help more than hurt.


CFSworks:
I'll go ahead and pull this in then, since we agree with the substance of the change.

The `EXPCL_LIBPANDA` looks out-of-place to me too. However, it's named that for consistency with `BUILDING_LIBPANDA` - I can't use `BUILDING_PANDA` for that because it's the shortcut used by makepanda to switch on *all* exports for libraries conventionally built inside the libpanda metalib, which means I can't use `EXPCL_PANDA` or it'd be inconsistent with `BUILDING_PANDA`.

I'm not sure I *quite* understand your question; I think what you're asking is "if the libpanda metalib is always built with BUILDING_PANDA, can't we just use EXPCL_PANDA in the metalib?" (correct me if I'm wrong). I've thought about this myself, but there could be configurations where the metalib is built as a stub just to provide a stable API (e.g. the plugin metalibs) but pulls in all of its code from dynamic libraries.

I plan on adding other `BUILDING_LIB*` switches, one for each plugin as well (gl, etc.) so I think it'll get a little less inconsistent as time goes by.

PR #291, opened 2018-03-24
Create a simple EventMapper

Moguri:
This new EventMapper generates a map from ConfigVariables that takes an
input event to generate output events. This makes it easier to create
abstract events (e.g., "jump") for inputs (e.g., "space"). The
EventMapper will take any ConfigVariable of the form
"event-map-item-<out_event> <in_event1> ... <in_eventN>" and create a
map where each <in_eventN> will trigger <out_event>.

There are still some TODO items to resolve, but I think there is enough to start some code review to make sure I am going down the right path.

TODO items:

- [x] ~~Allow using keyboard map to have config files be keyboard layout agnostic~~ (this can be done with raw events)
- [x] Add function that returns a list of input events for a given output event (useful for generating instructions such as "press <input> to jump")
- [x] Add license header and Doxygen comments

### DIFF
--- a/direct/src/showbase/ShowBase.py
+++ b/direct/src/showbase/ShowBase.py
@@ -306,6 +306,8 @@ class ShowBase(DirectObject.DirectObject):
 
         self.AppHasAudioFocus = 1
 
+        self.eventMapper = EventMapper.get_global_ptr()
+
         # Get a pointer to Panda's global ClockObject, used for
         # synchronizing events between Python and C.
         globalClock = ClockObject.getGlobalClock()
--- a/panda/src/device/p3device_composite1.cxx
+++ b/panda/src/device/p3device_composite1.cxx
@@ -1,4 +1,3 @@
-
 #include "clientAnalogDevice.cxx"
 #include "clientBase.cxx"
 #include "clientButtonDevice.cxx"
--- a/panda/src/event/config_event.cxx
+++ b/panda/src/event/config_event.cxx
@@ -21,6 +21,7 @@
 #include "buttonEventList.h"
 #include "event.h"
 #include "eventHandler.h"
+#include "eventMapper.h"
 #include "eventParameter.h"
 #include "genericAsyncTask.h"
 #include "pointerEventList.h"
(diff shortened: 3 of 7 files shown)


Moguri:
My todo items are complete, but I would like to figure out how to expose EventMapper::get_inputs_for_event() to Python.

rdb:
Thanks a lot for your work on this.

While I think there is a useful place for an event alias system, I think that we first need to have a discussion about the design of such a feature, before we get into the details of the code.  Some issues I'd like discussed are:

1. This implementation makes assumptions that an event has a corresponding `-up` and `-down` suffixed event.  This is not an inherent property of the event system, however; just something that ButtonThrower happens to be doing, alongside other related events like `*-repeat` and `time-*` which aren't handled here.  If we want to do something like that, I think it would make more sense if this were something in the data graph that renamed key inputs *before* they become events, or something that was interfaced with or integrated into ButtonThrower, in order to maintain the modularity of the subsystems.

FWIW, `-down` events don't actually exist, to my knowledge.  Down events simply have no suffix.  Perhaps you were thinking of `-repeat` events.

2. It seems to only handle defining aliases via config, without a corresponding API in Panda.  It seems that it should be straightforward to separate out the code that defines the alias into a method.

3. This implementation is implemented as a global singleton, even though it really operates on the scope of an event handler.  If we're going to define this globally as opposed to integrate it into data graph or ButtonThrower, it's not really clear to me why this needs to be a class in itself, rather than something integrated into EventHandler, with an `EventHandler::add_alias` method.  This leaves open the possibility of a potential ShowBase replacement that allows multiple App instances with separate event queues.

This is just some initial thoughts to get a discussion going.  I'm also happy to see what others think.

BMaxV:
Hey there

I kind of wrote something like this in python for the bge to do remappable keys / multi language support / networking more easily. If I understand this PR correctly it would do something similar?

My method was basically defining a key string in my game module mapping to a function and that was passed from the interface either directly or via networking to the game and triggered the event.

I'm just starting to look at panda, but from a just a quick glance at the event system a "one key broadcast" is already possible?

Moguri:
I am abandoning this PR for the time being. I am working on developing and proving out an API for this in Python as the [panda3d-eventmapper](https://github.com/Moguri/panda3d-eventmapper) project. I can convert it to C++ and upstream it if/when I am happy with.

PR #295, opened 2018-03-31
Fix DC file reading regression caused by ac8417f

darktohka:
In ac8417ffdf0f253d8c2653b9a27becee1833778c type checks were changed to use Python's inspect module in order to safeguard against any Python 2 and Python 3 differences. However, an inequality check was replaced with an incorrect equality check.

This means that bonafide classes are regarded as non-classes and vice-versa, and as such DC files can no longer load, raising a "Symbol ClassName is not a class name." exception on load.

This commit resolves that issue.


CFSworks:
Oh! Beg pardon, yes, that's absolutely correct. Sorry about that and thanks for catching it!

PR #298, opened 2018-04-06
Separating strings from blobs in dcparser

rdb:
This is a suggestion for how to solve the Python 3 compatibility issues in dcparser.  It already makes a distinction between text strings and binary blobs, so this change changes it so that `vector_uchar` is used for decoding and encoding blob data.

It also uses `vector_uchar` for representing any packed data.  A separate `get_bytes()` is added to DCPacker since the existing `get_string()` isn't really named accurately any more.

@cfsworks I've marked you as reviewer because you're in a better position than I am at checking whether this is going to break anything, or whether this it is at all a good idea to treat strings as UTF-8 in dcparser.

### DIFF
--- a/direct/src/dcparser/dcArrayParameter.cxx
+++ b/direct/src/dcparser/dcArrayParameter.cxx
@@ -263,6 +263,38 @@ pack_string(DCPackData &pack_data, const string &value,
 }
 
 /**
+ * Packs the indicated numeric or string value into the stream.
+ */
+void DCArrayParameter::
+pack_blob(DCPackData &pack_data, const vector_uchar &value,
+          bool &pack_error, bool &range_error) const {
+  // We can only pack a string if the array element type is char or int8.
+  DCSimpleParameter *simple_type = _element_type->as_simple_parameter();
+  if (simple_type == (DCSimpleParameter *)NULL) {
+    pack_error = true;
+    return;
+  }
+
+  size_t blob_size = value.size();
+
+  switch (simple_type->get_type()) {
+  case ST_char:
+  case ST_uint8:
+  case ST_int8:
+    _array_size_range.validate(blob_size, range_error);
+    if (_num_length_bytes != 0) {
+      nassertv(_num_length_bytes == 2);
+      do_pack_uint16(pack_data.get_write_pointer(2), blob_size);
+    }
+    pack_data.append_data((const char *)value.data(), blob_size);
+    break;
+
+  default:
+    pack_error = true;
+  }
+}
+
+/**
  * Packs the arrayParameter's specified default value (or a sensible default
  * if no value is specified) into the stream.  Returns true if the default
  * value is packed, false if the arrayParameter doesn't know how to pack its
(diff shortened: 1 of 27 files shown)


CFSworks:
@rdb Thanks! I'll review it when I'm not on mobile.

I've been thinking about this a lot myself and like the idea of making string officially a UTF8 type. This would mean some changes to Astron which currently assumes (I believe) Latin1, but UTF8 is much better.

One thought I'm grappling with, however, is how to treat range constraints on the string type. UTF8 breaks the assumption that the string length equals the data length (except for ASCII strings). On one hand, at the high level, users except those length constraints to apply to the actual value being sent. But on the other, low level users will need a mechanism to provide explicit size limits.

The former is more consistent with the other length constraints, but the latter is faster to check. Although, network intermediaries (e.g. Astron's CA) should verify well-formed UTF8 from any malicious source anyway, so maybe we should go that route and just tell people "maximum size is 4 times maximum length"?

rdb:
Theoretically we could distinguish latin-1 strings from UTF-8 strings, but I would prefer to just keep things simple and say that all strings everywhere in Panda are UTF-8.  I also think we should get rid of the use of wide strings in Panda3D except for communicating with the Win32 API.

So far, everywhere we're using the notion of 'size' and 'length' of strings (which are annoyingly the exact same thing in the C++ string class), it is referring to the number of code units, not code points, since most code that processes strings is completely oblivious of the encoding and just cares about the number of bytes in memory.  Similarly, it would be easiest for networking code not to have to parse the strings in order to process them.

What exactly is the aim of the range constraints you refer to?

CFSworks:
> What exactly is the aim of the range constraints you refer to?

The .dc syntax allows writing constructs like `setName(string(1-64) name)`.
When `string` was assumed to be ASCII, this would set both the length (in characters) and size (in bytes) limits for "name" - the DC system would then perform constraints checking on any received or to-be-sent message to ensure that constraint is respected. It's useful both as a debugging aid as well as a countermeasure against maliciously crafted input. For example, nobody can set a blank name (due to requiring a minimum length of 1) and nobody can overflow the handling code (due to having a maximum size of 64).

Decoupling length and size by switching to UTF8 means we need to decide which one the range syntax applies to - they both have their uses, but I'm thinking "length" (as in number of codepoints) is the more consistent with what a game programmer would expect (they tend not to care about the bytes on the wire), and with the other range constraints (e.g. in arrays it applies to the number of array elements).

I'm not in favor of Latin1, by the way. Just commenting that I'll need to make the same change to the Astron codebase once this change lands here.

rdb:
Ah, thank you for the explanation.  Then yes, number of code points makes the most sense here, since that's independent of the exact encoding.  People will expect the length restrictions to be consistent with the Python-reported `len()`, as well.

The only case where I thought people might expect it to be the number of code units (ie. bytes) is if this constraint is imposed based on a hard database or storage limitation, eg. if they need it to fit in a 64 byte field, but it seems that databases like MySQL actually measure string size constraints in number of codepoints too, even for UTF-8.

An optimized, quick range checker could entirely skip the range validation if `num_code_units <= limit`, and can early-reject the string if `num_code_units > limit * 4`, but I suppose it may be desirable to add additional validation that the string is indeed valid UTF-8 to catch encoding errors early.

abtechmg:
How would i go about adding this to the latest development build? I finally after hours of installing and uninstalling found out how to compile source without issue, i however attempted adding this but obviously did something wrong. I'm not pro at C++ so i have no clue what i'm doing to be honest.

drewc5131:
### praise the lord finally

rdb:
@abtechmg I just rebased the branch onto the latest master, so you should be able to just check out or merge the `dcparser-bytes` branch without issues now.

I'll leave it up to @CFSworks to merge this at his discretion since he may need to coordinate it with changes to Astron.

rdb:
@cfsworks Do you have time to look into this one within the next 10 days?  I'd like this to be in 1.10.

rdb:
@CFSworks It'd suffice to do a quick smoke test to ensure this at least doesn't break existing Python 2 applications.

CFSworks:
@rdb I just did a fairly heavy tire-kicking with this. It doesn't break Python 2. Is that enough to cherry-pick for 1.10?

rdb:
@CFSworks Perfect, thanks a ton!

PR #308, opened 2018-04-30
python3 prelinked modules names are wrong

pmp-p:
importlib is not there to correct ```__name__``` when doing package import

### DIFF
--- a/dtool/src/interrogate/interfaceMakerPythonSimple.cxx
+++ b/dtool/src/interrogate/interfaceMakerPythonSimple.cxx
@@ -95,7 +95,7 @@ write_module(ostream &out,ostream *out_h, InterrogateModuleDef *def) {
       << "#if PY_MAJOR_VERSION >= 3\n"
       << "static struct PyModuleDef python_simple_module = {\n"
       << "  PyModuleDef_HEAD_INIT,\n"
-      << "  \"" << def->library_name << "\",\n"
+      << "  \"" << def->module_name << "\",\n"
       << "  NULL,\n"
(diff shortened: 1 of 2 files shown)


BMaxV:
I would suggest to close this one as well? Seems a bit like info is lacking, build is failing...

pmp-p:
@BMaxV this is specific to webgl-port and for now dynamic linking in emsdk is still broken :(

rdb:
Maybe you could give a bit more info about the problem this is intended to solve?  I checked and Python expects us not to pass the fully-qualified module name here, because it assigns the name via _Py_PackageContext:
https://github.com/python/cpython/blob/d7c387384a27f37e4e3fa7890c859212c56b45b2/Objects/moduleobject.c#L199-L214
I could be persuaded (since it may not have ill effect), but I'd need a good reason.

rdb:
Thinking a bit about this, I can imagine that this would be an issue when static linking and manually initializing the libraries, since _Py_PackageContext does not get set in that case.  Is that correct?

I don't mind accepting this PR if it helps with that—but it does mean that you could no longer put core.so/pyd into a different package.  Maybe that is not such a bad thing?  Alternatively, we could use the fully-qualified name only if LINK_ALL_STATIC is set.

pmp-p:
> manually initializing the libraries

exactly, unless dlopen() and dlsym() work again in emsdk ( they could now : haven't tested for about a month since pyodide people are on it too), the module must be initialized manually ( even if loaded as javascript lib , like i was doing when i hit that problem ).

PR #314, opened 2018-05-06
deploy-ng

Moguri:
The deploy-ng branch is ready to begin a review process for merging into master. In addition to code review, this PR will be used to track any more pre-merge tasks that need to be done.


Moguri:
I have updated the [manual page](https://www.panda3d.org/manual/index.php/Using_Setuptools) (mostly formatting). There are still a couple of TODO items listed on the page that I need to address.

@rdb, anything you would like to see added/changed/removed from that page? Any other documentation you would like to see for deploy-ng?

Moguri:
The TODO comments have been cleaned up from the Manual page and some fixes have been added to get deploying to Linux from Windows working.

makepanda/makepanda.py
```diff
@@ -438,7 +438,7 @@ def parseopts(args):
 
     PLATFORM = 'macosx-{0}.{1}-{2}'.format(osxver[0], osxver[1], arch_tag)
 
-elif target == 'linux' and os.path.isfile("/lib/libc-2.5.so") and os.path.isdir("/opt/python"):
+elif target == 'linux' and (os.path.isfile("/lib/libc-2.5.so") or os.path.isfile("/lib64/libc-2.5.so")) and os.path.isdir("/opt/python"):
```
montreal91: This condition is very long and hard to understand and comprehend. Maybe it's better to put it into a separate method?

rdb:
I think we should decide where this belongs in the source tree.  Rather than reside in showutil, I'd rather take the distribution stuff out of there and put it inside some `direct.dist` package or something.

Eventually, it should probably be taken outside of `direct` altogether, but we don't currently have a tree where that would fit.

Moguri:
I am fine with `direct.dist`. We can rename `dist.py` to `commands.py`. Can `FreezeTool.py`, `pefile.py`, and `pfreeze.py` safely be moved as well, or would this break APIs?

I would be tempted to take the opportunity to move some more things from `dist.py` to `FreezeTool.py` such as:

* PACKAGE_* stuff
* SITE_* stuff
* dependency management

I would also like to create an `InstallerTools.py` or some such to put generic installer building tools that bdist_apps calls. The old p3d system would also be able to make use of it if we were keeping it around. However, I think I would prefer to not risk breaking things this close to a release, and moving this stuff around likely doesn't break any APIs.

As for a location outside of `direct`, we may want to consider a separate repo/project (still under the panda3d org) in the future to make it easier to use outside of Panda. This would also be a "sometime in the future" thing.

rdb:
I think those modules can be moved, as long as we fix imports to them.  Not sure if anything is using FreezeTool outside pfreeze and p3d, though we could leave an import redirection shim in place.

I think renaming `dist.py` to `commands.py` is a good idea.  Some stuff from there and from FreezeTool could be split out, but I don't have a strong opinion on the particulars of that at this point; as long as things are roughly in the right place.  We can always split things out more later.

rdb:
Thanks so much for all your work on this!

PR #320, opened 2018-05-15
delete *AI.py and *UD.py

manoguy535:
Panda does not provide an ai or ud repository so they are meaningless outside of specific code

@rdb mentioned in: https://github.com/panda3d/panda3d/pull/151


CFSworks:
I don't know that I like this. The intent is that there will eventually be first-class support for the Astron project within `direct`, so deleting this now just means it will have to be added back later, which may do nothing but destroy the history for these files. The AI files are also useful within the CMU server, and there may very well be projects out there relying on them.

What is this PR trying to achieve, ultimately? Are there other ways of accomplishing the same end goal?

CFSworks:
The CMU `ServerRepository` itself isn't exactly an object repository in its own right - you have to establish a loopback connection using an ordinary `ClientRepository` and use that as an "AI" Repository, and use that to manage objects that would typically be managed by the server. In this case, the server objects can be subclasses of `DistributedObjectAI`, and the loopback repository is used as `self.air`, where not all functions within `DistributedObjectAI` are usable due to some of them depending on a more feature-rich air implementation than Panda has.

Panda sadly doesn't have any samples that show this in action; I do remember drwr's tagger game demonstrating proper use of direct's CMU-compatible version of `TimeManagerAI`, and `ClientRepository.sendUpdateToChannel` shows one instance where the CMU `ClientRepository` took over the responsibility of being the server's "internal" repository. Sadly `direct.distributed` has turned into a big mess over the years, since it's never been a "complete" system - it definitely needs cleaning up, but I'd think we'd need to put together a broader plan so we can determine what's really vestigial, what is currently used, and what will be revived soon.

thetestgame:
I agree with @CFSworks. The distributed module has turned into a neglected mess. However I do not feel this is the answer. I know of a large number of people me included that still actively use the AI/UD code and have built up a internal repository with their own OTP like solution. This pull request would break not only our projects but it would also impact Astron users as CFWorks mentioned earlier.

I already have to use a custom almost archival panda for a lot of things due to changes in recent years. This change would only add to me having to use that particular panda more.

rdb:
This would not impact Astron users, since what we do here need not affect the Astron version of the `distributed` package.  However, since there is resistance to removing these modules, I will reject this PR for now.  My policy for Panda 1.x is not to remove things that are deemed useful, or to break functioning Panda3D code (at least not without an easy way to fix it).

@thetestgame Could you send me an e-mail or IRC message about which changes in recent years require you to use a custom version of Panda3D?  If there were changes that broke existing applications without an easy fix, I would like to hear about them so that we can consider reverting those changes.

PR #322, opened 2018-05-17
Get the cmake branch building on macOS

treamology:


treamology:
Whoops, this breaks builds on other platforms, I'll get that fixed when I get back near a computer.

CFSworks:
Sorry for the delay; I was out of town when you opened this PR. That being said: AAAAAAA! Thank you thank you thank you!

I'm restarting the two broken Travis jobs (which seemed to fail only due to timeout) - but I think the next thing to do here would be to edit the .travis.yml file so Travis also builds on macOS. Thoughts?

panda/src/cocoadisplay/CMakeLists.txt
```diff
@@ -31,5 +33,3 @@ find_library(APPKIT_LIBRARY AppKit)
 find_library(CARBON_LIBRARY Carbon)
 target_link_libraries(p3cocoadisplay
   ${APPLICATIONSERVICES_LIBRARY} ${APPKIT_LIBRARY} ${CARBON_LIBRARY})
-
-install(TARGETS p3cocoadisplay DESTINATION lib)
```
CFSworks: Shouldn't the p3cocoadisplay dylib be installed so that libpandagl can link against it?

panda/src/osxdisplay/CMakeLists.txt
```diff
@@ -26,5 +26,3 @@ find_library(COCOA_LIBRARY Cocoa)
 target_link_libraries(p3osxdisplay
   ${APPLICATIONSERVICES_LIBRARY} ${CARBON_LIBRARY} ${AGL_LIBRARY}
   ${CORESERVICES_LIBRARY} ${COCOA_LIBRARY})
-
-install(TARGETS p3osxdisplay DESTINATION lib)
```
CFSworks: Ditto above: Shouldn't this be installed as with other display modules?

panda/src/display/CMakeLists.txt
```diff
@@ -90,6 +90,7 @@ composite_sources(p3display P3DISPLAY_SOURCES)
 add_component_library(p3display SYMBOL BUILDING_PANDA_DISPLAY
   ${P3DISPLAY_HEADERS} ${P3DISPLAY_SOURCES})
 target_link_libraries(p3display p3cull p3pgraphnodes)
+target_use_packages(p3display X11)
```
CFSworks: Was there something stopping p3display from building on X11 platforms without this? As far as I know, p3display is independent of display API, and shouldn't depend on X11.

rdb: `display` uses the X11 headers to pull in the `Window` typedef only, because of `NativeWindowHandle::X11Handle`.  Although, it looks like `Window` is just an unsigned long, so I'm not sure why we had a particular need to do that.

panda/src/cocoadisplay/CMakeLists.txt
```diff
@@ -33,3 +33,5 @@ find_library(APPKIT_LIBRARY AppKit)
 find_library(CARBON_LIBRARY Carbon)
 target_link_libraries(p3cocoadisplay
   ${APPLICATIONSERVICES_LIBRARY} ${APPKIT_LIBRARY} ${CARBON_LIBRARY})
+
+install(TARGETS p3cocoadisplay DESTINATION lib)
```
CFSworks: No newline at end of file

panda/src/osxdisplay/CMakeLists.txt
```diff
@@ -26,3 +26,5 @@ find_library(COCOA_LIBRARY Cocoa)
 target_link_libraries(p3osxdisplay
   ${APPLICATIONSERVICES_LIBRARY} ${CARBON_LIBRARY} ${AGL_LIBRARY}
   ${CORESERVICES_LIBRARY} ${COCOA_LIBRARY})
+
+install(TARGETS p3osxdisplay DESTINATION lib)
```
CFSworks: Ditto missing newline

panda/src/display/CMakeLists.txt
```diff
@@ -90,7 +90,9 @@ composite_sources(p3display P3DISPLAY_SOURCES)
 add_component_library(p3display SYMBOL BUILDING_PANDA_DISPLAY
   ${P3DISPLAY_HEADERS} ${P3DISPLAY_SOURCES})
 target_link_libraries(p3display p3cull p3pgraphnodes)
-target_use_packages(p3display X11)
+if(HAVE_X11)
+  target_use_packages(p3display X11)
+endif()
```
CFSworks: This is redundant, since `target_use_packages` actually already performs this very check.

If X11 headers are used by p3display when HAVE_X11 is on (which leaves a sour taste in my mouth for other reasons), the best thing to do would be to leave it as you had it.

treamology: Hah, shows how much I know about cmake. I'll change that back.

treamology:
I'm working on getting a travis build working. I feel like we should exclude a few of the matrix combinations for macOS since Travis-CI only provides 2 jobs at a time, which will increase the length of a build quite a bit. I have no clue which ones are important or if this is even an issue, what are your thoughts?

panda/src/display/CMakeLists.txt
CFSworks: No worries - that's actually a macro defined in PackageConfig.cmake and not part of CMake itself. :)

CFSworks:
To be honest I went a bit overboard with the build matrix. The important thing is that we test each option, and not necessarily each combination of options. I went with a bunch of different combinations, because I wasn't sure what was likely to uncover problems. We could probably reduce the matrix to only:

- Python 3 with metalibs=on, composite source limit 30, on Linux, Clang (our baseline build)
- First, but with Python 2.7
- First, but with metalibs=off
- First, but with composite disabled
- The 4 above, but on macOS instead of Linux
- First, but with GCC instead of Clang (We could make GCC the baseline, but I'm something of a Clang fan)

treamology:
Took quite a while, but I think it's in a good spot now. For some reason, the GLSL tests aren't being skipped like they are on linux, causing a segfault when they try to run. The build configuration itself should be good to go, though.

.travis/install_macos.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+brew install python@3 eigen libtar fftw libsquish zlib ffmpeg freetype bullet \
+wxmac opencv assimp libvorbis openssl@1.0 || true
```
rdb: What is wxmac needed for?  Panda3D does not use wxWidgets.

.travis/script_macos.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+if [[ "$PYTHON_INTERP" == "python2.7" ]]; then
+  export PY2_CMAKE_ARGS="-DPYTHON_EXECUTABLE=/usr/local/bin/python -DPYTHON_LIBRARY=/usr/local/opt/python@2/Frameworks/Python.framework/Versions/2.7/lib/libpython2.7.dylib"
```
rdb: I'm not a big fan of this; if there is something wrong in the Python detection in CMake, we should fix it in the CMake scripts, unless this is a type of misconfiguration that only happens on Travis and nowhere else.

.travis.yml
```diff
@@ -50,15 +55,27 @@ before_install:
 - makepanda/selfdestruct.py --yes
 
 install:
-- virtualenv --python=${PYTHON_INTERP:-python3.4} venv && source venv/bin/activate
+- >
+  if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then 
+    ./.travis/install_macos.sh;
+  else
+    virtualenv --python=${PYTHON_INTERP:-python3} venv;
+  fi
```
rdb: I would rather either put both platforms in separate scripts or put both platforms in the .travis.yml.  I'm strongly leaning towards having everything in .travis.yml.

.travis/script_macos.sh
```diff
@@ -0,0 +1,9 @@
...
+fi
+
+cmake -DHAVE_CARBON=NO -DOPENSSL_ROOT_DIR=/usr/local/Cellar/openssl/1.0.2o_1 \
```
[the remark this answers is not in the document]
treamology: By default, CMake will always select the built-in crypto libraries even when homebrew's version is present. Using Apple's crypto libraries is probably fine anyways, so I can just remove that bit altogether.

.travis/install_macos.sh
treamology: You're right, not sure where I got that idea from.

.travis.yml
treamology: The reason I was hesitant on that was because it might be a pain to edit shell scripts embedded inside a .yml file, but you're probably right that having a bunch of small files littered about would be even more of one.

direct/src/showbase/CMakeLists.txt
```diff
@@ -1,5 +1,15 @@
...
+set(P3SHOWBASE_FILES
+    showBase.cxx
+    showBase.h)
+set(P3SHOWBASE_EXTRA_ARGS "")
+if(IS_OSX) 
+  list(APPEND P3SHOWBASE_FILES showBase_assist.mm)
+endif()
+
+add_component_library(p3showbase SYMBOL BUILDING_DIRECT_SHOWBASE ${P3SHOWBASE_FILES})
+if(APPLE)
```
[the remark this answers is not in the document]
treamology: The first conditional is for something that's really only applicable for macOS but not any other Apple platform, while the second would be potentially useful for both macOS and iOS. I realize macOS is the only Apple platform Panda supports at the moment, but I figured the distinction would be useful down the road when/if iOS starts being supported.

.travis/script_macos.sh
treamology: When searching for Python 2, CMake seemingly refuses to select any install other than the system python. This seems to be a long-standing issue and I haven't been able to find a clean solution. Maybe @CFSworks can chime in on this.

direct/src/showbase/CMakeLists.txt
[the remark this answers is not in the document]
CFSworks: [CMake uses APPLE to indicate macOS](https://cmake.org/cmake/help/v3.0/variable/APPLE.html) and the `IS_OSX` is the preprocessor macro indicating the same. They're largely interchangeable.

I'm wondering if we should remove `IS_OSX` from CMake (detect it the same way WIN32_VC is detected) and use `APPLE` everywhere in CMake.

.travis/install_macos.sh
CFSworks: That one's probably on me - I [had CMake checking for WxWidgets](https://github.com/panda3d/panda3d/commit/2d5d1400be58d86a6b42be1dad98b141be74422e) until a week ago.

direct/src/showbase/CMakeLists.txt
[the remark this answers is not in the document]
rdb: @treamology but it's silly to use different variables here, because the only reason the `-ObjC++` flag is passed is because an `.mm` (Objective-C++) file is included.

IIRC `IS_OSX` is a remnant from the ppremake days, and I think that ppremake used to also define `IS_OSX` for the iOS build of Panda.  So I suggest that we just use `APPLE` for both here.

treamology:
It ends up that Apple has deprecated the use of their built-in OpenSSL by not including headers for it as of OSX 10.11. On my system CMake isn't having trouble picking up the homebrew installed headers, but it seems to need some help in Travis which is why that hint is in there.

There is also a rewrite of FindPython* in the upcoming version of CMake that fixes the issue of always selecting the system python. As a stopgap measure I kept the Python 2 hint in the build script.

A solution still needs to be devised for the failing GLSL test scripts.

rdb:
We don't need to do the build with `COMPOSITE_SOURCE_LIMIT=0` on macOS.  It's just to detect missing includes, and there is not much additional value in doing it both on macOS and Linux.

The crash in the display test is addressed by 57602d750f16e5b7029385fa5f5fcf52147f5f70 on master and should be merged onto the cmake branch.

If the latest CMake version has a fixed FindPython.cmake script, we could consider putting it in the `cmake/modules` directory.

CFSworks:
> We don't need to do the build with `COMPOSITE_SOURCE_LIMIT=0` on macOS.  It's just to detect missing includes, and there is not much additional value in doing it both on macOS and Linux.

I'm really hesitant to lose this build job because it's currently the only way to detect missing includes in files that only build on macOS.

The options I can think of for resolving this are:
1. A special `COMPOSITE_SOURCE_LIMIT` variable that only affects platform-specific components. (Which doesn't need to be documented, since it won't be very useful to users.)
2. Do the builds with ccache, taking advantage of Travis-CI's [caching feature](https://docs.travis-ci.com/user/caching/) to preserve the ccache directory between jobs. We might need a 40-minute time limit on the build step to ensure that an uncached job fails early enough to upload its cache before termination.

While we're on the subject of `COMPOSITE_SOURCE_LIMIT`, we should really do some experiments to find a sensible default. A couple of years ago I actually found that, for my workstation, the ideal setting was 7, as anything more created chokepoints in the build parallelism. 30 seems like it might be a bit high.

PR #324, opened 2018-05-23
delete unused grid code

manoguy535:
this deletes the unused grid code from panda3d. this code was only used by the disney pirates game which closed down and its so specific that i dont believe this would hurt anyones code. leaving it in just makes it hard for people to use distributednode.


CFSworks:
I'm torn on this.

I know there's still folks out there using the grid system, but I think they all switched to internal forks of Panda3D years ago. Even if nobody actively uses the upstream copy, there's the broader question of whether future projects *could* use it (as seems to be the current state of the whole `direct.distributed` package, come to think of it). Cartesian grid interests are certainly very useful in any kind of open-world environment, where there are otherwise no obvious boundaries for grouping objects by visibility.

I think this PR speaks to a much larger problem, which is "nobody wants to touch `direct.distributed` because nobody knows how all of the functionality is used" - basically, we have poor (i.e. no) code coverage from the samples/tests, which means maintainers have low confidence in making changes to that package, which means bit rot. I tend to agree that if something isn't actively tested - either internally or by a project that actively follows master and provides quick feedback - it's an undue burden on upstream Panda3D to be carrying rotting code around.

What makes this really difficult is that `direct.distributed` is somewhat "specialty" code. It was written with a very particular architecture in mind. The CMU/FOSS Panda developers had no interest in that architecture - the closest thing `direct.distributed` has to a FOSS maintainer would be the Astron project, which is progressing somewhat more slowly than I'd like right now. But, being "specialty" code, only a few users of Panda3D actively rely on it. It's just not general enough for mainstream adoption.

I think this boils down to a question of what direct's role should be. Panda is already a fantastic 3D engine, but has pretty poor feature support for games in particular: save/load, entity management, networking, menus, etc. There's talks about fixing this, and implementing a lot of the framework in C++ land instead, which means eventual deprecation for `direct` anyway. Speaking as someone who actively uses `direct.distributed` and the .dc format, I actually really dislike it. It has several design faults, it's hard to understand, it's even harder for a beginning (or even intermediate) user to pick up and make a multiplayer game, it encourages insecure design patterns, and I do think the whole thing should be condemned.

The problem is, because there isn't even an alternative to `direct.distributed` yet, I'm really wary about removing any functionality from it, which means the best option is instead to increase its code coverage (within whatever battery of tests `direct` does/should have) and delete whatever remains uncovered, rather than try to trim the fat upfront without fully understanding what's being deleted. Once there's a better alternative to `direct.distributed`, we can then deprecate the whole package altogether.

@manoguy535 Your point about the dependency on the grid system from DistributedNode is well taken, however the current way of using DistributedNode without grid behavior is simply not to parent the node under a grid parent. If the extra grid logic seems unnecessary, perhaps we could consider moving it to a `DistributedGridNode` subclass (or a mixin), which downstream projects can use instead when the node is (sometimes) on a grid?

@rdb What are your thoughts on the purpose of `direct.distributed` and/or the .dc syntax going forward? Should there be a roadmap for test coverage and/or deprecation?

rdb:
I'm hesitant to remove anything for now, perhaps not until Panda 2.0.  My policy is generally not to remove things that may be in use if there is no straightforward replacement, since I don't want to give anyone a good reason not to upgrade to the latest Panda.

With regards to the larger problem: any attempt I've made in the past to clean up or deprecate the rotten parts of direct has generally gotten me quite some pushback. That said, it is quite a burden on Panda, so I am inclined to want to just be done with (this part of) direct by splitting it off to a separate optional package on PyPI.  Those who need it can easily install it with pip, and it could be maintained by people who are masochistic enough to want to do such a thing.

-1 on the removal of DistributedNode, anyway, since that seems to be used even outside Disney's legacy; there is indeed some use of the distributed system within the broader FOSS community that we must not forget about.

manoguy535:
@rdb i did not remove distributednode, i just removed unused grid parts of it. maybe alternate solution is to make a pull request to merge astron/panda3d into an astron branch in here?

montreal91:
@rdb, maybe, it could be a good thing in general: to separate one big Panda project to bunch of different modules and do development in [pallets/flask](https://github.com/pallets/flask) style. The core is small, simplistic and provides good API for writing extensions. This have two benefits: core team is focused on maintaining the core features do not have to think about which template engine to use and so on. Users can choose whatever they they want and easily implement lacking features and choose one of many solutions.

I think, this approach can work with game engine as well, as game engine has many different modules: like renderer, GUI, physics, etc and can simplify development a lot.

BMaxV:
To sum this up

- tests are missing to make sure this doesn't break anything
- removing stuff nobody understands may be bad

Can we formulate a todo from this?

rdb:
I'm closing this since there's no clear consensus on the matter.  I'm leaning towards deferring this decision to when we split off direct in the future.  But we can reopen this if needed.

voarsh2:
Thought I'd comment to chime in - that I very much use direct/distributed, TLOPO uses the grid system...

Haven't seen them on here in a long time - and likely they've their own Panda3D fork...

I myself am making a sci-fi themed game based on Astron / DC and using direct distributed... (and make heavy use of it)...

Happy to hear direct might be packaged off - but fear that it's still going to rot away...

And yeah, there's no alternative for what direct offers, especially for multiplayer games...  can to low level CR stuff... or use the whole Astron/DC system, which is larger scale...

In the case of rotting direct, I suppose I'd have to look at versioning it myself internally - which would be a shame because other's may benefit from improved foundations...

loblao:
Forgive me for necroposting, but let me add my 2 cents, even though I haven't recently collaborated on any project that uses grid code.

Yes, TLOPO does use an internal fork. That said, if this piece of code gets removed, you are going to impose a major burden on them the next time they need to pull from upstream in order to benefit from a bug fix or performance improvement change (and that happens quite often, if I'm not mistaken).

I'd defer this change until it's time to deprecate/move `direct` as a whole.

rdb:
Thanks for the comments. Sounds like we're converging on the side of "do not remove" wrt the grid code.

PR #327, opened 2018-05-24
Update checkPandaVersion.h

CFSworks:
See relevant discussion [here](https://github.com/panda3d/panda3d/commit/d96765b9575665b72defacb72a2ee80483f934e8).

This removes the requirement for `checkPandaVersion.h` to be included only once, and moves its inclusion to `dtoolbase_cc.h`, which ensures that every Panda binary includes the check.

This uses a workaround allowed by the C++ ODR, namely that template expansions can contain multiple definitions of the same template function as long as all definitions are the same. In C++17, we'll be replacing this with inline variables. There is a note in the CMake buildsystem to remind us (loudly) once we bump the standard version to C++17.

Comparing a build of this branch vs. master:

- On master, only the following files contained a reference to the version symbol:
```
Binary file bin/pview matches
Binary file lib/x86_64-linux-gnu/panda3d/libpandagl.so matches
Binary file lib/x86_64-linux-gnu/panda3d/libp3dtool.so.1.10 matches
Binary file lib/x86_64-linux-gnu/panda3d/libpanda.so.1.10 matches
Binary file lib/x86_64-linux-gnu/panda3d/libpandaegg.so.1.10 matches
Binary file lib/x86_64-linux-gnu/panda3d/libpandaphysics.so.1.10 matches
Binary file lib/x86_64-linux-gnu/panda3d/libp3framework.so.1.10 matches
Binary file lib/x86_64-linux-gnu/panda3d/libpandabullet.so.1.10 matches
Binary file lib/x86_64-linux-gnu/panda3d/libpandaode.so.1.10 matches
Binary file lib/x86_64-linux-gnu/panda3d/libpandafx.so.1.10 matches
Binary file lib/x86_64-linux-gnu/panda3d/libpandaexpress.so.1.10 matches
```

- On this branch, all ELF binaries now contain a reference.

- Size increase for each library is negligible (101 bytes in the worst case)

### DIFF
--- a/dtool/src/dtoolbase/dtoolbase_cc.h
+++ b/dtool/src/dtoolbase/dtoolbase_cc.h
@@ -19,6 +19,13 @@
 
 #ifdef __cplusplus
 
+// By including checkPandaVersion.h, we guarantee that runtime attempts to
+// load any DLL will fail if they inadvertently link with the wrong version of
+// dtool, which, transitively, means all DLLs must be from the same
+// (ABI-compatible) version of Panda.
+
+#include "checkPandaVersion.h"
+
 #ifdef USE_TAU
 // Tau provides this destructive version of stdbool.h that we must mask.
 #define __PDT_STDBOOL_H_
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -2606,18 +2606,21 @@ PANDAVERSION_H_RUNTIME="""
 
 CHECKPANDAVERSION_CXX="""
 # include "dtoolbase.h"
-EXPCL_DTOOL_DTOOLUTIL int panda_version_$VERSION1_$VERSION2 = 0;
+EXPCL_DTOOL_DTOOLBASE int panda_version_$VERSION1_$VERSION2 = 0;
 """
 
 CHECKPANDAVERSION_H="""
+# ifndef CHECKPANDAVERSION_H
+# define CHECKPANDAVERSION_H
 # include "dtoolbase.h"
-extern EXPCL_DTOOL_DTOOLUTIL int panda_version_$VERSION1_$VERSION2;
-# ifndef WIN32
-/* For Windows, exporting the symbol from the DLL is sufficient; the
-      DLL will not load unless all expected public symbols are defined.
-      Other systems may not mind if the symbol is absent unless we
-      explictly write code that references it. */
-static int check_panda_version = panda_version_$VERSION1_$VERSION2;
+extern EXPCL_DTOOL_DTOOLBASE int panda_version_$VERSION1_$VERSION2;
+// Hack to forcibly depend on the check
+template<typename T>
+class CheckPandaVersion {
+public:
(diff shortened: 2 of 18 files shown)


CFSworks:
@rdb The macros in `dtoolsymbols.h` aren't valid until `dtoolbase.h` has had a chance to get included anyway - if we're okay with implicitly relying on `dtoolbase.h` getting pulled in earlier in the translation unit, I'm just going to remove the include altogether.

rdb:
No strong objections from me.  I'll just do a test build on macOS to be sure.

rdb:
On second thought, not including `dtoolbase.h` from `checkPandaVersion.h` might cause issues for people who include `checkPandaVersion.h` at the top of their own files.  If that's the choice, I would prefer the relatively harmless include loop.

CFSworks:
Alright; I'm gonna force that last commit off so we can merge down what I had before today.

If we decide we don't like the include loop, we can always change where we're pulling in `checkPandaVersion.h` later.

Let me know when your macOS build finishes.

rdb:
Ah, drat, it does fail on the Mac buildbot.  Sorry for not catching this in my test build.
```
[T1] Building C++ object built-py3.6/tmp/p3showbase_showBase.o
clang++ -std=gnu++11 -ftemplate-depth-70 -fPIC -c -o built-py3.6/tmp/p3showbase_showBase.o -Ibuilt-py3.6/tmp -Ibuilt-py3.6/include -I/Users/buildbot/thirdparty/darwin-libs-a/eigen/include -Idirect/src/showbase -fvisibility=hidden -Wno-deprecated-declarations -isysroot /Developer/SDKs/MacOSX10.6.sdk -mmacosx-version-min=10.6 -arch i386 -arch x86_64 -pthread -fno-exceptions -msse2 -fno-strict-aliasing -ffast-math -fno-stack-protector -fno-finite-math-only -O2 -DBUILDING_DIRECT direct/src/showbase/showBase.cxx
In file included from direct/src/showbase/showBase.cxx:22:
In file included from direct/src/showbase/showBase.h:17:
In file included from built-py3.6/include/directbase.h:21:
In file included from built-py3.6/include/pandabase.h:21:
In file included from built-py3.6/include/dtoolbase.h:516:
In file included from built-py3.6/include/dtoolbase_cc.h:27:
built-py3.6/include/checkPandaVersion.h:10:15: error: expected member name or ';' after declaration specifiers
  int check() { return panda_version_1_10; }
  ~~~         ^
1 error generated.
Process exited with exit status 0 and signal code 1
```

rdb:
I have fixed this in 239f66cf1144d866c4d12c1cf483b5e4055982ef; it seems just renaming the method is enough.  I'm guessing there's some macro called `check` in the macOS headers that was causing this error, or something like that.

CFSworks:
@rdb Thanks for the fix! I'll do the same in the CMake branch.

But also... that's very strange for macOS. One would think `check()` to be an extremely common method name and a macro (or whatever) that interferes with that would get caught right away. Oh well, it is what it is I guess.

PR #329, opened 2018-05-26
express: Readd missing Datagram constructors

darktohka:
Some game code relies on this functionality. Even though it's possible to rewrite the logic by constructing an empty Datagram and assigning the data to it manually, it's more convenient to have a constructor for it.

This code was removed in e9ae7dcc407d8198a701b761d0431f456c8ad925, but I feel like it should be added back.

### DIFF
--- a/panda/src/express/datagram.I
+++ b/panda/src/express/datagram.I
@@ -53,6 +53,25 @@ Datagram(vector_uchar data) :
 }
 
 /**
+ * Constructs a datagram from an existing datagram.
+ */
+INLINE Datagram::
+Datagram(const Datagram &copy) :
+  _data(copy._data),
+  _stdfloat_double(copy._stdfloat_double)
+{
+}
+
+/**
+ * Assigns another datagram's data to the datagram.
+ */
+INLINE void Datagram::
+operator = (const Datagram &copy) {
+  _data = copy._data;
+  _stdfloat_double = copy._stdfloat_double;
+}
+
+/**
  * Adds a boolean value to the datagram.
  */
 INLINE void Datagram::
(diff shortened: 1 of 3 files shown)


CFSworks:
This looks like it just adds back the copy constructor and assignment operator. Looks good to me.

Could you also add a test case to `tests/putil/test_datagram.py`?

CFSworks:
@rdb I wrote [a pair of XFAIL tests](https://github.com/panda3d/panda3d/commit/339e1ce4d861d15084fd7f099d84a8ec7ba3dd88#diff-eb4c6d688e8a7a1fb590dbf9c49d0815R99) that show what I believe is the behavior @darktohka is trying to fix (indeed applying this PR to my local branch makes these tests XPASS).

Although from your reply it sounds like this might be an Interrogate bug and not an unintended side-effect of the e9ae7dc change.

rdb:
@cfsworks yes, interrogate should be recognising that there is an implicit copy constructor and implicit copy assignment operator, but if it does not, that is a bug.

A better approach than defining them with a user-defined body is to just explicitly declare copy/move constructor/assignment operator as defaulted, so that we can ensure that they can are parsed.  I will check in such a change.

rdb:
Better, but the explicit copy constructor declaration makes the compiler omit generation of a move constructor, which we do want in the case of Datagram.

I have already made some changes locally and will check them in shortly.

theclashingfritz:
Thanks for adding these back. It fixes some code that was broken in my projects because of their removal.

rdb:
Please let me know ASAP when any new change to Panda3D breaks existing code.

PR #332, opened 2018-05-28
discussion: Are we abusing NULL?

CFSworks:
This PR isn't so much a PR intended for inclusion as it is for indicating several bizarre asserts I'd like to discuss.

In particular, the asserts identified are:
1. Comparing `this` to `NULL`
2. Taking the address of a variable and comparing to `NULL`
3. Dereferencing `NULL` because of a failed assert

In many cases I understand what the asserts are trying to do; but these are still bad code smells.

For example, the asserts testing `this != NULL` are probably trying to verify that the destructor doesn't run twice. However, in those circumstances, `this` wouldn't be `NULL`: it would be a dangling pointer. The only thing this would catch is somebody casting NULL to a class pointer and then deleting it - and even then I think C++11's delete operator checks this for you (it ignores deletion of null).

### DIFF
--- a/dtool/src/interrogate/interfaceMaker.cxx
+++ b/dtool/src/interrogate/interfaceMaker.cxx
@@ -665,7 +665,6 @@ record_object(TypeIndex type_index) {
 
   InterrogateDatabase *idb = InterrogateDatabase::get_ptr();
   const InterrogateType &itype = idb->get_type(type_index);
-  assert(&itype != NULL);
 
   Object *object = new Object(itype);
   bool inserted = _objects.insert(Objects::value_type(type_index, object)).second;
--- a/dtool/src/interrogate/interrogateBuilder.cxx
+++ b/dtool/src/interrogate/interrogateBuilder.cxx
@@ -1696,8 +1696,6 @@ get_function(CPPInstance *function, string description,
     InterrogateFunction &ifunction =
       InterrogateDatabase::get_ptr()->update_function(index);
 
-    nassertr(&ifunction != NULL, 0);
-
     ifunction._flags |= flags;
 
     // Also, make sure this particular signature is defined.
--- a/panda/src/express/nodeReferenceCount.I
+++ b/panda/src/express/nodeReferenceCount.I
@@ -52,8 +52,6 @@ NodeReferenceCount(const NodeReferenceCount &copy) : ReferenceCount(copy) {
  */
 INLINE void NodeReferenceCount::
 operator = (const NodeReferenceCount &copy) {
-  nassertv(this != NULL);
-
   // If this assertion fails, our own pointer was recently deleted.  Possibly
   // you used a real pointer instead of a PointerTo at some point, and the
   // object was deleted when the PointerTo went out of scope.  Maybe you tried
@@ -74,8 +72,6 @@ operator = (const NodeReferenceCount &copy) {
  */
 INLINE NodeReferenceCount::
 ~NodeReferenceCount() {
-  nassertv(this != NULL);
-
   // If this assertion fails, we're trying to delete an object that was just
   // deleted.  Possibly you used a real pointer instead of a PointerTo at some
   // point, and the object was deleted when the PointerTo went out of scope.
--- a/panda/src/express/nodeReferenceCount.cxx
+++ b/panda/src/express/nodeReferenceCount.cxx
@@ -21,8 +21,6 @@ TypeHandle NodeReferenceCount::_type_handle;
  */
 bool NodeReferenceCount::
 do_test_ref_count_integrity() const {
-  nassertr(this != NULL, false);
-
   // If this assertion fails, we're trying to delete an object that was just
   // deleted.  Possibly you used a real pointer instead of a PointerTo at some
   // point, and the object was deleted when the PointerTo went out of scope.
--- a/panda/src/express/referenceCount.I
+++ b/panda/src/express/referenceCount.I
@@ -63,8 +63,6 @@ ReferenceCount(const ReferenceCount &) {
  */
 INLINE void ReferenceCount::
 operator = (const ReferenceCount &) {
-  nassertv(this != NULL);
-
   // If this assertion fails, our own pointer was recently deleted.  Possibly
   // you used a real pointer instead of a PointerTo at some point, and the
   // object was deleted when the PointerTo went out of scope.  Maybe you tried
@@ -80,7 +78,6 @@ operator = (const ReferenceCount &) {
 ReferenceCount::
 ~ReferenceCount() {
   TAU_PROFILE("ReferenceCount::~ReferenceCount()", " ", TAU_USER);
-  nassertv(this != NULL);
 
   // If this assertion fails, we're trying to delete an object that was just
   // deleted.  Possibly you used a real pointer instead of a PointerTo at some
--- a/panda/src/express/referenceCount.cxx
+++ b/panda/src/express/referenceCount.cxx
@@ -23,8 +23,6 @@ TypeHandle ReferenceCount::_type_handle;
  */
 bool ReferenceCount::
 do_test_ref_count_integrity() const {
-  nassertr(this != NULL, false);
-
   // If this assertion fails, we're trying to delete an object that was just
   // deleted.  Possibly you used a real pointer instead of a PointerTo at some
   // point, and the object was deleted when the PointerTo went out of scope.
--- a/panda/src/express/weakPointerTo.I
+++ b/panda/src/express/weakPointerTo.I
@@ -45,7 +45,6 @@ WeakPointerTo(const WeakPointerTo<T> &copy) :
 template<class T>
 INLINE TYPENAME WeakPointerTo<T>::To &WeakPointerTo<T>::
 operator *() const {
-  nassertr(!this->was_deleted(), *((To *)NULL));
   return *((To *)WeakPointerToBase<T>::_void_ptr);
 }
 
@@ -208,7 +207,6 @@ WeakConstPointerTo(const WeakConstPointerTo<T> &copy) :
 template<class T>
 INLINE const TYPENAME WeakConstPointerTo<T>::To &WeakConstPointerTo<T>::
 operator *() const {
-  nassertr(!this->was_deleted(), *((To *)NULL));
   return *((To *)WeakPointerToBase<T>::_void_ptr);
 }
 
--- a/panda/src/putil/cachedTypedWritableReferenceCount.I
+++ b/panda/src/putil/cachedTypedWritableReferenceCount.I
@@ -49,8 +49,6 @@ CachedTypedWritableReferenceCount(const CachedTypedWritableReferenceCount &copy)
  */
 INLINE void CachedTypedWritableReferenceCount::
 operator = (const CachedTypedWritableReferenceCount &copy) {
-  nassertv(this != NULL);
-
   // If this assertion fails, our own pointer was recently deleted.  Possibly
   // you used a real pointer instead of a PointerTo at some point, and the
   // object was deleted when the PointerTo went out of scope.  Maybe you tried
@@ -71,8 +69,6 @@ operator = (const CachedTypedWritableReferenceCount &copy) {
  */
 INLINE CachedTypedWritableReferenceCount::
 ~CachedTypedWritableReferenceCount() {
-  nassertv(this != NULL);
-
   // If this assertion fails, we're trying to delete an object that was just
   // deleted.  Possibly you used a real pointer instead of a PointerTo at some
   // point, and the object was deleted when the PointerTo went out of scope.
--- a/panda/src/putil/cachedTypedWritableReferenceCount.cxx
+++ b/panda/src/putil/cachedTypedWritableReferenceCount.cxx
@@ -21,8 +21,6 @@ TypeHandle CachedTypedWritableReferenceCount::_type_handle;
  */
 bool CachedTypedWritableReferenceCount::
 do_test_ref_count_integrity() const {
-  nassertr(this != NULL, false);
-
   // If this assertion fails, we're trying to delete an object that was just
   // deleted.  Possibly you used a real pointer instead of a PointerTo at some
   // point, and the object was deleted when the PointerTo went out of scope.
--- a/panda/src/putil/nodeCachedReferenceCount.I
+++ b/panda/src/putil/nodeCachedReferenceCount.I
@@ -49,8 +49,6 @@ NodeCachedReferenceCount(const NodeCachedReferenceCount &copy) : CachedTypedWrit
  */
 INLINE void NodeCachedReferenceCount::
 operator = (const NodeCachedReferenceCount &copy) {
-  nassertv(this != NULL);
-
   // If this assertion fails, our own pointer was recently deleted.  Possibly
   // you used a real pointer instead of a PointerTo at some point, and the
   // object was deleted when the PointerTo went out of scope.  Maybe you tried
@@ -71,8 +69,6 @@ operator = (const NodeCachedReferenceCount &copy) {
  */
 INLINE NodeCachedReferenceCount::
 ~NodeCachedReferenceCount() {
-  nassertv(this != NULL);
-
   // If this assertion fails, we're trying to delete an object that was just
   // deleted.  Possibly you used a real pointer instead of a PointerTo at some
   // point, and the object was deleted when the PointerTo went out of scope.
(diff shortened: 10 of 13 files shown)


dtool/src/interrogate/interfaceMaker.cxx
```diff
@@ -665,7 +665,6 @@ record_object(TypeIndex type_index) {
 
   InterrogateDatabase *idb = InterrogateDatabase::get_ptr();
   const InterrogateType &itype = idb->get_type(type_index);
-  assert(&itype != NULL);
```
CFSworks: It seems like we should be asserting that the database has the type before trying to get it, or having get_type return a pointer if it could be "nullable." Running around with a reference to a dereferenced NULL is a Bad Thing(TM).

rdb:
While we are already in UB territory if `this == NULL`, I do think there is value in having that check in specific circumstances.  Asserts are intended to reveal when something has gone significantly wrong, in a way that is more helpful than simply crashing (and we really want to avoid Panda crashing without leaving any kind of hints).  Usually these `this != NULL` asserts are in places where there is a significant risk that erroneous code ends up calling that method on a NULL pointer, so it is a useful and harmless way to expose these bugs where it makes sense.

Dereferencing NULL in a `nassertr` is bad.  We should not do that; if we really want to crash the application if the assert fails, we should just use the regular `assert()` function which calls `abort()`.

While on the subject, I think we should replace all instances of `NULL` and `(Type *)NULL` with `nullptr`.

CFSworks:
>While on the subject, I think we should replace all instances of NULL and (Type *)NULL with nullptr.

Already working on that in a local branch! But my compiler throws tons of warnings about `this != nullptr`, claiming this is tautological (which would imply the C++11 standard guarantees `this != nullptr` already).

rdb:
It does, but UB throws all guarantees out the water, so that doesn't mean that it can't actually be nullptr if a method is called on a null pointer.

Hmm, do compilers actually optimize out these comparisons?  Would this be different if we left in regular NULL in here for these comparisons?

panda/src/express/weakPointerTo.I
```diff
@@ -208,7 +207,6 @@ WeakConstPointerTo(const WeakConstPointerTo<T> &copy) :
 template<class T>
 INLINE const TYPENAME WeakConstPointerTo<T>::To &WeakConstPointerTo<T>::
 operator *() const {
-  nassertr(!this->was_deleted(), *((To *)NULL));
```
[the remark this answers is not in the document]
rdb: The `*` operator is not thread safe; I have recently superseded it with the `lock()` method, which is a safe (but slow) method of accessing a weak pointer.  But there might be some use case where you know for sure that the object is not yet destroyed, and you need a fast way of accessing the pointer.

In this case, we could consider using `assert()` which will abort the application.  Terminating a game is not nice, but it's probably what would have happened anyway if the caller did anything with the returned pointer.

For comparison, the C++11 weak_ptr throws an exception in this situation.

dtool/src/interrogate/interrogateBuilder.cxx
```diff
@@ -1696,8 +1696,6 @@ get_function(CPPInstance *function, string description,
     InterrogateFunction &ifunction =
       InterrogateDatabase::get_ptr()->update_function(index);
 
-    nassertr(&ifunction != NULL, 0);
```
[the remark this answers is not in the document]
rdb: This is a case where it is never supposed to return a null reference to begin with, and the assert was just added to save me a lot of time in the debugger in case it happened again.

I agree that this is kind of smelly and we could probably improve the InterrogateDatabase interface altogether (though I don't really think it's high priority).

dtool/src/interrogate/interfaceMaker.cxx
rdb: Yes, I agree with this.

panda/src/speedtree/speedTreeNode.cxx
```diff
@@ -145,9 +145,6 @@ count_total_instances() const {
  */
 SpeedTreeNode::InstanceList &SpeedTreeNode::
 add_tree(const STTree *tree) {
-  nassertr(is_valid(), *(InstanceList *)NULL);
-  nassertr(tree->is_valid(), *(InstanceList *)NULL);
```
rdb: This is bad (but I'm afraid of touching the SpeedTree code because I don't have a licence to compile it with).  Some code in Panda allocates a static const version of the class to return a reference to in case of assertions.

CFSworks:
> Hmm, do compilers actually optimize out these comparisons? Would this be different if we left in regular NULL in here for these comparisons?

Both compilers on my system (Clang 5.0.1 and GCC 7.3.0) assume (at -O2, not at -O0) that `this` is nonzero. Comparing to NULL, nullptr, or even just using a bare `if` statement, the `this` keyword, directly or indirectly (by assigning to a variable first), results in the optimizer assuming `this != 0`. I can't speak for Windows and macOS, but I can't imagine the story is much different.

I have to ask, though: Aren't we reinventing the wheel here? The CPU+OS already work together to ensure no NULL pointers are accessed, don't they? And dereferencing a NULL class pointer is one of the more common faults in C++ software - every C++ debugger or memory profiler I've ever used includes some way of identifying when you've accessed NULL.

rdb:
Do they?  Wow.  I wonder if this is a recent feature, or if my own tests at the time were done with MSVC.

If they really are being optimized out then I suppose there is little value in keeping them.  I just also discovered that there is the `-fsanitize=null` option, which adds automatic checks for those things, which we could consider enabling in debug builds (and in CI builds!).  This would let us catch null dereferences even when there don't cause crashes.

> Aren't we reinventing the wheel here? The CPU+OS already work together to ensure no NULL pointers are accessed, don't they?
Yes, by triggering a segmentation fault and terminating the program.  Unfortunately most operating systems (macOS being the exception here) give almost no information about where such a thing occurs, so having asserts is sometimes the only way we get any useful information at all when users report crashes.
Furthermore, sometimes the crash happens only much later when the reference has been passed around a bunch of times, so such an assert can save a lot of time backtracking through code with debugging tools.

CFSworks:
> If they really are being optimized out then I suppose there is little value in keeping them.  I just also discovered that there is the `-fsanitize=null` option, which adds automatic checks for those things, which we could consider enabling in debug builds (and in CI builds!).  This would let us catch null dereferences even when there don't cause crashes.

I really would like to make much better use of the Clang sanitizers and toolchain; it's top-notch and I'm trying to integrate things into the CMake branch wherever I can manage.

> Yes, by triggering a segmentation fault and terminating the program.  Unfortunately most operating systems (macOS being the exception here) give almost no information about where such a thing occurs, so having asserts is sometimes the only way we get any useful information at all when users report crashes.

The OS triggers the segmentation fault, but doesn't terminate the program: the program gets to decide what to do from there. The default is usually to display an error and exit (and sometimes generate a memory dump). Any attached debuggers will catch the fault, though, and pinpoint exactly where the problem happened, and generate a traceback, which is much more information than just "an assert failed on this line in this file, good luck figuring out who called it"

> Furthermore, sometimes the crash happens only much later when the reference has been passed around a bunch of times, so such an assert can save a lot of time backtracking through code with debugging tools.

I have to assume you're speaking generally, because the 12 asserts I identified are all immediately followed by an implicit `this` dereference (in a future assert) so this could not happen.

If we're interested in cases where a NULL `this` is passed around, we should really be using the instrumentation available to catch that. An assert doesn't give much more information than a crash report - in some cases even less (as I mentioned above), and I'd actually prefer that Panda not try to rescue bad pointers from being caught by the segfault machinery for this reason.

rdb:
That sounds good in principle; I'm all ears for ideas on how we can leverage said instrumentation to produce useable diagnostics in the SDK builds we ship to Python developers.

Note that Panda's asserts are specifically designed to be trappable in a debugger by setting `assert-abort true` in Config.prc, which is a must-have setting for anyone who knows how to use a debugger.

CFSworks:
> That sounds good in principle; I'm all ears for ideas on how we can leverage said instrumentation to produce useable diagnostics in the SDK builds we ship to Python developers.

This is tricky. On one hand, we'd like something that's on by default for folks who may be absolute beginners to C++, but on the other hand, it should have a minimal overhead so we don't bog down deployments when 99% of them won't need that kind of instrumentation anyway.

One possibility would be to ship the SDK linked against Breakpad/Crashpad and only register the segfault handler if we'd be replacing the default (so that we don't interfere with the user's own handler if they have one). We can also strip the debug symbols off of shipped SDK builds, so we don't make the user download too much larger. Breakpad/Crashpad also support automatic uploading of crashes as they happen, but I figure our users would find that to be a privacy violation and instead prefer to submit them manually.

Another option, if we didn't like the extra dependency, would be to find some library that can do stack unwinding and reports tracebacks on its own when a segfault is detected. We'd need to ship the debug symbols for such a traceback to be useful, however.

> Note that Panda's asserts are specifically designed to be trappable in a debugger by setting `assert-abort true` in Config.prc, which is a must-have setting for anyone who knows how to use a debugger.

Alas I only just now found out about this - it doesn't seem to be documented anywhere (the only place I can find it mentioned is on the big list of PRC variables, which is sorted alphabetically instead of categorically) and this being set to false has definitely been a bane to me when using a debugger. It's a bit of platform-specific code, but couldn't we detect the debugger and default that variable to true when one is attached?

dtool/src/interrogate/interfaceMaker.cxx
CFSworks: Resolution: remove

dtool/src/interrogate/interrogateBuilder.cxx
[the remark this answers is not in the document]
CFSworks: Resolution: remove

panda/src/express/nodeReferenceCount.I
```diff
@@ -52,8 +52,6 @@ NodeReferenceCount(const NodeReferenceCount &copy) : ReferenceCount(copy) {
  */
 INLINE void NodeReferenceCount::
 operator = (const NodeReferenceCount &copy) {
-  nassertv(this != NULL);
```
CFSworks: Resolution: remove (likely optimized out on many platforms; its presence may actually hinder, rather than help, investigation of mysterious segfaults)

panda/src/express/nodeReferenceCount.I
```diff
@@ -74,8 +72,6 @@ operator = (const NodeReferenceCount &copy) {
  */
 INLINE NodeReferenceCount::
 ~NodeReferenceCount() {
-  nassertv(this != NULL);
```
CFSworks: Resolution: remove

panda/src/express/nodeReferenceCount.cxx
```diff
@@ -21,8 +21,6 @@ TypeHandle NodeReferenceCount::_type_handle;
  */
 bool NodeReferenceCount::
 do_test_ref_count_integrity() const {
-  nassertr(this != NULL, false);
```
CFSworks: Resolution: remove

panda/src/express/referenceCount.I
```diff
@@ -63,8 +63,6 @@ ReferenceCount(const ReferenceCount &) {
  */
 INLINE void ReferenceCount::
 operator = (const ReferenceCount &) {
-  nassertv(this != NULL);
```
CFSworks: Resolution: remove

panda/src/express/referenceCount.I
```diff
@@ -80,7 +78,6 @@ operator = (const ReferenceCount &) {
 ReferenceCount::
 ~ReferenceCount() {
   TAU_PROFILE("ReferenceCount::~ReferenceCount()", " ", TAU_USER);
-  nassertv(this != NULL);
```
CFSworks: Resolution: remove

panda/src/express/referenceCount.cxx
```diff
@@ -23,8 +23,6 @@ TypeHandle ReferenceCount::_type_handle;
  */
 bool ReferenceCount::
 do_test_ref_count_integrity() const {
-  nassertr(this != NULL, false);
```
CFSworks: Resolution: remove

panda/src/express/weakPointerTo.I
```diff
@@ -45,7 +45,6 @@ WeakPointerTo(const WeakPointerTo<T> &copy) :
 template<class T>
 INLINE TYPENAME WeakPointerTo<T>::To &WeakPointerTo<T>::
 operator *() const {
-  nassertr(!this->was_deleted(), *((To *)NULL));
```
CFSworks: Resolution: `assert`

(And similar for the operators that return a NULL pointer, which is just going to be dereferenced immediately by the caller - e.g. `operator ->`)

panda/src/express/weakPointerTo.I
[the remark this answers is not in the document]
CFSworks: Resolution: `assert`

panda/src/putil/cachedTypedWritableReferenceCount.I
```diff
@@ -49,8 +49,6 @@ CachedTypedWritableReferenceCount(const CachedTypedWritableReferenceCount &copy)
  */
 INLINE void CachedTypedWritableReferenceCount::
 operator = (const CachedTypedWritableReferenceCount &copy) {
-  nassertv(this != NULL);
```
CFSworks: Resolution: remove

panda/src/putil/cachedTypedWritableReferenceCount.I
```diff
@@ -71,8 +69,6 @@ operator = (const CachedTypedWritableReferenceCount &copy) {
  */
 INLINE CachedTypedWritableReferenceCount::
 ~CachedTypedWritableReferenceCount() {
-  nassertv(this != NULL);
```
CFSworks: Resolution: remove

panda/src/putil/cachedTypedWritableReferenceCount.cxx
```diff
@@ -21,8 +21,6 @@ TypeHandle CachedTypedWritableReferenceCount::_type_handle;
  */
 bool CachedTypedWritableReferenceCount::
 do_test_ref_count_integrity() const {
-  nassertr(this != NULL, false);
```
CFSworks: Resolution: remove

panda/src/putil/nodeCachedReferenceCount.I
```diff
@@ -49,8 +49,6 @@ NodeCachedReferenceCount(const NodeCachedReferenceCount &copy) : CachedTypedWrit
  */
 INLINE void NodeCachedReferenceCount::
 operator = (const NodeCachedReferenceCount &copy) {
-  nassertv(this != NULL);
```
CFSworks: Resolution: remove

panda/src/putil/nodeCachedReferenceCount.I
```diff
@@ -71,8 +69,6 @@ operator = (const NodeCachedReferenceCount &copy) {
  */
 INLINE NodeCachedReferenceCount::
 ~NodeCachedReferenceCount() {
-  nassertv(this != NULL);
```
CFSworks: Resolution: remove

panda/src/putil/nodeCachedReferenceCount.cxx
```diff
@@ -21,8 +21,6 @@ TypeHandle NodeCachedReferenceCount::_type_handle;
  */
 bool NodeCachedReferenceCount::
 do_test_ref_count_integrity() const {
-  nassertr(this != NULL, false);
```
CFSworks: Resolution: remove

panda/src/speedtree/speedTreeNode.I
```diff
@@ -47,7 +47,6 @@ get_tree(int n) const {
  */
 INLINE const SpeedTreeNode::InstanceList &SpeedTreeNode::
 get_instance_list(int n) const {
-  nassertr(n >= 0 && n < (int)_trees.size(), *(InstanceList *)NULL);
```
CFSworks: Resolution: `assert` (and add TODO comment saying this needs to be nassertr but there's no good, safe, return type so we have to use assert for now)

panda/src/speedtree/speedTreeNode.cxx
CFSworks: Resolution: `assert` (and add TODO comment saying this needs to be nassertr but there's no good, safe, return type so we have to use assert for now)

panda/src/speedtree/speedTreeNode.I
rdb: OK, though I don't know if anyone is going to ever be able to compile this code, so I'm not so worried about it.

FWIW I think we should also try to replace int with size_t in methods like this so that we get less ugly asserts, but that is a separate issue.

rdb:
I'll agree to all of those, albeit somewhat reluctantly (those asserts were not added without reason).

CFSworks:
I'll close this PR and replace it with the actual `nullptr` changes. Stand by!

PR #334, opened 2018-05-29
dtool: Cull certain issues that arise with non-glibc builds

Psychotropos:
# Summary
Currently, the dtool branch makes some assumptions as to the characteristics of the standard C library implementations it's being compiled against (mostly relating to non-GNU-compliant libdl implementations). 

This PR serves to make it deal with non-glibc stdlib implementations in a more flexible manner. Note that this was written and has been tested with musl in mind, however, it should also theoretically allow for compiling against uClibc as well.


rdb:
Committed with a few formatting fixes, thanks!

PR #336, opened 2018-05-30
Replace NULL with nullptr and remove TYPENAME macro

CFSworks:
Here it is, the first of my major refactor PRs.

This replaces (almost) all instances of NULL with nullptr, without a cast. I also tracked down several cases where 0 is being used as a null pointer and addressed those (separate commits, based on how uncertain I am that we'll ultimately want the changes merged to master).

The final cherry on top gets rid of the TYPENAME macro.

Some notes:
- I did not rebuild the .prebuilt parser/lexers. I figure this will happen in due time, and the .prebuilt files aren't intended for human consumption anyway.
- I'm particularly unsure about the ODE changes. ODE IDs *are* pointers, but this might be in "subject to future change" territory so I don't know if keeping 0 is appropriate.
- This needs to be subjected to our full battery of tests against all supported compilers. I've tested on Clang, which is quite happy with these changes, but nowhere else.


rdb:
Fantastic!  Thank you!

I think we should probably squash (most of) these when merging so that we don't just introduce a lot of wasteful intermediate blobs.

I think the ODE changes are fine; I looked to see if there is a specialised macro in the ODE headers to represent a null object but it appears not so.

Is there a reason why we have WeakPointerTo comparisons with nullptr but not any other PointerTo?  There are lots of operations we could overload for nullptr on *PointerTo, though I'm sure many of them will be handled via automatic coercion just fine, though.

CFSworks:
> Is there a reason why we have WeakPointerTo comparisons with nullptr but not any other PointerTo?

WeakPointerTo is unique among PointerTos in that it has 3 (now 4) different overloads for each operation, and the compiler throws an error because it's ambiguous to which `nullptr` is supposed to coerce. The other PointerTos only have one possible candidate, so automatic coercion handles it without problems.

> I am not sure that the ordering comparisons with nullptr are a good idea.

I don't even know why WeakPointerTo has ordering comparisons. There isn't really a semantic meaning to pointers - the memory allocator just picks whatever address is convenient during new/malloc. I can only assume it's needed for some kind of deterministic sort somewhere, which is why I just arbitrarily decided that `nullptr` would be less than any non-null pointer.

rdb:
Ordering comparisons are indeed needed for sorting, but this is only relevant for comparing with other PointerTos.  A vector cannot contain both nullptr_t and a PointerTo.

I think we should probably be defining the same operators for WeakPointerTo and PointerTo; we could probably get away with defining all comparisons on PointerToVoid since both WeakPointerTo and PointerTo inherit from that base.

CFSworks:
@rdb I can spike the 141b3d854f74a12b1c44c9dd790457737c65fee7 commit if you'd prefer to unify the PointerTo implementations.

rdb:
What is "spiking" a commit?

I was going to suggest only having comparison defined on PointerToVoid, but come to think of it, defining comparisons on (Weak)PointerToVoid is a bad idea, because this would produce the wrong result when comparing a PT(PandaNode) to a PT(SubClassNode) (since even if they point to the same object, they may have different pointer values).  The `operator *` is going to produce the right result here for PointerTo, but in the case of WeakPointerToVoid, calling `operator *` is not safe, so we cannot rely on automatic coercion.

Hmm, it's not immediately obvious to me what the best solution is here.  Perhaps we can deal with this at another time since it's a bit out-of-scope of this issue.

CFSworks:
> What is "spiking" a commit?

Sorry, I guess that's a bit of jargon that doesn't translate to software development. The verb comes from [doing this](https://n7.alamy.com/zooms/f1c2a73447f5432b931d8348c22178de/1930s-1940s-1950s-retro-office-sharp-spindle-spike-with-paper-messages-aamtey.jpg) to an editorial story. So what I'm proposing is using `git rebase -i` to axe the commit pending a more proper resolution on master.

> Hmm, it's not immediately obvious to me what the best solution is here. Perhaps we can deal with this at another time since it's a bit out-of-scope of this issue.

Well, what's on this branch does preserve the same behavior as what's on master. Perhaps we should merge (after squashing down to more like 3-4 commits) and open an issue about the PointerTo mess?

panda/src/movies/microphoneAudioDS.cxx
```diff
@@ -248,12 +248,12 @@ open() {
   format.cbSize = 0;
 
   HWAVEIN hwav;
-  MMRESULT stat = waveInOpen(&hwav, _device_id, &format, NULL, NULL, CALLBACK_NULL);
+  MMRESULT stat = waveInOpen(&hwav, _device_id, &format, nullptr, nullptr, CALLBACK_NULL);
```
rdb: Same as above.

panda/src/movies/microphoneAudioDS.cxx
```diff
@@ -146,7 +146,7 @@ find_all_microphones_ds() {
         format.nBlockAlign = 2 * chan;
         format.wBitsPerSample = 16;
         format.cbSize = 0;
-        stat = waveInOpen(NULL, i, &format, NULL, NULL, WAVE_FORMAT_QUERY);
+        stat = waveInOpen(nullptr, i, &format, nullptr, nullptr, WAVE_FORMAT_QUERY);
```
rdb: The fourth and fifth argument are `DWORD_PTR` which does not accept `nullptr` because it is an integral type.  Changing to `0` fixes the build.

panda/src/dxgsg9/dxTextureContext9.cxx
```diff
@@ -1935,7 +1935,7 @@ fill_d3d_texture_pixels(DXScreenData &scrn, bool compress_texture) {
                 scaled *= 255;
                 color = D3DCOLOR_RGBA((int)scaled[0], (int)scaled[1], (int)scaled[2], (int)scaled[3]);
                 flags = D3DCLEAR_TARGET;
-                if (device -> Clear (NULL, NULL, flags, color, 0.0f, 0) == D3D_OK) {
+                if (device -> Clear (nullptr, nullptr, flags, color, 0.0f, 0) == D3D_OK) {
```
rdb: First argument is a `DWORD`, so should be `0`, or it does not compile.

https://msdn.microsoft.com/en-us/library/windows/desktop/bb174352.aspx

CFSworks:
This has been rebased onto master as of 5e82671084c9a08d907784fda2e407944fb30730.

rdb:
Thanks!  But it looks like the build is failing since it uses `nullptr_t` without `std::` prefix.

CFSworks:
That's ugly timing.

Issue #335, opened 2018-05-30
Remove `using namespace std;`


rdb:
It's not considered good style to indiscriminately pull in everything from `std`, especially not in a public header that is included by user code.

I would propose the following steps towards phasing this out:
- [x] 1. Replace `using namespace std;` in `dtoolbase.h` with `using std::x;` for each x that we use from `std` without prefix.  This may be a pretty long list at first.
- [x] 2. Start using `std::` prefix for most std types in all `.h` and `.I` files, with few exceptions (eg. swap).
- [ ] 3. As we finish affixing `std::` to our use of standard types in headers, move the relevant `using` declarations for the respective types to the C++ files that use them, or better yet, to the function bodies that would benefit from it.

@cfsworks thoughts?  Good idea or waste of time?

CFSworks:
@rdb Any kind of "best practice" cleanup gets my approval. I've also been pondering the idea of going even further and putting all Panda3D symbols in a `panda3d::` namespace, or perhaps even `panda3d::core::` to mimic the Python namespacing (we could have a `using namespace panda3d;` in our public headers to avoid breaking existing user code).

Regarding step 3: I think we should only have a `using` in a C++ file if the symbol in question is used enough times that it makes sense. (For example, `using std::atomic` makes sense in the context of threading code, but not in the context of a file that only ever creates one `std::atomic`.)

I'd be happy to get on this after I finish my nullptr changes.

rdb:
I've just addressed 1 in e0245d2777f096fd4cb1c1ea852dd7d3ef34c197; there is a list of `using` declarations in `dtoolbase_cc.h`:
https://github.com/panda3d/panda3d/blob/e0245d2777f096fd4cb1c1ea852dd7d3ef34c197/dtool/src/dtoolbase/dtoolbase_cc.h#L164-L189

CFSworks:
Checking second box per 7790f84.

As for the third box: As I mentioned elsewhere I have a local branch that doesn't even use `using`, but it messes up alignment in a few cases and leads to very verbose constructs like `std::cout << std::setw(2) << std::hex << std::setfill('0') << ... << std::dec << std::endl;` so in many cases, I'd say `using` is still appropriate.

[...]

PR #341, opened 2018-06-06
general: Fully qualify header references into the std namespace

CFSworks:
This is phase 2 of #335.

It's a bit heavy to review. I suppose some combination of "look at random files" and "compile on all supported platforms" will give us good confidence nothing went wrong.


rdb:
Thanks a ton!  I'll try building it on Windows and macOS.

CFSworks:
I see the VRML issue; stand by for a fix.

CFSworks:
@rdb Looks like dtoolbase.h pulling in Windows's minmax.h is causing min and max to be defined as macros, which is conflicting with std::min and std::max.

Should we be dropping that include?

rdb:
It does not work on Windows, because the Windows headers define min and max
as macros.

I am told that defining NOMINMAX should fix that, but I can't get that to
work, even when defining it on the compiler command-line.

Alternatives are to omit the prefix for these to, or use them like:
(std::min)(x, y)

rdb:
@CFSworks Ah, you beat me to it.  You're right, I didn't realise there was
an explicit include, that was part of the problem.

We will also need to add something like this to dtoolbase.h:

#ifdef _WIN32
#define NOMINMAX
#endif

...otherwise the Windows headers will still include it.

CFSworks:
Speaking of `_WIN32`: All of the various platform macros are starting to get a little confusing. Is there documentation about various supported platforms (including multiple APIs on one platform, like Cygwin vs. MinGW vs. MSVC) and what One True Macro we're supposed to use to detect that platform (e.g. I see a lot of `defined(WIN32_VC) || defined(WIN64_VC)` but the latter implies the former), or is this worth another issue?

CFSworks:
:+1: From me on Linux and Windows (w/ Clang). I've also done the "look at random files" test quite a bit and can't find anything else out of place.

rdb:
There's still some errors in the DX9 code due to use of min/max with mismatched argument types.  I'll push this as soon as I've fixed those, assuming that's OK.  Thanks!

PR #345, opened 2018-06-08
direct/filter: fixed initialization of FrameBufferProperties class, n…

AnythingTechPro:
…o longer takes a default argument...


rdb:
No, this changes behaviour; the original code was correct.  If the copy constructor of FrameBufferProperties no longer works then that is an issue we should fix.  I will look into why interrogate is not automatically generating a copy constructor for FrameBufferProperties.

rdb:
The FrameBufferProperties copy constructor issue is fixed in fa6d8b4b391ca1c100abb5fef772c81894e23bb6, and a unit test was added in c1e5a71904f043898f71330465f6d95d1d05df3b.

PR #348, opened 2018-06-11
cmake: Fix install procedure for direct

Psychotropos:
# Summary
Currently, the libp3direct shared object and any symlinks associated with it are not included in the install process for the direct library. This MR adds the necessary install directive to direct's CMakeLists in order to rectify this behaviour.


CFSworks:
Thanks!

Issue #250, opened 2018-02-17
BulletVehicle wheels not rendered in sync with chassis


tophatpanda:
I have created a bullet vehicle with four wheels, following the [manual](https://www.panda3d.org/manual/index.php/Bullet_Vehicles) -- specifically the chassis is parented to the vehicle's rigid body node and the wheels use BulletWheel.setNode() to assign a model to them.

The problem is that as the vehicle picks up speed, the wheels move ahead of the chassis:

When stationary:
![carwheelbug-stationary](https://user-images.githubusercontent.com/36481025/36340121-4ee8a1bc-13d6-11e8-8475-2a3b66a12b22.PNG)

When traveling at speed:
![carwheelbug-speeding](https://user-images.githubusercontent.com/36481025/36340122-520fcf78-13d6-11e8-9b93-cb796a8b9122.PNG)

Searching for the cause I didn't find anything on the forums or on GitHub. I did however find [a bullet forum post](https://pybullet.org/Bullet/phpBB3/viewtopic.php?f=9&t=1586&p=5984&hilit=chassis+position+bug#p5984) with a similar problem, which was caused by not properly handling bullet's interpolation.

I then had a look through the [bullet ForkLiftDemo](https://github.com/bulletphysics/bullet3/blob/master/examples/ForkLift/ForkLiftDemo.cpp), and noticed the following lines:

     //synchronize the wheels with the (interpolated) chassis worldtransform
     m_vehicle->updateWheelTransform(i,true);

Which sounds like it could be what's missing. (For reference, the function called is [btRaycastVehicle::updateWheelTransform](http://bulletphysics.org/Bullet/BulletFull/classbtRaycastVehicle.html#adf67ae421334225c886c3dc4c12fe1f2).)

I then started looking through the Panda3d code (mainly [bulletVehicle.cxx](https://github.com/panda3d/panda3d/blob/3eed7bc0427aa533db16fdcf0af402c1aaf11e35/panda/src/bullet/bulletVehicle.cxx) and
[bulletWheel.cxx](https://github.com/panda3d/panda3d/blob/3eed7bc0427aa533db16fdcf0af402c1aaf11e35/panda/src/bullet/bulletWheel.cxx)) and I can't seem to find the function updateWheelTransform() being called anywhere.

[...]

PR #349, opened 2018-06-12
bullet: Interpolate BulletVehicle-wheel transform before syncing it

tophatpanda:
Bullet automatically interpolate most bodies for us (for example the "chassis"
of a BulletVehicle), but BulletVehicle-wheels must be manually interpolated by
calling 'updateWheelTransform()'. Otherwise they will slowly rubber-band
between their intended position and a position one frame ahead in time.

For more information see issue #250.


rdb:
Thanks!

PR #350, opened 2018-06-13
No 'using' in headers

CFSworks:
Man, what a project.

* The first commit adds `using` statements as discussed in #335.
* The second commit gets the `using` statements out of the headers. It also cleans out `using namespace std;` - the only remaining instance is in `py_panda.h`, and that's needed to support generated Interrogate code, so removing that will be a separate task.
* The third and final commit cleans up many of the `using` statements by qualifying the accessed name instead. We might or might not want to squash it down.

Note, we will *definitely* be needing to regenerate the .prebuilt lexers/parsers after this lands. And as always with refactors of this magnitude, we're going to want to do heavy building on all supported platforms/compilers. I've done a bunch, but there's still probably something I missed in here.

Edit: There should also be some investigation about `abs` vs. `std::abs` vs. `fabs`, etc. This PR does not touch `abs` - we might want to use `std::abs` everywhere as it's a template function and may inline better.


rdb:
Thanks a lot!  I'll have a look and also compile on various platforms.

Yes, with `abs` we need to be careful not to accidentally use the integer-only version.  I ran into such a case the other day where GCC helpfully warned about this.  We could use `std::fabs` if we want to be sure that the floating-point version gets used.

I definitely think we should do a squash merge of this PR when testing is done.

CFSworks:
@rdb How do you feel about also removing the `using namespace std;` in `py_panda.h`? (This would involve making Interrogate either be careful to use fully-qualified std names and/or having it insert the `using namespace std;` in its generated output.)

CFSworks:
@rdb For FCollada I would guess adding the `using` would be a slightly cleaner approach. I don't want to risk another preprocessor macro mess if we don't define `FCOLLADA_NOMINMAX`.

I don't yet have my local build environments to the point that I can do much with macOS, so I think I'm going to have to call in the favor and have you look at those.

(Actually, if it'd be convenient for you to take the whole PR off my hands, that would probably provide the most mutual benefit. I'm a little eager to get back to working on CMake after handling all of these using statements.)

rdb:
Yeah, I'll take care of it.  Thanks for all the effort!

rdb:
@CFSworks All is checked in.  Interrogate code now also compiles correctly without `using namespace std;`, although I have not yet removed that line from `py_panda.h`.

PR #358, opened 2018-06-21
add AIRepository.py & edit TimeManagerAI.py

manoguy535:
new airepository. suitable for using distributed object ais & connects to an otp or server repository. should help people with multiplayer networking for now since VR Studio never released theyr repository.

### DIFF
--- a/direct/src/distributed/AIRepository.py
+++ b/direct/src/distributed/AIRepository.py
@@ -1,0 +1,159 @@
+"""AIRepository module: contains the AIRepository class"""
+
+from direct.distributed.ConnectionRepository import ConnectionRepository
+from panda3d.core import *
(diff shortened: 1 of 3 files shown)


thetestgame:
Why are you importing code from the Disney' OTP module in this? That module is not available to Panda3D users.

rdb:
Thanks for the suggested contribution!

Adding a functional AIRepository to Panda3D may be a valuable endeavour, but this attempt in its current state is not adequate or functional, due to important methods not being implemented, and the import to the non-existent otp module.

I'm going to close this PR for now, because of the age of this PR, and because I think an attempt to introduce an AIRepository would require some careful thought and discussion about what it should do and how it should be implemented.  Feel free to open a separate discussion issue for that if you wish.

PR #364, opened 2018-07-08
char: fix CharacterJointEffect dereferencing deleted Character pointer

lachbr:
This fixes a crash that would intermittently occur when deleting an Actor that had CJEs in the hierarchy.

### DIFF
--- a/panda/src/char/characterJointEffect.I
+++ b/panda/src/char/characterJointEffect.I
@@ -35,5 +35,9 @@ get_character() const {
  */
 INLINE bool CharacterJointEffect::
 matches_character(Character *character) const {
+  if (_character.was_deleted()) {
+    return false;
+  }
+
   return _character == character;
 }


rdb:
Where exactly does the crash occur?  It should be perfectly safe to compare a WeakPointerTo to a regular pointer even after it has been deleted, and in fact, it is necessary in this case because otherwise it will not properly clean up the CharacterJointEffect when the Character is removed.

Either way, this is not a thread safe solution since the Character could still be deleted immediately after the `was_deleted` check.

However, it should not be an issue as `operator ==` does not need to dereference the pointer to do the comparison.

I'd like some more information about the crash.  In particular, did you notice the crash after 3d49986ce26652871847594de2592f1452f5536d ?

rdb:
I see that the proper comparison operators were not enabled for the Windows build, for some reason.  I have remedied this in fb82a1c557d02f331fe1ccee46987ffa16232c16, and 409231d214ebfce847d6242cee587ab7f9f72f85 should prevent it from implicitly doing the dangerous cast to a pointer type.

Could you check if the problem has been fixed now?

lachbr:
Yes, thanks!

PR #376, opened 2018-08-13
distributed: Add missing `addTaggedInterest` method

pythonengineer:
This missing method is called in `GridChild`, and if your ClientRepository does not have it already then you'll get an AttributeError when attempting to set grid interest.

### DIFF
--- a/direct/src/distributed/DoInterestManager.py
+++ b/direct/src/distributed/DoInterestManager.py
@@ -192,6 +192,9 @@ class DoInterestManager(DirectObject.DirectObject):
             messenger.send(self._getAddInterestEvent(), [event])
         assert self.printInterestsIfDebug()
         return InterestHandle(handle)
+    
+    def addTaggedInterest(self, parentId, zoneIdList, mainTag, description, otherTags=[], event=None):
+        return self.addInterest(parentId, zoneIdList, description, event)
 
     def addAutoInterest(self, parentId, zoneIdList, description):
         """
--- a/direct/src/distributed/DoInterestManager.py
+++ b/direct/src/distributed/DoInterestManager.py
@@ -109,6 +109,9 @@ class DoInterestManager(DirectObject.DirectObject):
         # keep track of request contexts that have not completed
         self._completeEventCount = ScratchPad(num=0)
         self._allInterestsCompleteCallbacks = []
+        # keep track of tagged interests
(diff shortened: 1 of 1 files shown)


CFSworks:
This seems like it just discards the tags. Wouldn't it be better to make GridChild not rely on `addTaggedInterest` at all?

pythonengineer:
@CFSworks The DoInterestManager doesn't have any special tag functionality, and it doesn't seem to matter if it doesn't. It's possible it was only there to symbolize what the interest is, in which case it may be better to have addTaggedInterest.

May also have to account for removeTaggedInterest, which is used in the same class. I'll look into that in a bit.

CFSworks:
If `DoInterestManager` lacks tag functionality, it doesn't seem sensible to add a function that (in its name) purports to manage tags.

As well, if the function doesn't do anything differently, how does an app developer making use of `DoInterestManager` know whether `addTaggedInterest` or `addInterest` is the appropriate function to call?

If the tags aren't an important feature, it doesn't make sense to avoid removing them. If the tags are an important feature, it doesn't make sense to be discarding them.

Alas I don't know anything about `GridChild` so I have no idea what the intended purpose of the tags is. My gut feeling is that it allows a bunch of interests to be grouped so that the whole group can be removed at once. Nonetheless, the upstream Panda repository doesn't (yet) support interests and it might be better to wait until the Astron fork is merged down and tests/examples are implemented to better understand the impact of these changes.

pythonengineer:
@CFSworks I have a rough idea of how the DoInterestManager would handle tagged interests, but I'm a little busy so I'll update the pull request for that later. As for your concern about Astron, perhaps we should have a separate issue discussing a merge of the Astron fork down here?

I disagree that Panda right now doesn't support interests, it does but only if you have a gameserver that supports the current system used in `direct.distributed`.

CFSworks:
I should clarify that what I meant is that the code being changed has no coverage from either samples or unit tests, and there's no documentation about how GridChild is meant to be used, which makes it extremely hard to verify the correctness of any pull request affecting it. The code is present, but unsupported.

I strongly believe that it'd be a better use of time to get this code tested and documented before trying to modify it, otherwise we're just guessing about the function of dead code, and creating further work down the line for the guesses that turn out to be incorrect.

pythonengineer:
Updated with the tagging system.

Issue #317, opened 2018-05-10
Improving the UI


BMaxV:
Hello everyone,

I recently switched from the blender game engine to panda and I want to contribute code and help to improve it where I can. The most obvious thing is the UI, because there is something I want the UI to do, which it currently can't. That's drag drop functionality to build an RPG like inventory as an example, but I imagine there would be numerous use cases for that.

So I looked at the various bits of code and the development doesn't seem to be that active at the moment, so I'm opening this issue, for people to collect suggestions and requests, things they would like added or improved or maybe even removed or done differently.

I can read&write python and I can read c++, but not very confident in my ability to write good c++, so I'll need a bit of help for that.

To start out with some suggestions or critique:

I think the drag drop functionality I want is already present in the PGSlider and DirectSlider, it's just not exposed enough to be useable outside of that class. So that should be doable without too much trouble by moving some code around.

I also noticed that the Direct GUI elements do some things I would consider a bit strange: mostly the use of imo unneeded globals, the use kw** to handle arguments and offloading the actual construction and initializing of objects to other places than the classes `__init__ `. The DirectGUIBase has an in depth `__init__ `that is then overloaded and called with` super(self).__init__` in the child classes.

I'm sure there are a reasons for these, they're just not obvious to me. There are lots of commets (+1) but unfortunately they don't explain why it's this way, just how it's done.

One of the biggest questions I have that isn't answered by the code, comments or the manual is whether the functionality should be available from python or c++ or both equally.

It'd also be very nice to know if there are some kind of target tests that shouldn't be broken or some common use cases that should be considered.

BMaxV:
>I can post some Python code if you want me to

Please do, sounds like it can be improved either way, at least by adding this to the module.

wezu:
Here's a drag&drop demo:
https://gist.github.com/wezu/17c8021fe2bcbff3e0ef7d3a8460ab90

Moguri:
@BMaxV was @wezu's demo sufficient for your use case? If so, we should add a Manual page on drag and drop with DirectGUI based on the demo.

BMaxV:
The demo is nice, most drag and drop systems have some form of lock or snap though, I'm sure I can do something like that and then we/I can create a manual page for it.

montreal91:
> Here's a drag&drop demo:

Maybe we should do a page or repo (or both) with such code snippets written by users?

BMaxV:
[...]

PR #378, opened 2018-08-18
drag & drop

BMaxV:
This is a drag & drop system built with the directGUI, that I believe should be supported by panda out of the box in some way or another.

I adapted the script provided by wezu in this comment to do it:

https://github.com/panda3d/panda3d/issues/317#issuecomment-388323154

and it's part of the discussion in the same issue:

https://github.com/panda3d/panda3d/issues/317


rdb:
@drewc5131 I agree that DirectGUI is not a great GUI system in many respects, but this is not the right thread to be having that discussion.  For what it's worth, though, [LUI](https://github.com/tobspr/LUI) is intended to be a better alternative.

drewc5131:
@rdb Yeah I kinda just go off on rants without intending to haha, but yeah I have used LUI and it is nice, however theres some annoyances with it, mainly the fact that it doesnt scale like DGUI does with window scale.

xordspar0:
I want to bring attention to this again. Do we _not_ want to merge this because DirectGUI should be replaced and we don't want to invest in it at all? Or should we merge this and then update it when DirectGUI changes?

BMaxV:
It's a sample. While directGUI is in panda and while it is the more documented UI system, I'd say this is useful. I would have liked to have it.

If/When an alternative GUI system is included or merged and directGUI is replaced, I would like there to be feature parity and equally good or better documentation for the new system before directGUI is removed.

xxxshelby:
How to change your files from 2D to 3D?

xxxshelby:
How to change the file to C++?

Moguri:
I am rebasing this branch so CI checks can run. This seems like something useful and it is simple to include (just a sample). However, we still need to review the code to make sure it is in line with our other samples.

I believe the fate of DirectGui is an orthogonal discussion -- especially without a clear successor -- and should not impact this PR.

rdb (changes requested):
I think this needs more work in terms of code style, structure and readability to be useful as a sample.

PR #385, opened 2018-09-06
.setText better

BMaxV:
This is the ideal pull request: short and obvious.

DirectFrame has a setText function, but it only applies something you had to set with instance["text"]=yourtext . Maybe it's prettier that way, I don't know. I do know setText should actually be able to take some text (too) and set it.

So that's what it's going to do now.


kamgha:
Consistency: Why do it for 'text', but not the other properties? Same for the other widgets in whole DirectGUI.

BMaxV:
Hm. I don't know about is not vs. != I do know that python 2 support ends 2020. It's fine if you think panda should support python 2.7 for this time, but I'm not sure if that's worth the effort. In python3 all strings are unicode strings by default.

DirectLabelInstance["text"] does not produce a nice error message when you assign something else than a string anyway:


  File "setattrtest.py", line 6, in <module>
    Label["text"]=1.23
  File "/usr/local/lib/python3.5/dist-packages/direct/gui/DirectGuiBase.py", line 438, in __setitem__
    self.configure(**{key: value})
  File "/usr/local/lib/python3.5/dist-packages/direct/gui/DirectGuiBase.py", line 434, in configure
    func()
  File "/usr/local/lib/python3.5/dist-packages/direct/gui/DirectFrame.py", line 80, in setText
    text = textList[i]
TypeError: 'float' object is not subscriptable


https://gist.github.com/BMaxV/1467e9f7bfeefe33faca61779f845478

So there should be a typecheck there, if you think there should be type checks for types other than str, please tell me which ones.

I don't think not being able to pass None is a problem. We can think about a clearText function, I can add that if you want, but I think if the button or label is displaying nothing, it should be destroyed anyway.

In any other case, where it's just temporarly empty, I think it's acceptable for the user to tinker and test around a bit and use an empty string. That way the type that's contained in Foo["text"] is always a string, I can't see how that would be bad.



@kamgha because I had a problem with setText.

PR #392, opened 2018-09-19
chan: add missing export macro

bluekyu:
This fixes link error by missing export macro in stream functions for `PartBundle::BlendType`.


CFSworks:
Thank you!

PR #399, opened 2018-09-29
Input overhaul updated samples

fireclawthefox:
Updated and tested examples for steering wheel and flight stick devices


rdb:
Thanks a lot!

PR #406, opened 2018-10-05
dcparser: use vector_uchar for blobs, step towards Python 3 compat

smurdock1:
HUH?


rdb:
Why are you creating a duplicate of an existing pull request, #298 ?

Issue #427, opened 2018-10-20
threading-model cull/draw: crash on base.setFrameRateMeter(False)


kamgha:
I'm on Windows 7, Python 3.7:

When the `threading-model` is set to anything but the default empty (e. g. `cull/draw`), a call to `base.setFrameRateMeter(False)` or `base.setSceneGraphAnalyzerMeter(False)` will result in a crash.

Sample code:

```
from panda3d.core import *
from direct.showbase.ShowBase import ShowBase

loadPrcFileData("", "threading-model cull/draw")        # uncomment this


class FPSTest(ShowBase):

    def __init__(self):
        ShowBase.__init__(self)

        self.accept('1', self.show_fps)
        self.accept('2', self.hide_fps)

    def show_fps(self):
        base.setFrameRateMeter(True)

    def hide_fps(self):
        base.setFrameRateMeter(False)


app = FPSTest()
app.run()
```

xseano:
Can you detail what kind of crash you get?

I tested your code sample and got a segmentation fault when trying to disable the frame meter. The segfault could be from a variety of things, but given that cull/draw is a 3-stage pipeline, it could be that there's some error when it tries to access memory space during the first stage.

Alternatively, the default works single-threaded.

xseano:
Can you detail what kind of crash you get?

I tested your code sample and got a segmentation fault when trying to disable the frame meter. The segfault could be from a variety of things, but given that cull/draw is a 3-stage pipeline, it could be that there's some error when it tries to access memory space during the first stage.

Alternatively, the default works single-threaded.

xseano:
Can you detail what kind of crash you get?

I tested your code sample and got a segmentation fault when trying to disable the frame meter. The segfault could be from a variety of things, but given that cull/draw is a 3-stage pipeline, it could be that there's some error when it tries to access memory space during the first stage.

Alternatively, the default works single-threaded.

xseano:
Can you detail what kind of crash you get?

I tested your code sample and got a segmentation fault when trying to disable the frame meter. The segfault could be from a variety of things, but given that cull/draw is a 3-stage pipeline, it could be that there's some error when it tries to access memory space during the first stage.

Alternatively, the default works single-threaded.

xseano:
Can you detail what kind of crash you get?

[...]

PR #430, opened 2018-10-22
Allow SceneGraphAnalyzer and FrameRate to render and de-render with 3-stage pipeline (cull/draw)

xseano:
In progress...


xseano:
Fixes issue https://github.com/panda3d/panda3d/issues/427

rdb:
This is not a bad way to fix this at all, thanks for your attempt.  However, I think the issue at hand merits a deeper fix at the C++ level, which I've addressed in afc994b2fb6ed6258be8529e8e9ecd43d11948f7.

PR #438, opened 2018-11-04
glstuff: send fog parameters to GLSL shaders

lachbr:
How to access:
uniform struct
{
    vec4 color; // Cg attr_fogcolor
    vec4 parameters; // Cg attr_fog
} p3d_Fog;


rdb:
Great, thank you!  Although, I would personally prefer that we break out the parameters into four separate floats, since they are semantically different:
```
uniform struct p3d_FogParameters {
  vec4 color;
  float density; // SMP_row3x1
  float start;   // SMP_cell13
  float end;     // SMP_cell14
  float scale;   // SMP_cell15
} p3d_Fog;
```
Would that work for you?

lachbr:
Yeah, that sounds better. I've pushed that change.

rdb:
Thanks a lot!

Issue #387, opened 2018-09-07
Split Interrogate support code "igateruntime" out of interrogatedb


CFSworks:
Hey @rdb! I know you wanted to explore doing this; I have something of a PoC about how easy it is to do in the CMake branch here: e501d20927ad1b036e939ba1c204c1481424e324

This would also eliminate pystub as interrogate(_module) no longer depend on the Python library themselves. I also like this option because we can (theoretically) compile igateruntime multiple times, once per supported Python version, and we can support many Python versions with a single Panda build.

Thoughts?

rdb:
This is an improvement over the existing approach.  Though I seem to recall that we decided against this particular approach in #244, since it not only requires us to anticipate and make a version for each ABI but it also requires changes to people's build chains if they use interrogate.  I am sitting on changes that move the type tables to dtoolbase, and if we then make changes to make the rest of py_panda subsist of inline files only, that would obviate the need for a libp3igateruntime.

Are there considerations that make you want to go for a separate support library anyway?  I suppose that the size of the support code has been growing with the addition of `py_wrappers.cxx`, though I don't have the numbers to compare.

If we want to go for a support library anyway, I would suggest that we look into refactoring it to use the [Stable ABI](https://www.python.org/dev/peps/pep-0384/) only, which means we would only need to ship two support libraries: one for Python 2.7, one for Python 3.  I am not sure if this is doable it is (and whether there would be a performance cost) given how extensively the py_panda code relies on playing with the type system internals, but it might be better than having to ship ten copies of the same library.

I'll put my type system changes on a branch shortly, so that they're not just sitting in my patch archive.

CFSworks:
> This is an improvement over the existing approach.  Though I seem to recall that we decided against this particular approach in #244, since it not only requires us to anticipate and make a version for each ABI but it also requires changes to people's build chains if they use interrogate.  I am sitting on changes that move the type tables to dtoolbase, and if we then make changes to make the rest of py_panda subsist of inline files only, that would obviate the need for a libp3igateruntime.

That's fine - I pretty much did this just to unblock myself (needing Python##.dll on the path for Interrogate to work in the build process wasn't really feasible for my own Windows build worker). I still think there should be a p3igateruntime, but it could be a headers-only library of static inline functions, rather than a shared object. (That is to say, the runtime code makes no sense in interrogatedb either way. Only apps interested in .in files should link against p3interrogatedb.)

[...]

PR #439, opened 2018-11-05
Remove Python type tables from interrogatedb

rdb:
This is the first step to solving #387, based on the suggestion in [my latest comment](https://github.com/panda3d/panda3d/issues/387#issuecomment-435705406).  It does two things:

- It replaces the dynamic type map with a `PyObject *` stored on TypeRegistryNode.  This is used for resolving a TypeHandle to the corresponding Python type.  It seems justified to rely on dtool for this one since using TypedObject adds a dependency on dtool anyway.  This does not create any dependency on a particular Python ABI, since it is just a forward declaration, so effectively a void pointer.
- It moves the "named type map" to the Python interpreter state, for resolving types defined in the global C++ namespace by their fully scoped name.  No dependency on dtool needed for this one.

Thereby, it clears the way to implement the other things suggested #387, depending on whether we go the baked-preamble route or the cxx-include route.

@cfsworks Thoughts?  I'm anxious to move forward on this so that we can get multi-Python-version builds going well before the 1.10 release cycle.

### DIFF
--- a/dtool/src/dtoolbase/typeHandle.cxx
+++ b/dtool/src/dtoolbase/typeHandle.cxx
@@ -154,6 +154,19 @@ deallocate_array(void *ptr) {
 }
 
 /**
+ * Returns the internal void pointer that is stored for interrogate's benefit.
+ */
+PyObject *TypeHandle::
+get_python_type() const {
+  TypeRegistryNode *rnode = TypeRegistry::ptr()->look_up(*this, nullptr);
+  if (rnode != nullptr) {
+    return rnode->get_python_type();
+  } else {
+    return nullptr;
+  }
+}
+
+/**
  * Return the Index of the BEst fit Classs from a set
  */
 int TypeHandle::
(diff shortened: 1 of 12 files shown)


rdb:
@CFSworks Thanks for reviewing.  Yeah, you got that right.  Nothing earth-shattering, but I figured I'd run it by you since you had objections to my previous approach on this.

I agree with moving `py_*.*` out of interrogatedb; I'm just thinking about the shortest possible path right now in the interest of getting to the end goal more quickly, but it would be a good idea to move this to an `interrogatert` or `interrogatepy` (or something else) directory in a later change.  We should probably then also consider cleaning up these headers, and replacing the `Dtool_` naming scheme with something better.  I wouldn't terribly mind if we lost the history on these files, because it's dark and full of terrors.

The next step is compiling the `py_panda.cxx` stuff as part of the `*_module.cxx` files; awaiting your feedback on #387 for that.

PR #445, opened 2018-11-10
x11display: Cleanup properties reported after requesting new WindowProperties

Moguri:
This includes:
  * Do not set cursor_filename if it was not requested
  * Clear minimized and undecorated from rejected properties
  * Update origin, size, z_order, and foreground if there were set

This helps clean up some tests I am writing, but I do not know how correct the changes are.

### DIFF
--- a/panda/src/x11display/x11GraphicsWindow.cxx
+++ b/panda/src/x11display/x11GraphicsWindow.cxx
@@ -651,6 +651,18 @@ set_properties_now(WindowProperties &properties) {
     properties.clear_fullscreen();
   }
 
+  // Same for minimized.
+  if (properties.has_minimized()) {
+    _properties.set_minimized(properties.get_minimized());
+    properties.clear_minimized();
+  }
+
+  // Same for undecorated.
+  if (properties.has_undecorated()) {
+    _properties.set_undecorated(properties.get_undecorated());
+    properties.clear_undecorated();
+  }
+
   // The size and position of an already-open window are changed via explicit
   // X calls.  These may still get intercepted by the window manager.  Rather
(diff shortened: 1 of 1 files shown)


rdb:
I'm not sure all of this is correct.  Setting the corresponding property in `_properties` implies that the change was successful, whereas most of these property changes are friendly requests to the window manager that we won't know are successful until the X11 event comes in that indicates that the window geometry has changed.  A tiling window manager, for example, may simply reject all positioning or sizing changes.

Moguri:
Is there a reliable way we can have a test wait for the X11 event? I was testing this stuff out on a tiling window manager, which is probably why I had so many issues with size and origin. I eventually told i3 to float the windows I was making to make the testing easier. Any thoughts on improving testing to work better with tiling window managers? Do we care if tests pass when using a tiling window manager?

What are your thoughts on the cursor_filename, minimized, and undecorated changes?

rdb:
The window system is designed not to assume that window requests will be honored.  So it is "by design" that properties may be rejected.  That does complicate testing, though.

We can probably do something like call `GraphicsEngine::open_windows()` to run the event loop, maybe a few times, and assert that the properties are either in the `rejected_properties` list or have become the configured properties.

The `minimized` property should be set upon receiving `UnmapNotify`.  I don't know exactly how we can check whether the `undecorated` change has been processed correctly.

The `cursor_filename` change seems OK to me, as long as we test that setting the cursor and changing it afterwards both still works properly.

Moguri:
I am abandoning this PR for now. I might bring back the `cursor_filename` changes as a separate PR when I have tests in place to catch potential issues.

PR #446, opened 2018-11-10
 tests: Expand test_window to test simple property changes

Moguri:
Not tested:
* cursor_filename
* fixed_size
* icon_filename
* parent_window
* raw_mice


Moguri:
This does not work correctly with tiling window managers. I am abandoning this PR until we determine a better testing strategy for window properties (#572).

PR #447, opened 2018-11-10
WindowProperties: Expose raw_mice property

Moguri:


Moguri:
In my opinion, we should not be adding property interfaces for deprecated methods. Will input-overhaul be merged in for 1.10? If so, I think we can drop this PR.

rdb:
Closed since input-overhaul is merged and the old raw mice interface is now obsolete.

PR #448, opened 2018-11-11
Include cleanup

CFSworks:
Ordinarily I'd just toss this directly on `master`, but I've done some edits to makepanda in the process and it's not extremely clear how makepanda's include discovery machinery works, so it's possible I broke the Assimp build.

A quick check of Assimp on all platforms oughta do it.


rdb:
I'm actually not sure what the recommended form is; I keep reading that `<>` is for system headers and `""` for programmer-defined headers, but not a lot about what the right style is for third-party libraries.

I'm happy to go with this if this is a widely accepted style.

rdb:
I pulled up the C++ standard and it seems to suggest that angle includes are meant to be used for system headers:

CFSworks:
At least on Linux, [the /usr/include directory is defined as the location to place system headers](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s05.html), which means that any third-party package maintainer that installs their headers into `/usr/include` intends for them to be treated as system headers.

_At least on Linux._

We could probably argue:
1. This extends to all Unices: The Linux FHS tries to document the current conventions already in play, and doesn't really attempt to change those conventions. So if this is how it works on Linux, this is likely how it works on Darwin and the BSDs.
2. This indicates the intent of the package maintainer: For any package with a Linux version, we can interpret the fact that they install their headers to `/usr/include` on Linux to indicate that they would prefer that their headers are always treated as system headers. This helps us resolve the ambiguity on e.g. Windows, which doesn't even really _have_ a system include path (e.g. headers in Microsoft-provided SDKs, which are definitely treated as system headers, are still thrown onto the VS project include path like anything else).

The other option is we can consult the code samples for each third-party library to gain some insight into whether each one considers their headers to be system headers. For example, there's [this snippet](https://github.com/openexr/openexr/blob/master/OpenEXR_Viewers/exrdisplay/ImageView.cpp#L45-L48) from the OpenEXR repository is including the OpenEXR headers as system headers. We could probably use this strategy to figure out the type of header for the remaining packages that don't have Linux versions.

Instead of approaching this from a "philosophical" angle, there are some practical reasons for this change which we might consider instead:

1. It makes the code easier to read by creating a clear visual cue for when a header is something we provide (and a curious programmer should look in our repository) vs. when the header is something found separately. That helps prevent situations like [this](https://github.com/panda3d/panda3d/blob/66c5d65bf69fa267270c70c2815cf4f08ed3dc9b/pandatool/src/maxegg/maxEgg.h#L44-L53), where one of the includes doesn't actually belong in that group and it's not visually obvious.
2. [Some compilers silence warnings in system headers](https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html), which is nice because we have really no control over those headers and warnings from those would only hinder our ability to notice warnings we can address and thus keep the Panda code clean. A third-party header is similarly out of our control, so if we draw the line at "headers we control vs. headers we don't" we don't get stray warnings from those headers either.

rdb:
>    1. This extends to all Unices: The Linux FHS tries to document the current conventions already in play, and doesn't really attempt to change those conventions. So if this is how it works on Linux, this is likely how it works on Darwin and the BSDs.

It does not; on FreeBSD, `/usr/include` is for system headers *only*, and `/usr/local/include` is used for all thirdparty libraries, even ones installed via a package manager.  I think this is also the norm for macOS software that doesn't follow its frameworks system.

If there really is no clear precedent and people aren't largely following that note in the standard, I guess this comes down to a style choice.  I guess your argument of clearly visually distinguishing between own headers and external headers makes sense.  One could raise a similar argument about separating standard headers from thirdparty headers, but then one could also argue that the standard library is a kind of thirdparty library.

> Some compilers silence warnings in system headers, which is nice because we have really no control over those headers and warnings from those would only hinder our ability to notice warnings we can address and thus keep the Panda code clean.

I think this refers to something else entirely, namely headers specified with the `-isystem` option.

CFSworks:
> It does not; on FreeBSD, `/usr/include` is for system headers _only_, and `/usr/local/include` is used for all thirdparty libraries, even ones installed via a package manager. I think this is also the norm for macOS software that doesn't follow its frameworks system; the official Python installer also installs to `/usr/local`.

For FreeBSD we may then have to look at which kind of include directive is used for headers in `/usr/local/include`, or perhaps look at the system compiler to see whether `/usr/local/include` is added to the search path for system headers or for regular headers.

> If there really is no clear precedent and people aren't largely following that note in the standard, I guess this comes down to a style choice. I guess your argument of clearly visually distinguishing between own headers and external headers makes sense. One could raise a similar argument about separating standard headers from thirdparty headers, but then one could also argue that the standard library is a kind of thirdparty library.

This was the motivation behind this change. Older Panda code mostly does treat third-party headers as system headers, the newer additions seem not to do so. It'd be great to have consistency either way.

> I think this refers to something else entirely, namely headers specified with the `-isystem` option.

The `-isystem` option is used to add directories to the system search path, thereby declaring that all headers found within are system headers (and the standards-correct way to include them is `#include <header>`). The CMake branch already does this for warning suppression. It is true that `#include <localHeader.h>` does not mean that the header is treated as a system header - the include directive will fall back on searching for a local header when the system header search fails.

My point was that declaring a third-party header to be a system header had practical advantages. In any case we should follow the standard's syntax for including system vs. regular headers.

rdb:
I don't really like trying to rationalise why thirdparty headers are like system headers, but let's just go ahead with this style change anyway, lest this devolves into bikeshedding.  If it comes down to a style choice, we might as well go with what seems to produce the most useful behaviour.  We don't want the compiler to look for these headers in the current directory, so `<>` it is.

(It appears that some libraries take this further and even use something like `#include <panda3d/nodePath.h>` for their own public headers, and only use `#include "cocoaPandaView.h"` for private headers that should not be included publicly.  But we don't currently really have such a distinction in Panda3D, and this would require people to change how they are including Panda3D, so I'm not going that far.)

FWIW, /usr/local/include is not implicitly added to the search path on FreeBSD (in other words, you need to opt-in to use any thirdparty libraries at all).

CFSworks:
> I don't really like trying to rationalise why thirdparty headers are like system headers, but let's just go ahead with this style change anyway, lest this devolves into bikeshedding. 

Ouch - feels a bit harsh to have my reasoning shot down like that, but I guess if it means progress I'll take it.

You've been silent on the Assimp change - does that continue to work on all platforms? Assimp's own CMake configuration defines the include path as the directory containing the `assimp` directory, so this change is necessary before I can get Assimp going.

rdb:
@cfsworks Sorry, it was not intended to be harsh; I found myself wanting to get sucked into a discussion about why one style might be philosophically better than the other but then I realised that this is just a style point and our efforts are probably better spent discussing more important design aspects.  (I'm typically the worst offender when it comes to bikeshedding, so I regularly need to catch myself.)

FWIW, the reason I didn't like that particular path of reasoning was that whether BSD and Linux organise their filesystems to put thirdparty headers together with system headers seems more based on historical convention and doesn't seem like it should be particularly relevant for our style choices.  I would rather weigh precedent set by other libraries more heavily, which it seems there is enough of to justify this change.

The change builds on Linux; the FreeBSD build is still running but I don't expect issues.  To my surprise, we're not currently building with Assimp on macOS, which should be remedied.

I will merge this branch as soon as my tests succeed.

PR #451, opened 2018-11-14
Undeprecate FFT compression

CFSworks:
This reverts commit 2c6a6c9f062dbb7df340133ccdce37fef2c06f2c.

This commit constitutes a commitment to keep the FFT compression feature
around in 1.10 and beyond. The plan at this time is to replace FFTW
support with FFTS (which is comparable in speed and much more liberally
licensed) and have a builtin DFT/FFT implementation to fall back on when
FFTS is not available. In this regard, FFTS does for Fourier transforms
what Eigen does for linear transforms: it provides a more efficient
implementation for Panda to use, but isn't necessary for the functionality
to work.

I should note that I'm not interested in FFT just for the animation compression
code (honestly, I think a better way to store animations would be to make use of
parametric curves), but because having a `FastFourierTransform` class in mathutil
can come in very handy for e.g. audio effects, apps working with multimedia, etc.

I made this a pull request so that we can be sure we're in agreement with this
direction before it lands on master.

### DIFF
--- a/panda/src/chan/animChannelMatrixXfmTable.cxx
+++ b/panda/src/chan/animChannelMatrixXfmTable.cxx
@@ -327,16 +327,10 @@ void AnimChannelMatrixXfmTable::
 write_datagram(BamWriter *manager, Datagram &me) {
   AnimChannelMatrix::write_datagram(manager, me);
 
-  if (compress_channels) {
-    chan_cat.warning()
-      << "FFT compression of animations is deprecated.  For compatibility "
-         "with future versions of Panda3D, set compress-channels to false.\n";
-
-    if (!FFTCompressor::is_compression_available()) {
-      chan_cat.error()
-        << "Compression is not available; writing uncompressed channels.\n";
-      compress_channels = false;
-    }
+  if (compress_channels && !FFTCompressor::is_compression_available()) {
+    chan_cat.error()
+      << "Compression is not available; writing uncompressed channels.\n";
+    compress_channels = false;
   }
 
   me.add_bool(compress_channels);
--- a/panda/src/chan/animChannelScalarTable.cxx
+++ b/panda/src/chan/animChannelScalarTable.cxx
@@ -146,16 +146,10 @@ void AnimChannelScalarTable::
 write_datagram(BamWriter *manager, Datagram &me) {
   AnimChannelScalar::write_datagram(manager, me);
 
-  if (compress_channels) {
-    chan_cat.warning()
-      << "FFT compression of animations is deprecated.  For compatibility "
-         "with future versions of Panda3D, set compress-channels to false.\n";
-
-    if (!FFTCompressor::is_compression_available()) {
-      chan_cat.error()
-        << "Compression is not available; writing uncompressed channels.\n";
-      compress_channels = false;
-    }
+  if (compress_channels && !FFTCompressor::is_compression_available()) {
+    chan_cat.error()
+      << "Compression is not available; writing uncompressed channels.\n";
+    compress_channels = false;
   }
 
   me.add_bool(compress_channels);


Moguri:
I like the idea of having a FastFourierTransform class available. I am also not sure if FFT for animation compression is all that useful, but I would be okay with keeping it around if it's not a maintenance burden.

theclashingfritz:
I like the idea of undeprecating this. It'll allow for older animations compressed with FFTW to be compatible with future versions which is always a +1 in my book.

rdb:
Do we know of anyone using the lossy bam compression feature?  And do we know of any potential uses of FFT in other parts of Panda?  If I wanted to use FFT in my game, I'd just pull in numpy.

Note that we have been building the public builds without FFTW for quite a long time now.  Do we intend to have FFTS before 1.10 is released?  Or would this still just be a feature you can only get by building Panda3D from source?

rdb:
I'm personally mostly indifferent about this.  It's okay if as long as people have actual uses for it, and if we have fairly comprehensive unit tests written.  I also want to be careful not to give off a message that we promise to support FFT-based compression in the (near-)future if we don't in fact have it planned for 1.10—it will still remain disabled in the builds until we have a better-licensed alternative, so its use should still be discouraged until we actually support it in the official builds.

Apropos, since you are struggling with the size of your bam files, you might also want to look at integrating something like [Google draco](https://github.com/google/draco).

> (true parametric curves would be my preferred solution, but barring that, lossy FFT compression is an okay alternative).

I've been thinking for a while that it might be good to have an AnimChannel subclass that just encodes keyframes instead of sampled frames, offering a choice of different interpolation methods.  This would make it easier to build importers for formats like glTF 2.0, and to vary animation speed in Panda without needing `interpolate-frames`.

> (But at any rate, I think the answer will end up being, "More useful than ARToolKit" ;) )

A notable difference with ARToolKit being that ARToolKit is in an easily excludable plug-in, whereas this would be part of the core library that can't be excluded without rebuilding Panda from source.

rdb:
>    1. Try to get the newer FFT implementation in 1.10 (rejecting this PR in favor of a more complete solution)

You're certainly welcome to work on FFTS before 1.10 if you want to, but I wouldn't try to squeeze that in that before 1.10 if it's not a feature people are actively clamouring for.

>    2. Remove the deprecation warning in 1.10, pending a decision about what we'll do after 1.10 (i.e. postpone the deprecation warning)
> 
>    3. Let 1.10 include the deprecation warning, because it's just a warning and we can always change our minds after it's included in a release.

I'm happy to defer to your judgment on whether to eliminate the deprecation warning early or not.  My only concern is that we don't give off the wrong message that we support or maintain it *at the moment*.  But, since it is not a documented feature to begin with, it seems unlikely to cause confusion.

> Are we interested in integrating Draco after 1.10?

It seems like one of those curiosities that I can only imagine becoming a priority if a game in production finds itself in need of it (unless someone feels like adopting it as a pet project).  It's not something I'd be inclined to put on the roadmap otherwise.

> Really the whole concept of a "frame" in an animation should be struck down because it implies discrete steps. A skeletal animation should be thought of as a continuous function of time (e.g. a piecewise Bezier curve or a B-spline) since that allows changing animation speed (as you say) and stops weird interference patterns from showing up due to the frequency difference between the window frame rate and the animation frame rate.
> 
> The animators I work with are very, VERY particular about their animation blending and have a pretty extreme distrust of interpolation systems as a result, so they may not like the keyframe-and-interpolate approach. Storing the curve control points wouldn't be much different than storing keyframes and would put the ultimate control of how the "interpolation" works in the hands of the animators since that's how most animation tools (or, at least, Maya and Blender) present animations for editing.

Yeah, I think we are talking about the same thing; by "choice of different interpolation methods" I included the possibility that it may be interpolated using a B-spline, including user-supplied tangents.  That does indeed essentially make the whole animation channel a piecewise curve with what I called "keyframes" being the control points.  But indeed, the term "frame" becomes kind of meaningless since it is just a unit of time when not dealing with discrete steps.  The ideal design would have an abstract idea of a "curve" that is able to act as an input to an AnimChannel, though it unfortunately seems that the curve implementations in Panda aren't currently truly abstract since they inherit from PandaNode.

My intuition with these kinds of things is usually to look at how transmission formats store these things and set that as a baseline of what to aim to support, since they are ultimately the limiting factor in most art pipelines.  I named glTF 2.0 as an example because it widely supported and supports (among other methods) using cubic spline-based interpolation with supplied tangents.

Moguri:
With 1.10 having been released, does this PR still make sense? There is a lot of good discussion here, but maybe we move it to an issue instead of a PR?

rdb:
Closing this since it is now moot.  However, I will welcome a PR that adds back FFT in a compatible form.  I could easily be convinced otherwise, but I'm leaning towards not removing the warning in the meantime since FFT is not one of our major priorities.

PR #459, opened 2018-11-29
Fix BulletCapsuleShape creation from Panda CollisionTube

PointSource-xx:
Functions get_point_a() and get_point_b() appear to return the top of the cylinder section of the tube, not the tip of the endcap, making subtracting the radius unnecessary.


rdb:
@lachbr Could you confirm this?

lachbr:
Yep, this comparison shows it all (when the radius is subtracted).
[https://gyazo.com/c1df27024ed43989063699e547797976](url)

PR #463, opened 2018-12-01
makepanda: find OpenEXR 2.3 and its static variant (MSVC-case)

kamgha:
This PR adds the ability for makepanda to find and link against OpenEXR 2.3 (built with `NAMESPACE_VERSIONING=ON`) and its static variant (`OPENEXR_BUILD_STATIC=ON`).


rdb:
This changes the behaviour for 2.2 to look for `Half-2_2.lib` where it previously looked for `Half.lib`.  Is that intended?

kamgha:
Indeed, when compiling OpenEXR 2.2 and IlmBase with `NAMESPACE_VERSIONING=ON`, it will produce a `Half.lib`. This has changed for OpenEXR 2.3, where a `Half-2_3.lib ` and/or `Half-2_3_s.lib` will be generated. I've corrected this accordingly, now.

PR #467, opened 2018-12-04
Fix the flickering issue in macOS 10.14

### LINKED ISSUE #414: Window flickers on macOS Mojave 10.14 with included Python 2.7.10
So this is a bit of an interesting issue, but luckily there is at least one solution. After building Panda3D successfully after resolving the issue noted on #412, I encountered an issue where the window would flicker, demonstrated [here](https://www.youtube.com/watch?v=CeJW6-WPdaA). This video demonstrates two scenarios: the first scenario is attempting to run a ShowBase instance using /usr/bin/python2.7 (The included copy of Python 2.7 in macOS, which on Mojave is 2.7.10), and the second scenario is attempting to run a ShowBase instance using python (Python 2.7 installed via Homebrew, which is 2.7.15). You'll notice that running Panda3D under the copy of Python 2.7 installed via Homebrew works as intended, but running Panda3D under the copy of Python 2.7 provided by Apple results in window flickering. This will likely be something worth noting in the README for macOS.

treamology:
It looks like this only happens when building with the 10.14 SDK, which is why we were seeing it so inconsistently. Fixes #414.

### DIFF
--- a/panda/src/cocoadisplay/cocoaPandaView.mm
+++ b/panda/src/cocoadisplay/cocoaPandaView.mm
@@ -26,6 +26,10 @@
   [self setCanDrawConcurrently:YES];
 #endif
 
+  // If a layer ends up becoming attached to the view, tell AppKit we'll manage
+  // the redrawing since we're doing things our own way.
+  self.layerContentsRedrawPolicy = NSViewLayerContentsRedrawNever;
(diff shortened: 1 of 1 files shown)


rdb:
Thanks a ton!

PR #472, opened 2018-12-09
NodePath: Implement replace_material

loblao:


rdb:
This looks like a very useful addition, thanks!

However, this also has the effect of composing all the render states in the graph and writing the composed states.  It would be better to examine every individual state at each level and replace it at that level, without composing.  This would probably be faster too.

I'm happy to fix this when merging, though.

rdb:
I've checked this in with a few minor changes to make it more efficient.  Please let me know if this broke something.

PR #476, opened 2018-12-15
example of shader-based particles

cflavio:
This adds a shader-based example of particles. This approach
allows to define different kinds of particle systems than
direct's ones.

### DIFF
--- a/samples/particles/advanced.py
+++ b/samples/particles/advanced.py
@@ -1,0 +1,265 @@
+#!/usr/bin/env python
+
+# This program shows a shader-based particle system. With this approach, you
+# can define an inertial particle system with a moving emitter whose position
+# can not be pre-determined.
+
+from array import array
+from itertools import chain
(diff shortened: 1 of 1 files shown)


rdb:
Thanks so much!  This is a great addition!

I'll be merging this to master soon.  I've made a few minor alterations, for your information:
* Mesa shader compilers seem to require the GLSL version to be 1.50 to allow layout specifiers at all, probably because they're not a part of the basic grammar of 1.30.  So I updated the shaders to 1.50.
* Added config `gl-version 3 2` since some Mesa-based drivers require this to be able to get GLSL 1.50 shaders.
* Used `setShaderInputs` instead of `setShaderInput` for setting multiple inputs at a time.
* `tex.setRamImage` directly accepts an array, no need for going through strings.
* Used `osg_DeltaFrameTime` for dt, which should be preferred over passing it in via shader input manually.

PR #477, opened 2018-12-15
Fix Static Building for .pyd targets and etc.

theclashingfritz:
Have --static build a .dll or .pyd as .lib no matter what. It's a static build not a mixed built.

I discovered this issue when trying to build statically in a certain way and saw the compiler trying to make a .pyd when those can't be used statically. This was not an issue before a revision to the way Python targets were added but became one after a check no longer passed for them.

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -2199,6 +2199,9 @@ def CompileAnything(target, inputs, opts, progress = None):
         if (origsuffix==".exe"):
             ProgressOutput(progress, "Linking executable", target)
         else:
+            if GetLinkAllStatic(): # If we're building statically, dynamic libraries are un-needed and may cause issues.
+                ProgressOutput(progress, "Linking static library", target)
+                return CompileLib(target, inputs, opts)
             ProgressOutput(progress, "Linking dynamic library", target)
 
         # Add version number to the dynamic library, on unix
--- a/makepanda/makepandacore.py
+++ b/makepanda/makepandacore.py
@@ -3265,8 +3265,12 @@ def GetExtensionSuffix():
 
     target = GetTarget()
     if target == 'windows':
+        if GetLinkAllStatic():
+            return '.lib'
         return '.pyd'
     else:
+        if GetLinkAllStatic():
+            return '.a'
         return '.so'
 
 def GetPythonABI():


rdb:
This one comes up once in a while; let's collect some feedback on how people actually build Panda.  Does everyone who builds Panda statically also want the Python targets to be used statically, or are there people who want to build `.pyd`s with the core libraries linked in statically?  If we want to make both options possible, maybe we need a separate option?

FWIW, I think a better way to implement this might be to change `GetExtensionSuffix()` in `makepandacore.py` to return a `.lib` extension.

theclashingfritz:
^ You're probably right. I'd be a good idea to wait and i agree with your idea on a better change. But both options being available would most likely be the best bet even with the added hassle. But it depends on how most people build Panda3D i suppose.

theclashingfritz:
Another good question to ask is that should statically linked Python files should be in the panda3d folder or if they should be moved to the lib folder with all the other lib files.

loblao:
Just for the record,  I "fixed" this with this change in makepandacore.py:

```diff
 def FindLocation(fn, ipath):
-    if (GetLinkAllStatic() and fn.endswith(".dll")):
+    if (GetLinkAllStatic() and (fn.endswith(".dll") or fn.endswith(".pyd"))):
```

In my opinion, I don't think anyone would want to have static Panda and dynamic Python libs, but maybe it's a good idea to support both.

CFSworks:
For reference, this is what the CMake branch does: 1970751c8affde691a16b8bc41da2040a3ca5e06

When building statically, the libraries are output as e.g. `libpython_panda3d_core.a` to the library path. The user of the library is expected to add lines like `{"panda3d.core", PyInit_core},` to their Python build's  `_PyImport_Inittab`.

This is, however, a fairly complicated and beginner-unfriendly way to use Panda3D. I've been thinking it would actually be better to turn off Python support by default when a static build is performed, and let the user override that if they really do want to use the Python modules in this way. But since the CMake branch is going to land very soon, it might actually be detrimental to add this to makepanda: until the former fully takes over, the latter will likely be preferred by users that want a more stable/simple build at the expense of customization. Such users will not benefit from a build option like this, so I can't see the added complexity being justified.

rdb:
I just tried it and the current `--static` implementation doesn't actually work for any modules other than `panda3d.core` since `panda3d.egg` doesn't have access to the core library symbols defined by `panda3d.core`.  Since I don't know if anyone is using Panda like that, but there do seem to be people using them statically, I'm inclined to accept a change like this one for now, and add a switch to support both methods later if someone complains.

I don't really care much about the naming, but they should probably be in the `lib` dir.  We do use a scheme like `libpy.panda3d.core.so` on Android already (though for different reasons; Android does not support modules, so we have to package them as libraries), so we might as well follow that convention for the static libraries here?  @cfsworks or is there any particular reason to prefer an underscore over a period here?  I thought a period might be less ambiguous for a packaging tool like pfreeze to be able to find.

CFSworks:
@rdb No hard reason other than that underscores are less likely to upset linkers (which typically add a suffix containing a period) than periods.

I think I was also toying with the idea of adding the Python `EXT_SUFFIX` to the library name (when building against multiple Python versions simultaneously), and since that adds a period, I figured a period might be a good separator between "extension module name" and "ABI description" and wanted to keep that possibility open.

But nonetheless it'd be easy to change the `_` in CMake back to a `.` if we decide to go that route.

rdb:
I'm about to check in a change that will build the module as `built/lib/libpy.panda3d.core.lib`, or for Python 3, something like `libpy.panda3d.core.cp37-win_amd64.lib`, depending on the `EXT_SUFFIX`.  We can change tools like pfreeze to be able to pick up these files.  I used a dot since it is less ambiguous (a dot is illegal in a module name), and the `EXT_SUFFIX` component is also not a valid module name so can be stripped off easily.

Issue #474, opened 2018-12-11
Billboarding in older models broken.


theclashingfritz:
As of somewhere around 89236ac1360e5f0249d425e61d537deef40e71c3, Billboarded .bam models made by older versions of Panda3D seem to be broken on viewing via pview or by other means in Panda3D. However if you put them through bam2egg the proper model is restored (mostly with some possible issues because of issues like #464 and #237). As such this seems to be a purely visual issue as they are internally okay for the most part it seems. But this is still a issue to be fixed. 

I have attached a older model with it's newer re-bamifed counterpart for comparison.
[example.zip](https://github.com/panda3d/panda3d/files/2668458/example.zip)

theclashingfritz:
[http://prntscr.com/ltmr97](url) Photo of what this can result in.

rdb:
Both load correctly for me with latest master; I think this might have been fixed in a17ac6516c36a25d28d940cc13bcf9c1480f20f7 already?

theclashingfritz:
Possibly, I'll check it out and let you know.

theclashingfritz:
Nope. Issue is still present. I don't know the root cause of this but it is present on my end still.

rdb:
Fixed in b7d638e53d03046527fe7323bf383c2ddb44d3c8

PR #482, opened 2018-12-18
pgraph: fix BillboardEffect regression with older models

darktohka:
In older BAM files (version <= 6.42), _fixed_depth will have an uncertain value after loading, leading to incorrect transform calculations, which cause rendering issues (see issue #474).

This commit sets _fixed_depth to false when loading older models, which restores the correct legacy behavior.

Fixes issue #474.


darktohka:
Seems like the tests passed just fine!

PR #483, opened 2018-12-19
x11display: Release M_relative mouse when window loses focus

### LINKED ISSUE #480: On X11, mouse does not release from relative mode when switching away from a panda3d app
The fix for this may be similar to what was done for #393.

Moguri:
Fixes #480


Moguri:
I am less confident about this fix versus the confined mouse mode fix. However, the mouse_mode sample appears to work with both relative and confined when I switch workspaces in i3.

rdb:
Thanks a lot!  It works for me, but I also can only test it by switching workspaces since I'm also using i3.

Issue #470, opened 2018-12-06
sync-video 1 with multiple windows causes insane time in Wait:Flip:End


lachbr:
Test case:
```
from panda3d.core import *
loadPrcFileData("", "load-display pandagl")
loadPrcFileData("", "sync-video 1")

PStatClient.connect()

from direct.showbase.ShowBase import ShowBase
base = ShowBase()

base.openWindow()
base.openWindow()
base.openWindow()

base.run()
```

[PStats with sync-video 0, four windows](https://gyazo.com/22665ec001a43f02b3d219bd2bda0778)
[PStats with sync-video 1, four windows](https://gyazo.com/c4c6f21ef0af8a76388bb412a7a6f2b2)

This is on Windows 10 with an integrated Intel graphics card, 60 Hz monitor refresh rate.

rdb:
Well, intuitively, it kind of makes sense: after rendering every window, it waits until the next vblank opportunity before swapping the buffers.  But, it is not very useful behaviour here.

I do note that your code creates four different OpenGL contexts.  What happens if you create the other windows with the same OpenGL context?  This will allow you to share resources as well, so this is probably wise to do either way:

```python
base.openWindow(gsg=base.win.gsg)
base.openWindow(gsg=base.win.gsg)
base.openWindow(gsg=base.win.gsg)
```

If that doesn't help, we could look into the [WGL_NV_swap_group](https://www.khronos.org/registry/OpenGL/extensions/NV/WGL_NV_swap_group.txt) extension, although it is only supported on Quadro and AMD cards.

lachbr:
Passing in the GSG fixes the problem. Thanks, I’ll close the issue.

PR #487, opened 2018-12-21
Use CVDisplayLink on macOS to sync video to display(s)

### LINKED ISSUE #486: sync-video 1 on macOS has no effect on recent development builds
It indeed works in the 1.9.4 release build. I've checked a few builds from several months ago and they all exhibit the same issue. Does anyone know of anything that could have caused this?

treamology:
This PR is intended to fix #486, however it is not ready to be merged. The initial code here is making a few assumptions, but I'm wanting to see if I'm going in the right direction here.

In macOS 10.14, among other issues, `NSOpenGLContext` no longer responds to requests to sync to the display. An alternative to this is `CVDisplayLink`, which is less "set and forget" but offers greater flexibility for applications with multiple windows that are used across monitors with different refresh rates.

I'm not overly familiar with the timing system of Panda, but it appears there's just one global clock used for measuring frametime that is ticked in `GraphicsEngine`. As a result, although the draw rate is properly regulated using this method, the frames per second measurement is inaccurate. I'm interested in hearing suggestions for possible solutions to this.

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsStateGuardian.h
+++ b/panda/src/cocoadisplay/cocoaGraphicsStateGuardian.h
@@ -49,11 +49,11 @@ public:
   FrameBufferProperties _fbprops;
   
   CVDisplayLinkRef _display_link;
-  Mutex swap_lock;
-  ConditionVar swap_condition;
-  std::atomic_bool should_wait;
+  Mutex _swap_lock;
+  ConditionVar _swap_condition;
+  AtomicAdjust::Integer _last_wait_frame;
+  bool _will_vsync;
   bool setup_vsync();
-  bool will_vsync;
 
 protected:
   virtual void query_gl_version();
--- a/panda/src/cocoadisplay/cocoaGraphicsStateGuardian.h
+++ b/panda/src/cocoadisplay/cocoaGraphicsStateGuardian.h
@@ -17,9 +17,11 @@
 #include "pandabase.h"
 #include "cocoaGraphicsPipe.h"
 #include "glgsg.h"
+#include <atomic>
 
 #import <AppKit/NSOpenGL.h>
 #import <OpenGL/OpenGL.h>
+#import <CoreVideo/CoreVideo.h>
 
 /**
  * A tiny specialization on GLGraphicsStateGuardian to add some Cocoa-specific
@@ -45,6 +47,13 @@ public:
   NSOpenGLContext *_share_context;
   NSOpenGLContext *_context;
   FrameBufferProperties _fbprops;
+  
+  CVDisplayLinkRef _display_link;
+  Mutex swap_lock;
+  ConditionVar swap_condition;
+  std::atomic_bool should_wait;
+  bool setup_vsync();
+  bool will_vsync;
 
 protected:
   virtual void query_gl_version();
--- a/panda/src/cocoadisplay/cocoaGraphicsStateGuardian.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsStateGuardian.mm
@@ -36,7 +36,8 @@ TypeHandle CocoaGraphicsStateGuardian::_type_handle;
 CocoaGraphicsStateGuardian::
 CocoaGraphicsStateGuardian(GraphicsEngine *engine, GraphicsPipe *pipe,
                            CocoaGraphicsStateGuardian *share_with) :
-  GLGraphicsStateGuardian(engine, pipe)
+  GLGraphicsStateGuardian(engine, pipe),
+  swap_condition(swap_lock)
 {
   _share_context = nil;
   _context = nil;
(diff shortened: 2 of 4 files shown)


rdb:
Great!  Thanks for working on this.  From what I understand, the CVDisplayLink fires a callback at the beginning of the vblank interval, so the idea is to swap all the buffers ASAP right after the callback?

It's kind of a waste to skip flipping entirely if the swap interval has not yet passed, since this means we've wasted a frame doing something the user isn't going to see.  To implement VSync in a way that matches the other platforms, we need to have `end_flip()` wait for the vertical blanking interval.

Since CVDisplayLink uses a separate thread for the callback, we're gonna need to implement something that's thread-safe.  It is probably best to use a ConditionVar for this, call `wait()` inside `end_flip()`, and `notify()` in the callback to wake up the waiting thread.  You'll also need to use a mutex to protect access to te condition variable.  (It looks like [SDL's implementation](https://hg.libsdl.org/SDL/rev/73f3ca85ac0e) also uses a condition variable, so you can use that as example.)

We will want to think about how to handle the case of multiple windows and multiple contexts well.  The naïve approach of using one condition variable per window and calling `wait()` in `end_flip()` could cause each window to wait for the subsequent swap interval, which means framerate would (possibly intermittently) halve for two windows.  Off the top of my head, we could fix this by having the callback increment an `AtomicAdjust::Integer` and have each window record which blank interval they are waiting for in `end_frame()`.  Downside is that this may still cause tearing in multi-window scenarios as windows may end up flipping "too late" if the second window takes longer to draw than the blanking interval.

A slightly more elegant solution might be to have Panda call `begin_flip()` for every window individually first before calling `end_flip()` for every window, so that each window can record which swap interval it's waiting on in `begin_flip()` and skip calling `wait()` in `end_flip()` if their desired interval has already just passed.

If we do something like either approach, then we might as well move the display link stuff to CocoaGraphicsStateGuardian, unless having a display link per window happens to make implementation easier.

I'm not sure what you mean by your last paragraph.  If CVDisplayLink works as I have understood, then the Panda clock is not relevant here.

treamology:
Okay, I have something basic put together. I don't believe my strategy covers all of the potential edge cases, however, and I'm not particularly experienced with multi-threaded programming, so certainly let me know if something needs to be changed.

There's an undocumented prc variable named `sync-flip` that does what you are describing (batching begin_flips and end_flips) together, but it is off by default, so I'm not sure if any changes need to made because of that.

Also, I was misunderstanding the timing system when I wrote the last paragraph of my original post, so you are correct there.

panda/src/cocoadisplay/cocoaGraphicsStateGuardian.h
```diff
@@ -45,6 +47,13 @@ class CocoaGraphicsStateGuardian : public GLGraphicsStateGuardian {
   NSOpenGLContext *_share_context;
   NSOpenGLContext *_context;
   FrameBufferProperties _fbprops;
+  
+  CVDisplayLinkRef _display_link;
+  Mutex swap_lock;
+  ConditionVar swap_condition;
+  std::atomic_bool should_wait;
```
rdb: You should use an `std::atomic_flag` for this.  The only operations you are using are `clear()` and `test_and_set()`.

panda/src/cocoadisplay/cocoaGraphicsStateGuardian.h
```diff
@@ -45,6 +47,13 @@ class CocoaGraphicsStateGuardian : public GLGraphicsStateGuardian {
   NSOpenGLContext *_share_context;
   NSOpenGLContext *_context;
   FrameBufferProperties _fbprops;
+  
+  CVDisplayLinkRef _display_link;
+  Mutex swap_lock;
+  ConditionVar swap_condition;
```
rdb: Members should have `_` prefix to distinguish from local variables.

panda/src/cocoadisplay/cocoaGraphicsWindow.h
```diff
@@ -107,6 +109,9 @@ class CocoaGraphicsWindow : public GraphicsWindow {
 public:
   // Just so CocoaPandaView can access it.
   NSCursor *_cursor;
+  
+  bool has_frame;
+  void display_ready();
```
rdb: These two are unused now.

panda/src/cocoadisplay/cocoaGraphicsStateGuardian.mm
```diff
@@ -52,12 +53,62 @@
...
+ * Called whenever a display wants a frame. displayLinkContext contains the
+ * applicable CocoaGraphicsWindow.
+ */
+static CVReturn
+DisplayLinkCallback(CVDisplayLinkRef displayLink, const CVTimeStamp* now, const CVTimeStamp* outputTime, CVOptionFlags flagsIn, CVOptionFlags* flagsOut, void* displayLinkContext) {
+  CocoaGraphicsStateGuardian *gsg = (CocoaGraphicsStateGuardian *)displayLinkContext;
+  gsg->swap_lock.lock();
+  gsg->swap_condition.notify();
+  gsg->swap_lock.unlock();
+  gsg->should_wait = false;
```
rdb: It certainly can't hurt to move this assignment inside the lock, I think.

panda/src/cocoadisplay/cocoaGraphicsWindow.mm
```diff
@@ -269,11 +268,20 @@
...
+    
+    if (cocoagsg->will_vsync) {
+      if (cocoagsg->should_wait) {
+        cocoagsg->swap_lock.lock();
+        cocoagsg->swap_condition.wait();
+        cocoagsg->swap_lock.unlock();
+      }
+      cocoagsg->should_wait = true;
+    }
+
```
rdb: It's probably also slightly nicer to use `MutexHolder holder(gsg->_swap_lock);` than a lock/unlock pair in general (include `mutexHolder.h`), though not really a big deal.

treamology:
Thanks for your patience on this. Your suggestion would only work with windows sharing a GSG though, right? If there is more than one GSG, such as each window having its own, each one would be waiting at least once for the condition variable. I feel like it would be best for there to be a structure that contained information for each display, along with its display link, so it would be easier to tell which window needs flipping, regardless of its GSG. I also might be missing something obvious regarding your suggested solution, though.

rdb:
Yes, that's true; multiple GSGs would cause it to wait for the flip multiple times.  It is worth noting that this limitation also exists on other platforms (see #470).

If we wanted to handle this case, we would actually need to store a mapping between display ID to some per-display structure on GraphicsPipe, which would be reference counted so that it would destroy the display link after the last window on the display was closed.  If you want to tackle that, feel free, but I'd be fine storing it on the GSG for now and leaving that open as a possible future improvement.

treamology:
Alright. Also, another unrelated issue popped up when testing: when two windows share the same GSG, each window begins flickering (almost as if each window is trading the rendered image back and forth) and eventually freezes the entire system after several seconds, requiring a reboot. I'll see if I can track this down as well.

rdb:
Thanks a lot!  I made some minor style changes and some changes to make it still work correctly on High Sierra.

I tested on High Sierra and did not encounter issues with sharing contexts between windows.  If it's malfunctioning on Mojave, that might be a different regression.  Unfortunately I don't have a Mojave-capable Mac to test.

PR #491, opened 2018-12-26
Prevent a deadlock when using SIMPLE_THREADS on macOS.

### LINKED ISSUE #490: [macOS] Building with SIMPLE_THREADS=1 results in deadlock
On the latest commit (https://github.com/panda3d/panda3d/commit/862906b86d4a363c9d52afc00a5c6527bb07d5b8), I am able to build Panda3D on macOS with SIMPLE_THREADS=1 without any problems, however, when I try to instantiate ShowBase, it crashes with the following error:

`
:thread(error): Deadlock!  All threads blocked.
`
`
Abort trap: 6
`

Building with SIMPLE_THREADS=1 on macOS worked as of around a week ago, so this seems to be a fairly new issue, though I am not sure which commit introduced it. For what it's worth, this is on macOS Mojave 10.14.2.

treamology:
Panda's simulated threads and real threads are incompatible, so we need to take special consideration since CVDisplayLink notifies from a separate thread. Instead of waiting on a thread that Panda doesn't know about, it will just yield in a loop until a flag is set. Should fix #490.

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsStateGuardian.h
+++ b/panda/src/cocoadisplay/cocoaGraphicsStateGuardian.h
@@ -53,6 +53,9 @@ public:
   Mutex _swap_lock;
   ConditionVar _swap_condition;
   AtomicAdjust::Integer _last_wait_frame = 0;
+#ifdef SIMPLE_THREADS
+  std::atomic_flag _vsync_wait_flag = ATOMIC_FLAG_INIT;
+#endif
 
 protected:
   virtual void query_gl_version();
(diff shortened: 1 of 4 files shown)


rdb:
Not a fan of the spinlock here, which is wasteful given that it will likely be waiting multiple milliseconds much of the time.

Instead we can use TrueMutexImpl, which is a type of mutex that always maps to a real mutex, even with simple threading.  For the condition variable we can also then directly use a ConditionVarPosixImpl—we should probably define a TrueConditionVarImpl in `conditionVarImpl.h` similar to in `mutexImpl.h` while we're at it.

rdb:
Merged, thanks a lot!

PR #503, opened 2019-01-02
setuptools build_apps: Add p3assimp to the check_plugins list

Moguri:
This gets rid of the runtime error message about missing Assimp.


Moguri:
Pushed to incoming.

PR #505, opened 2019-01-03
interrogate: Export Python entry points with EXPORT_CLASS

CFSworks:
~~This is for compatibility with unusual variants of the CPython API, as well as using Python's own recommended linkage type (which might change in the future and/or on certain platforms).~~
**EDIT:** That approach wasn't compatible with `-fvisibility=hidden`, so I've changed it to use `EXPORT_CLASS` instead - we still get the benefit of consistency and now we can control the return type (so we _could_ bring back the Py2<->Py3 error stubs, but I still think they should be removed).

Note that this does remove the "wrong version of Python" notification that's normally present when !NDEBUG, but I suspect that that's actually dead code: the Python modules are linked against a specific versioned Python##.dll or libpython#.#.so, which the dynamic linker will either not find, find but encounter symbol conflicts (on Unix), or use the wrong interpreter's `PyErr_SetString` (may occur on Windows due to symbol imports being explicit about which DLL provides which symbol). In any case, that notification will be of no use to the user.

The presence of a `PyInit_` symbol on Python 2 or `init` symbol on Python 3 might actually create problems for tools attempting to discern the Python version of the module, so it may be for the best to remove it.


rdb:
@cfsworks is this critical for 1.10?

CFSworks:
Good to know regarding `-rdynamic`, though Windows doesn't support that kind of thing as its dynamic linker is far more rigid about symbol resolution, so the problem will still happen there.

Putting the check in `panda3d/__init__.py` is much nicer, actually. That could even allow us to catch instances where the wrong minor version of Python is being used as well.

Although, if we're wondering about "criticality for 1.10" - this was intended as a quick "as-is" tidy-up, which would be great to have in 1.10 but could definitely be merged later.

(Is `master` in a freeze right now? I'm wondering if the 1.10 maintenance branch should be forked so that development can continue on `master` while the 1.10 release candidates are tested.)

PR #509, opened 2019-01-06
leveleditor: LevelEditor fixes

ksmit799:
This PR fixes the inbuilt level editor and allows it to run. It is still quite buggy, but it's the start of potentially supplying a basic level editor with new Panda releases. This PR also fixes deprecation warnings/errors associated with wx, which is used within direct.wxwidgets and ShowBase.


CFSworks:
Thanks a ton! I'm pretty sure `direct.leveleditor` is largely unsupported "as-is" code these days, but PRs are still very appreciated.

Give me a bit to let Travis CI check the branch.

PR #514, opened 2019-01-06
Increased resolution of Panda3D's README logo

thetestgame:
Shortly after the logo change happened the logo started having pixelation problems around its edges as shown here. 

image

This change simply bumps up the size its been read from Github at and adjusting the img tag size to keep the same look. Resulting in this
image


rdb:
Nice!  Thanks!

PR #517, opened 2019-01-08
gui: Add new repeatCommand argument for DirectButton

BelloqZafarian:
Three new arguments have been added to DirectButton: repeatCommand, repeatThreshold, and repeatDelay.

"repeatCommand" is disabled by default. If set to True, holding down a DirectButton will cause it to repeat it's command until the mouse button is released.

"repeatThreshold" is the fastest speed the repeat effect can reach. It has a default of 0.01.

"repeatDelay" is the multiplier that affects how fast the speed of the repeat effect increases. It has a default of 0.5.

### DIFF
--- a/direct/src/gui/DirectButton.py
+++ b/direct/src/gui/DirectButton.py
@@ -25,19 +25,25 @@ class DirectButton(DirectFrame):
         # Responds to click event and calls command if None
         optiondefs = (
             # Define type of DirectGuiWidget
-            ('pgFunc',         PGButton,   None),
-            ('numStates',      4,          None),
-            ('state',          DGG.NORMAL, None),
-            ('relief',         DGG.RAISED, None),
-            ('invertedFrames', (1,),       None),
+            ('pgFunc',          PGButton,   None),
+            ('numStates',       4,          None),
+            ('state',           DGG.NORMAL, None),
+            ('relief',          DGG.RAISED, None),
+            ('invertedFrames',  (1,),       None),
             # Command to be called on button click
-            ('command',        None,       None),
-            ('extraArgs',      [],         None),
+            ('command',         None,       None),
+            ('extraArgs',       [],         None),
             # Which mouse buttons can be used to click the button
-            ('commandButtons', (DGG.LMB,),     self.setCommandButtons),
+            ('commandButtons',  (DGG.LMB,),     self.setCommandButtons),
+            # Whether the button continuously runs the command if pressed
+            # The default repeat threshold is 0.01
+            # The default delay multiplier is 0.5
+            ('repeatCommand',   False,      None),
+            ('repeatThreshold', 0.01,       None),
+            ('repeatDelay',     0.5,        None),
             # Sounds to be used for button events
-            ('rolloverSound', DGG.getDefaultRolloverSound(), self.setRolloverSound),
-            ('clickSound',    DGG.getDefaultClickSound(),    self.setClickSound),
+            ('rolloverSound',   DGG.getDefaultRolloverSound(), self.setRolloverSound),
+            ('clickSound',      DGG.getDefaultClickSound(),    self.setClickSound),
             # Can only be specified at time of widget contruction
             # Do the text/graphics appear to move when the button is clicked
             ('pressEffect',     1,         DGG.INITOPT),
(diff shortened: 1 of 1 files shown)


rdb:
Thanks!  Could you give a use case for this feature?

BelloqZafarian:
Recently I've been using DirectButton in some new GUI I've been creating.  I made a counter that the user can decrease or increase, but after a while it got annoying to constantly click the button to reach the value I desired. I was a bit surprised DirectButton didn't already support pressing and holding for a command to repeat.

This feature does just that. Instead of the developer manually setting up bindings and functions each time they want this functionality, they can simply set repeatCommand to True. Holding the DirectButton will then exponentially continue to repeat it's command until you let go. The repeatThreshold and repeatDelay arguments allow you to customize how fast the command repeats.

By the way, congratulations on the release of 1.10.0!

darktohka:
This sounds like a really useful feature to me.

BelloqZafarian:
To answer your questions:

- The hardcoded 1.0 value was intentional when I first made this commit, however your question made me rethink it. In my latest commit, I've made it an argument with a default of 1.0 named "repeatDelay". This is the initial delay between the execution of the command.
- I've gone ahead and made those functions private. My apologies if my code is not entirely up to par in a case like this. I consider myself a novice developer so I'm still learning the best way to do certain things.
- That makes a lot of sense, I've gone ahead and renamed it to "repeat".
- This resulted in multiple tasks being created, which is not something I intended to happen. I've added a check in my most recent commit to not generate a new task until the current one is removed.

In my latest commit, I've also renamed what was originally "repeatDelay" to "repeatDecay", as I think that term better describes the use for the argument. It is the rate of decay of the **new** "repeatDelay" argument, which as stated previously is the initial delay between execution of the command. I hope this name is satisfactory.

BMaxV:
@rdb I think you should review this again and merge if possible, if there is something wrong I volunteer to take over fixing it until is passes review.

rdb:
The functionality does look useful, though the parameterisation of the exponential increase is hard to understand.

I'm trying to imagine use cases for the exponential increase except for spin boxes—and calling the command more repeatedly does not seem like the best way to solve it for the spin box case (given that you can end up calling the command thousands of times).  I can accept that a periodic repeat (with only a repeatThreshold and repeatPeriod) is generally useful, but perhaps increases over time could be left to be handled by the command itself, and could even be integrated into DirectSpinBox (which is introduced by #742).

Moguri:
@Benjamin8693 or @BMaxV Are either of you still interested in pursuing this PR? If not, or if I don't here from either of you in two weeks, I intend to close this PR. It can always be re-opened.

BMaxV:
Yes, I still think this is useful. #742 is not merged yet either, so I don't think this is redundant yet.

As far as I can see, the comments or "requested changes" were answered or implemented.

Please merge it or provide feedback what needs to change for it to be merged.

Moguri:
@BMaxV has @rdb's concern over the exponential increase and calling the command more repeatedly been addressed?

Benjamin8693:
Sorry for the late reply. My implementation of exponentially calling the command was inspired by a feature in Toontown Online. 

Seen in the video below, the DirectButton is implemented in a way so that holding it down for longer steadily increases how much currency flows from one storage medium to the other. I think this is a very useful feature to have.

KKOmPNUlYm

ArsThaumaturgis:
I can definitely see the utility of such an "exponential increase" feature--but I also think that, for a general "button" widget, such functionality might better be implemented in the method that's called by the button's "repeatCommand" than integrated into the "button" widget itself. This would allow the user to implement it as they see fit.

As mentioned above, in the case of something as specialised as a spin-box widget I think that having such a feature built in might well be fitting, however!

rdb:
I agree with @ArsThaumaturgis.  Exponentially calling a function is a feature that will only be useful in very specific uses of DirectButton, and doesn't seem general enough to warrant putting it into a general-purpose button class.

In general I think it would be good to avoid feature bloat in DirectGUI as much as possible, since it is already a somewhat problematic UI library.

direct/src/gui/DirectButton.py
```diff
@@ -102,6 +127,23 @@ def commandFunc(self, event):
             # Pass any extra args to command
             self['command'](*self['extraArgs'])
 
+    def __repeat(self, state, event):
+        if state and not taskMgr.hasTaskNamed(self.uniqueName('__repeatFunc')):
+            task = self.addTask(self.__repeatFunc, self.uniqueName('__repeatFunc'), extraArgs=[event], appendTask=True)
```
rdb, line 132: Rather than using a task by name, maybe it should just store the task as `self.__task` so that we can do `self.__task.remove()`?

rdb:
Thinking a little more about a more generic way to implement such an exponential decay without adding too specific code to DirectButton…

Perhaps the user could pass in an iterable or a generator to provide the repeat delay, like so?
```python
def exponential():
    i = 1.0
    while True:
        yield i
        i *= 0.9

button = DirectButton(repeatDelay=exponential)
```

Alternatively, instead of a repeatCommand, we could support passing in a generator as command, which could yield the desired wait time:
```python
def command():
    delay = 0.1
    do_a_thing()

    while True:
        yield delay
        do_a_thing()
        delay *= 0.9

button = DirectButton(command=command)
```

Thoughts?

ArsThaumaturgis:
Both of those approaches look good to me.

I do also think that they would also both call for explicit description (with examples as above) in the manual page for DirectButton, however: while both are sensible, neither is really obvious (at least to my mind), and so I suspect that without explicit mention and instruction devs would otherwise be likely to miss that the feature in question is available, and how to use it.

Of the two... I feel that the "repeatDelay" option is a little cleaner than the "yielding command" option: it keeps separate the functionality of the command and the functionality of the command-delay, while the other option intertwines them with each other.

PR #519, opened 2019-01-09
tkwidgets: Fix missing import on rgbPanel.

LittleToonCat:
This PR fixes a crash when trying to start the RGB (color) Panel using the DIRECT tools on the specified node that doesn't have a custom color applied.


rdb:
Thanks.  This import should be at the module level, though, especially since it's also used elsewhere in the module, so I've just done that.

PR #521, opened 2019-01-09
ShaderAttrib: add set_shader_inputs() for C++ users

lachbr:
This is an alternative to the Python extension for C++ users.
Example:
`shattr = DCAST(ShaderAttrib, shattr)->set_shader_inputs({ inp1, inp2, inp3, inp4 });`


rdb:
Great idea, but should this not take an `std::initializer_list` instead?

lachbr:
Thanks. I would say no to ‘std::initializer_list’ because the user may want to build up (or already have) a vector of ‘ShaderInputs’ to pass in.

PR #522, opened 2019-01-10
Merge AstronInternalRepository et al

ksmit799:
This PR merges AstronInternalRepository from Astron/Panda3D and makes it compatible with Python 3 (as well as other misc fixes). This PR doesn't include tests or samples, these will be submitted as a PR shortly.

### DIFF
--- a/direct/src/distributed/AstronDatabaseInterface.py
+++ b/direct/src/distributed/AstronDatabaseInterface.py
@@ -1,0 +1,307 @@
+from pandac.PandaModules import *
+from MsgTypes import *
+from direct.directnotify import DirectNotifyGlobal
+from ConnectionRepository import ConnectionRepository
+from PyDatagram import PyDatagram
+from PyDatagramIterator import PyDatagramIterator
+
+class AstronDatabaseInterface:
+    """
+    This class is part of Panda3D's new MMO networking framework.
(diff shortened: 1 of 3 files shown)


CFSworks:
There are four things I don't like about this PR:

1. It does not preserve the history for the files being merged; there are some commits on those files by authors that should appear in the commit stats, so that they can get due credit for their work.
2. There's a revert in the pull request. Please rebase that away, unless it's important for the file's history in some way.
3. The Python 3 support appears incomplete; it doesn't seem to me like it handles coroutines yet. With most of direct supporting coroutines for event handlers, an incoming message should be supported in this way too.
4. Most importantly, how do I know this code is correct without tests/samples? There's nothing that actually runs it, so this PR is bringing in dead code. Any new files being brought into `direct.distributed` should have a minimum of around 90-95% coverage as reported by coverage.py after running all tests and samples.

I know you said the tests and samples will come later, but why not wait until those are ready? It's much easier to defend the correctness of what you're trying to add when you include a program/script demonstrating correct behavior, and it's just going to be dead code until those are added.

We should probably also use the Astron fork as a staging ground for these changes, since it's easier to collaborate in a fork than in a pull request.

ksmit799:
@CFSworks Sorry, I sort of rushed this pull request through. I'll submit my changes to the Astron fork, and once it's ready it can be merged.

PR #524, opened 2019-01-10
Changes intended to allow "burst emission" of particles.

### LINKED ISSUE #340: Feature request: Particle "burst emission"
There are cases in which it's handy to be able to generate a "burst" of particles, potentially more than one in a given frame. Think of weapons-fire impacts; for multiple in a frame, imagine a hit-scan shotgun.

To my mind, the obvious approach to handling this is to have a "template" particle effect, and to simply have it emit briefly at each of the desired points.

Panda's particle system doesn't currently support this (as far as I'm aware). There is at least one workarounds (using "softstop" and a "doMethodLater" call, I think it was), but it's awkward, and seems unlikely to work for multiple "bursts" per frame.

What I suggest, then, is a new method in ParticleEffect. This would take a position parameter, and for each system in the effect, create a single "litter" of particles placed relative to that point. Ideally, multiple calls in a single frame would create multiple "litters".

What do you think?

ArsThaumaturgis:
These changes should, hopefully, allow the developer to spawn particles in a single "burst"--and multiple times per frame, if called for. This may be useful in cases in which an effect is intended to be brief and potentially-repeated, such as weapon impacts, explosions, fireworks, and the like.

I re-created these changes from my previous commit, so I hope that I didn't miss anything. ^^;

Fixes #340


Moguri:
@rdb any issues with these changes? They seem pretty small and safe to me.

ArsThaumaturgis:
Honestly, I don't recall right now. I just vaguely recall that I had some specific reason for putting it in--it wasn't my first instinct, and I think that I did find it a bit clunky, but saw no better solution to whatever the problem was at the time. :/

I do seem to recall that calling "start" alone didn't work; setting the system up with "softStop" was important to getting burst emission to function, I seem to think.

Ah, here--I found this by digging back to my original feature request: https://github.com/panda3d/panda3d/issues/340#issuecomment-399717867

It looks like I was likely concerned about potentially repeating the reparenting step multiple times per frame (i.e. if multiple bursts are generated per frame). It seems that I felt that it was more lightweight to just set everything up once, then just birth litters when emitting bursts.

ArsThaumaturgis:
Just to check: Is there anything more that you want from my end? Is the above explanation sufficient?

(I do realise that you're likely busy, and may simply have not gotten around to returning to this. I primarily want to check that this isn't quiet again because you're waiting for more from me!)

ArsThaumaturgis:
> Firstly, the method is named for something specific that actually does something very generic ...

That's fair; I'm quite open to renaming it to something like "birth_litter"/"birthLitter".

> ... and on top of that a behaviour which is already implemented by other methods. Is there a good reason why people can't use `start()` and `softStop()` themselves if they need?

Consider the following scenario: you're making a first-person shooter, and you've implemented a shotgun weapon. This weapon uses several scattered ray-casts, done in a single frame, to determine where the pellets have hit. You then want to put a burst of particles at each impact-point. How would you implement those particle-bursts with the current methods?

Perhaps a better approach has been subsequently found, but from what I recall, creating bursts of particles in Panda's particle system has always been rather unwieldy, and ill-suited to the sort of situation that I described above.

(And I may be mistaken, but I seem to recall that the workaround for particle-bursts didn't respond well to being moved multiple times within a single frame. The burst-emission mode in this pull-request does respond well to that, as I recall.)

> What should be the interaction between softStart/softStop and burst emission?
>  I think the easiest behaviour to understand is that softStart() doesn't stop until you call softStop()

That's the thing: burst-emission is more or less the inverse of this behaviour.

What I'd imagine is that softStart/softStop are geared towards a streaming emission from a single source; perhaps there are interruptions, but not on the time-scale of a single frame. A good example might be a smoke signal, interrupted at intervals to create a message.

Burst emission, on the other hand, would be geared towards situations in which the default state of the particle system is that it's _not_ emitting, but in which one might want a sudden burst of particles at some arbitrary moment and location (or set of locations). Again, think of impact-sparks from gunfire.

> Furthermore, the Python method that doesn't match the C++ method even though this is generally the norm for this class

Hmm... The Python method is in a class that doesn't reflect a C++ class, if I'm reading the code correctly. (I may not be--it's been a while, and I am a little tired.) The "ParticleEffect" class inherits from NodePath.

However, that said, I do think that it might be clearer if the "emitBurst" method were renamed to something like "birthLitter", as mentioned above!

> There's also no snake-case alias for the method added to the bottom.

You're quite right!

Hmm... I don't see those in the "ParticleEffect.py" file from this pull request, unlike the version from the main repository, meaning that my version here is somewhat behind. Should I just add a snake-case alias for the methods added here (once we've settled on their naming, and presuming that we do go ahead with this pull request at all), and leave merging to you, or should I perhaps start the pull-request anew...?

> Unit tests would be appreciated.

Indeed! I think that this pull request is from before I was aware of the desire for such tests. If we do decide to go forward with "burst emission" (under whatever name), I can add those in, I daresay.

rdb:
Sorry for not being clear; the first part of my question was about the prepareForBurstEmission method, rather than the emitBurst method (which is fine, but renaming it to birthLitter is maybe better).  Certainly I'm not questioning the existence of the latter, but rather that of the former.

ArsThaumaturgis:
Aaah, I see!

Right, and indeed, looking back at that method I do wonder whether it's the way to go. It does have the benefit of providing a single-method initialisation that readies the particle effect for burst-emission--without such a method, burst-emission might not work as one might be inclined to expect. At the least some degree of explanation would be called for regarding how to get burst-emission working properly.

But on the other hand, if we rename the "emitBurst" method to "birthLitter", then it might not carry the same expectations, and thus it might be more reasonable to expect that devs would think that some basic initialisation is called for ahead of using it.

(By the way, perhaps better still than "birthLitter" might be something like "forceLitterBirth", so that it's clear that this isn't involved in the standard operation of particle emission.)

From here I'm somewhat speculating: it's been a while since I wrote the method in question. That said:

As to using "enable" rather than "softStart", I _think_ that the reason for that may have been to prevent the particle-effect from emitting any particles before actually called upon to emit a burst. But "enable" doesn't allow one to set the parent or render-parent--hence doing so manually in the method.

I'm really not sure of why "softStop" was called for--perhaps so that the birth-rate is set...?

Hmm... Perhaps I should copy these changes to my own machine and mess with them a bit, to see why "softStop" was called for. I'm going to want them if I'm to write unit-tests, in any case!

ArsThaumaturgis:
All right, based on some experimentation:
 - Whether I use "start" or "enable", the particle system begins emitting immediately. Thus "softStop" is required to prevent that.
 - It looks like the initialisation for burst-emission can be reduced to simply the following:
```python
effect.start(render, render)
effect.softStop()
```
In short, you were quite correct in your earlier critique!

(It's possible that I've missed some case in which doing only that would result in unwanted particle-births. I don't _think_ so, based on my experiments and my understanding of the particle system, at least.)

If I'm right, then we can probably just drop the dedicated initialisation method and expect users to just run the two lines of code above.

All that said, I've now renamed "emitBurst" to "forceLitterBirth", and added a snake-case version at the bottom of the class. I've also implemented a short unit-test. All of this should be in the updated pull-request, I believe!

NathanX-S:
Adding a keyword argument to stop might be a better solution then adding another function altogether.

ArsThaumaturgis:
> Adding a keyword argument to stop might be a better solution then adding another function altogether.

Are you referring to initialisation, or to the "forceLitterBirth" method?

Either way, how would a keyword argument to the "stop" method help?

If you're referring to initialisation, I could see the addition of a keyword argument to the "start" method, which would obviate the call to "softStop" (likely by just doing it internally).

NathanX-S:
I'm referring to replacing softStop with just a keyword argument in stop.

ArsThaumaturgis:
> I'm referring to replacing softStop with just a keyword argument in stop.

Ah, I see. Replacing "softStop" is a bit out of the scope of this pull request however, I feel. And it wouldn't change the initialisation code--either way, the user would be calling two methods.

(Just in case it's a point of confusion, let me note for clarity that "softStop" isn't a method of my own; it already exists in the engine.)

NathanX-S:
Ah my bad, didn't think softSpot was in-engine.

ArsThaumaturgis:
> Ah my bad, didn't think softSpot was in-engine.

Fair enough, and not a problem! ^_^

ArsThaumaturgis:
A quick question, if I may: Has a release been decided upon for this to be included into? I see that it's in the "master" branch, but not yet in the "1.10.x" branch, and I don't see another release-related branch to check.

(I took a quick look over the thread, but didn't see any mention of this. However, my apologies if I simply missed it!)

rdb:
@ArsThaumaturgis It is going in 1.11 by default (it's in master, and the 1.11 release branch will be branched off from master).  Do you need it to be in an earlier release?

ArsThaumaturgis:
> It is going in 1.11 by default (it's in master, and the 1.11 release branch will be branched off from master).

Ah, I see! Fair enough, then!



> Do you need it to be in an earlier release?

Hmm... At the time of my previous comment, I would have wanted it so, I think--but the project that if I recall correctly called for it (a little first-person shooter side-project) has been somewhat shelved for the moment.

So no, I think that I'll be fine without as things stand.

Thank you for answering! And indeed, it's good to know that it's set for 1.11. ^_^

rdb:
All right, thanks.  In general, changing access modifiers of a method changes the ABI, which we can't do in a minor release, so we'd have to think of some creative solution (like adding a new method) if we wanted to to get this in 1.10.x.

ArsThaumaturgis:
That's very fair, indeed.

PR #525, opened 2019-01-10
1.10.0 release readme update

montreal91:


montreal91:
@rdb, I can fix it tomorrow.

PR #530, opened 2019-01-11
readme: Add fakeroot into Linux build command

darktohka:
fakeroot is required to build the .deb package for Panda3D, but it is not installed by default on Debian systems. Therefore, it should be part of the dependency install command in the README.


rdb:
Thanks, merged into `release/1.10.x`.

PR #536, opened 2019-01-13
Fix an instance where a macOS window will stick around when trying to have it destroyed.

### LINKED ISSUE #534: OSX Cocoa window doesn't close on destroy()
Minimal example on OSX 10.14.1, Python 3.7.2, Panda3d 1.10.0:

```python
from direct.showbase.ShowBase import *
from time import *

base = ShowBase()
base.destroy()
sleep(5)
```

The expected behaviour - and the behaviour I see on my Linux box - is that a window appears momentarily, and then the process exits five seconds later. On OSX though, the window hangs around until the process exits five seconds later.

Here's a [log file](https://github.com/panda3d/panda3d/files/2751988/logs.txt) generated by this script:

```python

from panda3d.core import *
loadPrcFileData("", "notify-level spam")

from direct.showbase.ShowBase import *
from time import *

base = ShowBase()
print('Destroying...')
base.destroy()
print('Destroyed. Sleeping.')
sleep(5)
```

For context, my original problem is that I'd like to launch multiple, successive `ShowBase` from an interactive console. Having a blank display window hanging around isn't a showstopper, but it is a bit annoying.

treamology:
This commit also removes an extraneous call to [_window setReleasedWhenClosed], which is already set in cocoaPandaWindow's init method. Fixes #534.

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -711,8 +711,11 @@ close_window() {
   }
 
   if (_window != nil) {
-    [_window setReleasedWhenClosed: YES];
     [_window close];
+    
+    // Process events once more so any pending NSEvents are cleared. Not doing
+    // this causes the window to stick around after calling [_window close].
+    process_events();
     _window = nil;
   }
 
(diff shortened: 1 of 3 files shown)


rdb:
Great, thanks!  But does this cause the window delegate to receive the close event, invoke the close request handler, and cause `GraphicsWIndow::close_window()` to be called twice?

treamology:
That does indeed happen, though it was already happening before this change. I can go ahead and fix that as well and just push it to this PR.

rdb:
Before I merge this, can someone confirm that this change does not break when using the multithreaded render pipeline?

Moguri:
Is there someone we can poke to test the multithreaded render pipeline with this change? It would be nice to close out #534.

treamology:
Apologies for the delay on this. It seems to work, and by that I mean nothing explodes and the threads are properly deleted when destroying `ShowBase`.

Issue #506, opened 2019-01-04
EggLoader: Box collision surfaces with non-global axes imported incorrectly


PointSource-xx:
When creating collision geometry from an EggGroup (eggLoader.cxx, method `EggLoader::make_node()`), the code calls `make_collision_solids()`, then transform the solids into local space with
`node->xform(LCAST(PN_stdfloat, egg_group->get_vertex_to_node()));`
In the case of geometry that has non-global axes, the transform should be performed first before calculating the solids.

In my case, Box collision surfaces were created with incorrect dimensions due to the rotation that was applied to them. I suspect this may also be the case for other solids, but haven't tested them.

[simpleBox_collideAttr_30deg.zip](https://github.com/panda3d/panda3d/files/2726044/simpleBox_collideAttr_30deg.zip)
egg file

![boxmodel](https://user-images.githubusercontent.com/3308510/50674409-fa8c3000-1020-11e9-9d89-b932d1c8d6ed.png)
screenshot of file with `show()` called on CollisionNode, and `keep` flag on the Collide group attribute.

[makebox_localcoords.patch.txt](https://github.com/panda3d/panda3d/files/2726045/makebox_localcoords.patch.txt)
Workaround patch for `make_box()`

rdb:
Sorry, I didn't notice there was a patch attached.  Could you file this as a pull request, and fix this to conform to coding style (space between `//` and comment, `snake_case` for variable names)?  Thanks!

PR #549, opened 2019-02-06
EggLoader: correctly import collision surfaces with non-global axes

PointSource-xx:
When creating solids from egg collision group, convert all vertices into local space before calculating the collision solids, otherwise solids with non-global axes and other transforms will be imported incorrectly. Fixes panda3d/panda3d#506

### DIFF
--- a/panda/src/egg2pg/eggLoader.cxx
+++ b/panda/src/egg2pg/eggLoader.cxx
@@ -96,6 +96,7 @@
 #include "uvScrollNode.h"
 #include "textureStagePool.h"
 #include "cmath.h"
+#include "lmatrix.h"
 
 #include <ctype.h>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


rdb:
Looks like your change broke the double-precision build.  You probably should use `LCAST(PN_stdfloat, ...)` instead of `cast_to_float(...)`.

PointSource-xx:
Changes made, checks now passing.

PointSource-xx:
left an errant #include, oops.

rdb:
Upon looking further at this, I'm actually not a big fan of transforming the entire group back and forth, and then leaving a transform on the resulting node.  I've checked in a fix based on your original patch in 4f94b3aa0ed9535d51e4112294c4927f300576ac (using a hack to make it work in 1.10 without breaking the ABI; I'll remove the hack when it lands on master).

PR #553, opened 2019-02-08
fixed an issue that prevented saving with Python 3

### LINKED ISSUE #543: [Particle Panel UI Bug] Cannot Save File
OS: Linux Ubuntu 18.04
Python Version: 3.6

This happens when you try to save a file after entering the file name and clicking on the save button:

> TypeError Exception in Tk callback
>   Function: <bound method ParticlePanel.saveParticleEffectToFile of <direct.tkpanels.ParticlePanel.ParticlePanel object at 0x7fdf810b0668>> (type: <class 'method'>)
>   Args: ()
> Traceback (innermost last):
>   File "/usr/local/lib/python3.6/dist-packages/Pmw/Pmw_2_0_1/lib/PmwBase.py", line 1776, in __call__
>     return self.func(*args)
>   File "/home/ibispi/.local/lib/python3.6/site-packages/direct/tkpanels/ParticlePanel.py", line 1295, in saveParticleEffectToFile
>     self.particleEffect.saveConfig(Filename(particleFilename))
>   File "/home/ibispi/.local/lib/python3.6/site-packages/direct/particles/ParticleEffect.py", line 168, in saveConfig
>     f.write('\n')
> TypeError: a bytes-like object is required, not 'str'

The file is created with the proper name but is completely blank, nothing is saved.

tcdude:
Since in Python 3 `str == bytes` no longer holds true, I modified
the source to adopt byte-like encoding for everything that gets
written to files.
Due to how the components of particle_panel have been designed,
I used something a little "hacky" to enable both writing to sys.stdout 
and to a file with the same method.

Resolves #543

### DIFF
--- a/direct/src/particles/ParticleEffect.py
+++ b/direct/src/particles/ParticleEffect.py
@@ -163,44 +163,41 @@ class ParticleEffect(NodePath):
 
     def saveConfig(self, filename):
         filename = Filename(filename)
-        f = open(filename.toOsSpecific(), 'wb')
-        # Add a blank line
-        f.write('\n')
+        with open(filename.toOsSpecific(), 'w') as f:
+          # Add a blank line
+          f.write('\n')
 
-        # Make sure we start with a clean slate
-        f.write('self.reset()\n')
+          # Make sure we start with a clean slate
+          f.write('self.reset()\n')
 
-        pos = self.getPos()
-        hpr = self.getHpr()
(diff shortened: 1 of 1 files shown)


tcdude:
Ok, modified to using text file for writing instead of the other *hack*.

Tested in Python 2.7 and 3.6 on Ubuntu 18.10

PR #566, opened 2019-02-25
Update download link for SDK to 1.10.1

typewriter1:
The link was to 1.10, an old release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Installing Panda3D
 ==================
 
 The latest Panda3D SDK can be downloaded from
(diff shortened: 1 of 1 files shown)


Moguri:
Maybe we should just point to `https://www.panda3d.org/download/` instead of a specific version that we will forget to update. The download page has a link to download the latest SDK build. This *does* introduce an extra click, which I am okay with.

typewriter1:
Or a `https://www.panda3d.org/download/latest` page maybe.

PR #567, opened 2019-02-26
Pushy Ralph

Moguri:
Some updates to the Roaming Ralph demo:
  * Use a CollisionHandlerPusher
  * Add ability to run backward


rdb:
Thanks!  However, relying on the pusher to handle the terrain movement is causing Ralph to slide down from hills.  Perhaps it would be better to keep using the ray for vertical movement, and set the pusher in horizontal mode (`pusher.horizontal = True`) with its solid hovering a bit off the ground.

We should probably use a sphere somewhere in the middle of Ralph's body, to avoid jittering when bumping into objects; the capsule-into-polygon collision test is very new and still has a few issues calculating the penetration depth properly in edge cases.

rdb:
You're right that it would be ideal to fix the jitter with the capsule-into-poly test; I had added the capsule-poly test three weeks ago but didn't have time to fully work out the interior point issues in the edge cases, reasoning that it would still be more useful to have a test that would work fine for non-pusher uses than not to have it.  I'll see if I can dig up my test code and file an issue to track this.

I'm not quite convinced that a pusher is the right way to handle collision with terrain in general, especially as the pusher doesn't implement friction like a physics engine.  The pusher is mostly intended to handle pushing things out of walls and pushing other objects out of the way.

The backwards movement and sky colour seem like fine changes, thank you.  Though, the sample does need a big graphical overhaul at some point, so I would hold off on replacing textures for now.

Moguri:
I have switched back to using the ray test for vertical movement. However, with a bit more testing, I have noticed that I can get Ralph stuck again.

Moguri:
I'm going to try swapping out the player raycast logic for a CollisionHandlerGravity.

Moguri:
The ground ray was still controlling horizontal movement on a miss. I cleaned this up and Ralph does not appear to get stuck for me anymore.

Moguri:
@rdb Can you please check and see if you are still running into issues? I have yet to reproduce them, but I am hoping that the last commit will help.

rdb:
Yeah, will process this today or tomorrow.  Sorry for the delay; I've been too busy with other things in the past two weeks to stay ahead of issues.

Moguri:
Looks like this didn't get closed when 017b4b58350b98666580ae92b74fbf06cd477890 was pushed; closing now. Thanks!

PR #569, opened 2019-02-28
Add ability to adjust suspension rest length of BulletWheel

TheEnbyperor:
Simply what the title says. The hardcoded default of 0.4 didn't work for my applications so I made this change.


rdb:
Thanks, though what's the reason for the change removing `core.hpp`?

panda/src/bullet/bulletVehicle.h
```diff
@@ -85,7 +85,7 @@ class EXPCL_PANDABULLET BulletVehicle : public TypedReferenceCount {
   void apply_engine_force(PN_stdfloat force, int idx);
 
   // Wheels
-  BulletWheel create_wheel();
+  BulletWheel create_wheel(PN_stdfloat suspension_rest_length);
```
rdb: I think this should have a default value of `0.4` so that it doesn't break existing code.

TheEnbyperor:
I have no idea why that file was deleted, I can only assume that was deleted by the compiler. I'll re-add that now.

panda/src/bullet/bulletVehicle.h
TheEnbyperor: I've added a default in [this commit](https://github.com/panda3d/panda3d/pull/569/commits/7412688aa916a9539418c941cf3d50b0ef94b036)

PR #574, opened 2019-03-03
direct: Resolve DirectScrolledFrame crash on destroy

darktohka:
Destroying DirectScrolledFrames currently crashes the game, because they are destroyed twice. This commit makes sure to destroy the scrolls only once.


JoelStienlet:
Perhaps we could also add a message in the logs when this happens? (on the first call only). If this is not recommended, it may be better not to be completely silent either?

PR #578, opened 2019-03-09
Texture Memory Watcher Python 3 Compatibility fixes

drewc5131:
Fixes the texture memory viewer crashing instantly upon launch in python 3.


rdb:
Rather than casting to int, I think we need to be using `//` in some places where ints are being divided by ints instead of `/`.

drewc5131:
@rdb Where? Nothing I changed involved dividing integers

rdb:
Well, your patch casts some floats to `int` that weren't floats in Python 2.  They are only floats in Python 3 as a result of a float division that should have been a floor division.  Just casting them to int only hides the underlying problem.

No worries, though, I think I've found where it is, and will be checking that in shortly.

PR #583, opened 2019-03-13
add collision tests

hecris:
Added collision tests for the following:
- sphere into sphere
- sphere into box
- box into sphere
- box into poly
- sphere into poly


hecris:
Thanks for taking a look @rdb !
As per your request, I've made the following changes:
- no longer adding "into" object to traverser
- changed`get_into_count()` into `make_collision()`, which returns a CollisionEntry.
In addition, it's pretty easy for other people to write tests between two CollisionSolids. If you're ok with this format, I'll proceed by adding more tests for normal surfaces, etc.

hecris:
Thanks for the feedback @rdb ! I agree with your suggestions and I'll make the requested changes.

PR #590, opened 2019-03-25
Create test_sparsearray.py

Epihaius:
Here are some unit tests for SparseArray.

As I'm still unsure about the expected results of bitshift operations, I can't tell if what SparseArray currently does is correct or not. My tests assume that it is, so I might have to change them.


rdb:
Thanks, these were very useful!

PR #608, opened 2019-04-04
Unit test to check for a valid file extension

DrEuler:
Added Unit test to check for a valid file extension. updated test_loading.py

### DIFF
--- a/tests/audio/test_loading.py
+++ b/tests/audio/test_loading.py
@@ -1,5 +1,11 @@
 import pytest
-
+import os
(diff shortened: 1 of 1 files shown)


Moguri:
As I mentioned on Discord, this test doesn't really check anything since it is not asking anything Panda-related to do anything (e.g., audiomgr is unused). You need to make sure that when `audiomgr.get_sound()` is given a valid path with a valid extension that is successfully loads the sound.

PR #610, opened 2019-04-05
Added new tests for video

DrEuler:
Created new tests to check with valid return names and a valid cursor return value

### DIFF
--- a/tests/video/test_loading_cursor.py
+++ b/tests/video/test_loading_cursor.py
@@ -4,6 +4,6 @@ import pytest
 
 def test_cursor_check(): 
     movie_path = "../../samples/media-player/PandaSneezes.ogv"
-    get_File = MovieVideo.get(movie_path)
-    cursor_type = MovieVideo.open(get_File)
+    get_file = MovieVideo.get(movie_path)
+    cursor_type = MovieVideo.open(get_file)
     assert cursor_type is not None
--- a/tests/video/test_loading_cursor.py
+++ b/tests/video/test_loading_cursor.py
@@ -1,27 +1,9 @@
 from panda3d.core import MovieVideo
-from panda3d.core import MovieVideoCursor
 import pytest
 
-def get_movie():
-    movie_path = "../samples/media-player/PandaSneezes.ogv"
-    getFile = MovieVideo.get(movie_path)
-    return getFile #Obtains a MovieVideo that references a file
 
-@pytest.fixture
-def cursor_check():
-    movie_path = "../samples/media-player/PandaSneezes.ogv"
-    getFile = get_movie() #references the function above
-    valid_cursor_check = MovieVideo.open(getFile) #Open this video, returns a MovieVideoCursor 
-
-
-def test_loading():
-    getFile = get_movie()
-    FileReturn = MovieVideo.getFilename(getFile)
-    assert FileReturn == FileReturn
-
-def test_cursor_check(cursor_check):
-    valid_cursor_check = cursor_check
-    assert valid_cursor_check != None
-
-
-    
+def test_cursor_check(): 
+    movie_path = "../../samples/media-player/PandaSneezes.ogv"
+    get_File = MovieVideo.get(movie_path)
+    cursor_type = MovieVideo.open(get_File)
+    assert cursor_type != None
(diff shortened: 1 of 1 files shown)


tests/video/test_loading_cursor.py
```diff
@@ -0,0 +1,27 @@
+from panda3d.core import MovieVideo
+from panda3d.core import MovieVideoCursor
+import pytest
+
+def get_movie():
+    movie_path = "../samples/media-player/PandaSneezes.ogv"
+    getFile = MovieVideo.get(movie_path)
+    return getFile #Obtains a MovieVideo that references a file
+
+@pytest.fixture
```
Moguri: I doesn't really matter for one file, but fixtures should probably go in `tests/video/conftest.py`

tests/video/test_loading_cursor.py
```diff
@@ -0,0 +1,27 @@
...
+def cursor_check():
+    movie_path = "../samples/media-player/PandaSneezes.ogv"
+    getFile = get_movie() #references the function above
+    valid_cursor_check = MovieVideo.open(getFile) #Open this video, returns a MovieVideoCursor 
+
+
+def test_loading():
+    getFile = get_movie()
+    FileReturn = MovieVideo.getFilename(getFile)
+    assert FileReturn == FileReturn
```
Moguri: A variable compared to itself should always be True, so this assert does nothing.

tests/video/test_loading_cursor.py
```diff
@@ -0,0 +1,27 @@
...
+from panda3d.core import MovieVideoCursor
+import pytest
+
+def get_movie():
+    movie_path = "../samples/media-player/PandaSneezes.ogv"
+    getFile = MovieVideo.get(movie_path)
+    return getFile #Obtains a MovieVideo that references a file
+
+@pytest.fixture
+def cursor_check():
```
Moguri: This should `return` or `yield` a value to be usable as a fixture. As written, I am fairly certain it returns `None`.

tests/video/test_loading_cursor.py
```diff
@@ -0,0 +1,27 @@
...
+    movie_path = "../samples/media-player/PandaSneezes.ogv"
```
Moguri: `movie_path` is unused

tests/video/test_loading_cursor.py
```diff
@@ -0,0 +1,27 @@
+from panda3d.core import MovieVideo
+from panda3d.core import MovieVideoCursor
+import pytest
+
+def get_movie():
```
Moguri: This should be a fixture

tests/video/test_loading_cursor.py
```diff
@@ -0,0 +1,27 @@
...
+    valid_cursor_check = MovieVideo.open(getFile) #Open this video, returns a MovieVideoCursor 
+
+
+def test_loading():
+    getFile = get_movie()
+    FileReturn = MovieVideo.getFilename(getFile)
+    assert FileReturn == FileReturn
+
+def test_cursor_check(cursor_check):
+    valid_cursor_check = cursor_check
```
Moguri: This does nothing

tests/video/test_loading_cursor.py
Moguri: I think you might want to do `assert FileReturn.open() is not None`

tests/video/test_loading_cursor.py
```diff
@@ -0,0 +1,9 @@
+from panda3d.core import MovieVideo
+import pytest
+
+
+def test_cursor_check(): 
+    movie_path = "../../samples/media-player/PandaSneezes.ogv"
+    get_File = MovieVideo.get(movie_path)
```
Moguri: To follow PEP8, this should be `get_file`

tests/video/test_loading_cursor.py
```diff
@@ -0,0 +1,9 @@
...
+    cursor_type = MovieVideo.open(get_File)
+    assert cursor_type != None
```
Moguri: Use `is not None` instead of `!= None` https://stackoverflow.com/questions/2209755/python-operation-vs-is-not

DrEuler:
Changes for this test are also in #612 along with other new tests

PR #611, opened 2019-04-05
Added test to check for valid cursor(audio)

DrEuler:
Checks whether the returned cursor value is valid


DrEuler:
Making new pull request

Issue #603, opened 2019-04-02
Icon window property is not honoured by X11 display


el-dee:
Setting the runtime application icon using WindowProperties.setIconFilename() or usinr PRC property 'icon-filename' does not work on X11 display, the property is simply ignored (it works fine on osx display and win display).

To support this,  x11GraphicsWindow::set_wm_properties() should be modified to :
* get the allowed icon sizes from the WM using XGetIconSizes()
* create a pixmap from the icon bitmap using XCreateBitmapFromData() (and maybe resizing to the closest allowed icon size)
* set icon_pixmap field of wm_hints to the pixmap created above
* set IconPixmapHint in wm_hints flags

The tricky part being read the icon image and convert it to ARGB bitmap :)

rdb:
X11GraphicsWindow already contains a `read_ico` function that reads an .ico file into an XCursor image; maybe we can leverage the same code to read icons?

el-dee:
You're right and it's much easier. On Mac you can use almost any kind of image as icon, but that's the exception not the rule :)

el-dee:
I have it working, I just need to clean up and make a proper PR. In the end I did not use XGetIconSizes() as modern WM does not force a given size for the icons and the returned list is empty. and when using NET_WM_ICON, the image is resized if needed instead of cropped/tilled with wm_hints.

PR #614, opened 2019-04-07
Add support for X11 window icon

el-dee:
This PR adds support of window icon for X11 application. It uses the wm hint NET_WM_ICON to transfer the icon image to the X11 server. Only one image is transferred, the largest one found in the ICO file, the WM is supposed to perform scaling if needed.

Tested OK with "icon-filename" property and using setIconFilename() method of WindowProperties.

Note that it does not work with Ubuntu Unity as it ignores NET_WM_ICON and only uses the icon in the .desktop file. It also uses its own WM Hint (NET_WM_DESKTOP_FILE) to be notified when the .desktop file has been changed.

Finally, read_cursor (formerly read_ico) and read_icon are 95% identical, but as memory allocation for the image is done differently (using XCursor lib for the cursor, vanilla new[] for the icon) and as the output image format is slightly different, I'm not sure that it is worth the effort.

### DIFF
--- a/panda/src/x11display/x11GraphicsPipe.cxx
+++ b/panda/src/x11display/x11GraphicsPipe.cxx
@@ -308,6 +308,7 @@ x11GraphicsPipe(const std::string &display) :
   _net_wm_state_add = XInternAtom(_display, "_NET_WM_STATE_ADD", false);
   _net_wm_state_remove = XInternAtom(_display, "_NET_WM_STATE_REMOVE", false);
   _net_wm_bypass_compositor = XInternAtom(_display, "_NET_WM_BYPASS_COMPOSITOR", false);
+  _net_wm_icon = XInternAtom(_display, "_NET_WM_ICON", false);
 }
 
 /**
--- a/panda/src/x11display/x11GraphicsPipe.h
+++ b/panda/src/x11display/x11GraphicsPipe.h
@@ -136,6 +136,7 @@ public:
   Atom _net_wm_state_add;
   Atom _net_wm_state_remove;
   Atom _net_wm_bypass_compositor;
+  Atom _net_wm_icon;
 
   // Extension functions.
   typedef int (*pfn_XcursorGetDefaultSize)(X11_Display *);
--- a/panda/src/x11display/x11GraphicsWindow.cxx
+++ b/panda/src/x11display/x11GraphicsWindow.cxx
@@ -84,6 +84,291 @@ static int xcursor_seek(XcursorFile *file, long offset, int whence) {
   return str->tellg();
 }
 
+IcoFile::
+IcoFile(Filename filename)
+{
+  VirtualFileSystem *vfs = VirtualFileSystem::get_global_ptr();
+  Filename resolved(filename);
+  if (vfs->resolve_filename(resolved, get_model_path())) {
+    // Open the file through the virtual file system.
+    istream *str = vfs->open_read_file(resolved, true);
+    if (str != nullptr) {
+      load_all_entries(*str);
+    } else {
+      x11display_cat.warning()
+        << "Could not open icon file " << filename << "\n";
+    }
+  } else {
+    // The filename doesn't exist.
+    x11display_cat.warning()
+      << "Could not find icon filename " << filename << "\n";
+  }
+}
+
+IcoFile::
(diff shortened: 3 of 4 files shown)


el-dee:
Slight update to fix compressed ICO and compresed grayscale ICO support (I believe there is the same bug for grayscale compressed cursor, but I haven't tested it)

Moguri:
I might be missing something (e.g., some modern C++ thing), but don't we need to check that `_icon_image` is not `NULL`/`nullptr` before freeing it with `delete[]`? It looks like opening and closing a window that has no icon set will result in `_icon_image` being `nullptr` in the destructor.

el-dee:
The default C++ delete operator handles gracefully nullptr (it's required by the standard), if the operator is overloaded, it is strongly recommended by the standard to handle nullptr too :)

el-dee:
That could be indeed a nice way to avoid code duplication, I will look into that. 
I will also have a look at libxcursor,  maybe it's possible to create a cursor with an already allocated bitmap. So we would have a common function to read the icon / cursor file then create the cursor object or icon accordingly.

el-dee:
Libxcursor does not provide functions to create a cursor from an already allocated bitmap,  so that's a dead end.

Just reading the ICO header does not work with compressed ICO as yo need PNMImage to read the encapsulated PNG image to know the actual size, and anyway the format of the bitmap for the net_wm_hints is not the same as the bitmap used by libxcursor (libxcursor uses 32 bits pixel entries, net_wm_hints uses 32 bits pixels packed into 64 bits (long) entries).

So I think I will make a ICOImage class that will read the ICO file in memory and will provides methods to select which entry of the ICO file to use and return pointer to the uncompressed image. get_cursor() and set_wm_properties() will be responsible to select the entry, copy and format the raw image into the proper format.

el-dee:
I have updated the code accordingly, now there is a helper class, IcoFile, that is responsible to load all the images present in a ICO or CUR file and manage the associated memory.
The specific code has been moved to get_cursor() or set_wm_properties() and the duplicated code removed.

It works fine for the application icon, I still have to make and run some tests for the cursor, but, in the mean time, someone could review the current code and point me what should be modified to be compliant with the Panda3D coding rules.

el-dee:
I did more tests with cursors and it all works fine :
* .ico cursor
* .cur cursor
* .cur cursor with non-zero hot spot
* .ico icon
* .ico compressed icon

Nota: looking a tit more carefully, curXor does not actually leak : it's a pointer inside xorBmp which is correctly released in the cleanup section of the function.

Moguri:
@rdb any further thoughts on this? This PR fixes issue #603.

el-dee:
Are there any issues / comments on the code that still need to be addressed (I don't mind a "please redo everything" :) )

rdb:
As-is this cannot go into 1.10.4 because of the ABI changes, hence the delay--I'm first focused on the non-1.10.4 changes.

From a cursory look, I do think that putting a big class like that at the top of the header file is not right.  In this case you might well move it to a separate file, possibly even outside x11display.  Docstrings are also missing from the class.

el-dee:
Fair points, I will rebase it on the V1.11 branch and move the code to its own files (and indeed add a bit more documentation).
I will then also add support for uploading multiple resolutions icon to X11.

rdb:
Friendly poke — are you still interested in updating this?

el-dee:
Ah yes, this one slipped out of my mind :) I will try and find some time to work on it, I just have to refresh my mind of what is still needed to do.

rdb:
Converted to Draft since it's not yet finished; feel free to mark it back as ready for review when you've finished working on this.

el-dee:
I will probably work on it in a week or two (so you know I didn't entirely forget about it ;) )

duburcqa:
Any news on this ? It would be awesome to have this feature !

duburcqa:
One more year and still in draft. X11 is not dead and it is not going to be the case before at least Ubuntu 24.04, since X11 is still the default on Ubuntu 22.04 for Nvidia GPU.

Moguri:
@el-dee are you still intending to get this PR ready for merge?

Moguri:
Closing due to inactivity. Feel free to reopen the PR if/when you are ready to continue work on it.

PR #622, opened 2019-04-13
Fix Shader::_error_flag not being set for GLSL shader compile errors

Moguri:
This PR also adds some unit tests around compiling GLSL shaders.


rdb:
Sorry, didn't notice that you updated this.  Merged, with some whitespace fixes.

PR #624, opened 2019-04-15
Removes Non Linear addScreen() with one parameter

FReeshabh:
Updating the codebase to remove addScreen()'s version with only one parameter

https://www.panda3d.org/reference/python/classpanda3d_1_1fx_1_1NonlinearImager.html


rdb:
Apropos, I think we need to decide on a proper deprecation policy.  The method was marked deprecated in the API reference, sure (though not using a `@deprecated` tag), but we've not mentioned it in any migration guide or anything of the sort.

I think we need to have a discussion about what the right deprecation policy is, and possibly look into a standard way of showing warnings, which could possibly automatically be generated by Interrogate upon detecting a `@deprecated` tag in the comment.

rdb:
Thank you for this PR, but I think this is not going to be merged at this point.  We may need to first have a discussion about the right deprecation policy and process.

PR #628, opened 2019-04-18
Fix DirectEntry._autoCapitalize() on Python 3.x

jwcotejr:
This pull request aims to resolve the following crash with DirectEntry._autoCapitalize() under Python 3.x (Specifically Python 3.7.2 included with the current release of Panda3D 1.10.2, 64-bit, under Windows 10):

```
  File "_autoCapitalizeTest.py", line 8, in <module>
    base.run()
  File "C:\Panda3D-1.10.2-x64\direct\showbase\ShowBase.py", line 3109, in run
    self.taskMgr.run()
  File "C:\Panda3D-1.10.2-x64\direct\task\Task.py", line 531, in run
    self.step()
  File "C:\Panda3D-1.10.2-x64\direct\task\Task.py", line 485, in step
    self.mgr.poll()
  File "C:\Panda3D-1.10.2-x64\direct\showbase\EventManager.py", line 49, in eventLoopTask
    self.doEvents()
  File "C:\Panda3D-1.10.2-x64\direct\showbase\EventManager.py", line 43, in doEvents
    processFunc(dequeueFunc())
  File "C:\Panda3D-1.10.2-x64\direct\showbase\EventManager.py", line 99, in processEvent
    messenger.send(eventName, paramList)
  File "C:\Panda3D-1.10.2-x64\direct\showbase\Messenger.py", line 333, in send
    self.__dispatch(acceptorDict, event, sentArgs, foundWatch)
  File "C:\Panda3D-1.10.2-x64\direct\showbase\Messenger.py", line 418, in __dispatch
    result = method (*(extraArgs + sentArgs))
  File "C:\Panda3D-1.10.2-x64\direct\gui\DirectEntry.py", line 213, in _handleTyping
    self._autoCapitalize()
  File "C:\Panda3D-1.10.2-x64\direct\gui\DirectEntry.py", line 218, in _autoCapitalize
    name = self.get().decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```

Two simple changes in the `_autoCapitalize` function were all that was needed to resolve this crash under Python 3.x, while also preserving compatibility with Python 2.7.

A simple unit test was also included to test this, and a simple test program can be found here: https://pastebin.com/P5sjhYQt


rdb:
I presume the original call was there for a reason; I guess if the case that there was a `str` object in Python 2 that had multi-byte UTF-8 sequences in it.  Maybe you should include a UTF-8 string (without `u` prefix) in your test.

jwcotejr:
As in, use UTF-8 characters in a string, or just use any string that just so happens to be UTF-8 encoded?

rdb:
As in, use a string that has non-ASCII characters, in UTF-8 encoding.  (If a string doesn't have non-ASCII characters, its UTF-8 encoded form is identical to the ASCII form, because of how UTF-8 is designed.)

jwcotejr:
@rdb Hi there, I have made the necessary changes you have requested. Sorry about the extended delay.

I am not entirely sure if this is 100% ideal, as I am setting the UTF-8 encoding within the unit test itself, nevertheless it does function as intended on both Python 2.x & Python 3.x.

Take care!

ItsEris:
Here I go again! Hopefully reviving @rdb on this one. A definite good push towards full Python 3 support. 👍

rdb:
Yeah, let's get this into the next release.  I'd only additionally like to see tests with Unicode objects in Python 2, but I can make that change myself if needed.

Issue #634, opened 2019-04-27
LinuxInputDeviceManager segfaults when no `/dev/input/event*`


CFSworks:
[This line](https://github.com/panda3d/panda3d/blob/3da32b7b7cd00ba5dfe9abbe251848665d105c61/panda/src/device/linuxInputDeviceManager.cxx#L63) calls `indices.back()` but it's possible for `indices` to be empty if there are no `event*` character devices in `/dev/input`; C++ defines this as undefined behavior, and on Linux I'm getting a segfault.

Fix is easy, but I'm in a rush so reporting as a bug instead: Check that `indices` isn't empty before the whole sort/resize/`consider_add_evdev_device` block.

PR #635, opened 2019-04-27
Linux input Checks if it's zero

DrEuler:
Addresses Issue #634


panda/src/device/linuxInputDeviceManager.cxx
```diff
@@ -62,6 +62,9 @@ LinuxInputDeviceManager() {
     // We'll want to sort the devices by index, since the order may be
     // meaningful (eg. for the Xbox wireless receiver).
     std::sort(indices.begin(), indices.end());
+    if(indices.empty()){
```
CFSworks: I was thinking this check would go earlier - between lines 60 and 62. std::sort on an empty vector probably isn't expensive but we may as well skip it if it's easy, right?

Also, the if should be spaced out: `if (X) {`

panda/src/device/linuxInputDeviceManager.cxx
```diff
@@ -61,6 +61,9 @@ LinuxInputDeviceManager() {
 
     // We'll want to sort the devices by index, since the order may be
     // meaningful (eg. for the Xbox wireless receiver).
+    if (indices.empty()) { //Check if indices are empty
```
rdb: This comment is redundant, because the code speaks for itself.  Comments should never be used to explain *what* the code does, only *why* the code does it.

Issue #636, opened 2019-04-27
DirectOptionMenu resets popupMarker's pos


ArsThaumaturgis:
DirectOptionMenu allows one to pass the keyword-parameter "popupMarker_pos" into its constructor, or to call "myMenu.popupMarker.setPos( <some position> )" thereafter.

However, the former appears to have no effect, and while the latter works, it's overridden as soon as the menu's list is changed.

Looking at the code, the problem seems to lie in the "setItems" method, which resets the popupMarker's position--which makes sense when the popupMarker hasn't been given an explicit position.

What I suggest is that the "setItems" method handle the popupMarker's position in pretty much the same way as it does the menu's "frameSize" (just above the point at which the marker's position is set): if the user has provided an explicit position for the marker, that position is used instead of calculating one.

It wouldn't solve the case in which "setPos" is used, but it should at least improve the behaviour of the menu, I think.

PR #637, opened 2019-04-28
DirectOptionMenu: Fix popup menu reset of the popup marker's position…

xMakerx:
… in response to ArsThaumaturgis's issue https://github.com/panda3d/panda3d/issues/636.

Also defines `popupMarker_pos` and raises an assertion error if #showPopupMenu() is called when no items have been specified.

### DIFF
--- a/direct/src/gui/DirectOptionMenu.py
+++ b/direct/src/gui/DirectOptionMenu.py
@@ -22,10 +22,12 @@ class DirectOptionMenu(DirectButton):
             # List of items to display on the popup menu
             ('items',       [],             self.setItems),
             # Initial item to display on menu button
-            # Can be an interger index or the same string as the button
+            # Can be an integer index or the same string as the button
             ('initialitem', None,           DGG.INITOPT),
             # Amount of padding to place around popup button indicator
             ('popupMarkerBorder', (.1, .1), None),
+            # The initial position of the popup marker
+            ('popupMarker_pos', (0, 0, 0), None),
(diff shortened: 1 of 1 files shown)


rdb:
It looks like this change broke this simple piece of code:
```python
from direct.gui.DirectGui import DirectOptionMenu
app = ShowBase()
menu = DirectOptionMenu(items=["item1","item2","item3"])
app.run()
```
```
  File "<stdin>", line 1, in <module>
  File "/home/rdb/panda3d/built/direct/gui/DirectOptionMenu.py", line 48, in __init__
    self.initPopupMarkerPos = self['popupMarker_pos']
  File "/home/rdb/panda3d/built/direct/gui/DirectGuiBase.py", line 477, in cget
    '" for ' + self.__class__.__name__)
KeyError: 'Unknown option "popupMarker_pos" for DirectOptionMenu'
```

fireclawthefox:
The problem lies on the name of the option. Due to it having an underscore in it's name the DirectGuiBase addoption method stores it in keywords not optionInfo. So it will only be available after the popup marker component had been set up.
Simply moving the line
    self.initPopupMarkerPos = self['popupmarker_pos']
below the
    self.popupMarker = self.createcomponent(...
will fix this. Otherwise the options name has to be changed.


On a sidenote, I'd recommend to change the default value for the initial pos from (0, 0, 0) to None to keep the old default position if the user doesn't pass it to the constructor.

PR #652, opened 2019-05-11
DirectOptionMenu "Empty-List" Label

ArsThaumaturgis:
Changes to DirectOptionMenu.py intended to provide a label that is shown by the menu when no items are present, and the logic to show it thus.


rdb:
> The diff tool seems to think that I deleted these lines, but most have simply been indented.

Under "Diff settings" you can tell it to ignore whitespace changes in the diff (similar to using the `-w` argument to `git diff`).

ArsThaumaturgis:
Ah, thank you for that. ^_^ (I think that my main concern is that it is apparent that that the changes are just indentation; I don't want it to seem that they're more extensive than they really are! ^^; )

BMaxV:
This also seems like it can be merged? Any counter arguments?

Moguri:

fireclawthefox:
I think the label could be made a bit more optional. For example if the developer don't set any text on it, it could be ignored as it'd also overwrites the text attribute of the menu, which may not always be intended if it's empty.
Alternatively and also in addition some more options could be added. One for if the label should be used and one for the labels text, which then could also be filled with a default value.

ArsThaumaturgis:
@fireclawthefox: Those seem like good ideas; I'll try to get around to setting aside some time to look into them soon, then. ^_^

rdb:
Marked as draft since from the comments I'm gleaning there's still more work to be done? Feel free to request review otherwise.

ArsThaumaturgis:
Hi, sorry, yes--I've been meaning to get to these changes but just haven't thus far! ^^;

On the plus side, this update does push it up a little in my mind, I daresay!

ArsThaumaturgis:
Okay, in looking at this again today I've encountered a wrinkle:

It's all very well and good to have a label for empty lists--but attempting to open the popup-menu for such a widget asserts out, warning that no popup an be shown for an empty list.

So, what should happen here? Should an empty list be considered an error-condition? Or should we consider it reasonable to open an empty list?

There are ways around this, of course: for example, the button can be disabled in developer-code when an empty list is found.

Conversely, allowing the user to open an empty list would allow them to _see_ that the list is empty.

Thoughts?

rdb:
I think in these cases we should follow established practices, and look at how these kinds of widgets behave in other toolkits (Windows, macOS, etc.)

ArsThaumaturgis:
A good thought.

Looking quickly at HTML--specifically [via W3Schools](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select)--I see that in that language a drop-down with no entries is in fact allowed, with no (immediately apparent) error.

Specifically, the result is a small, blank button, which produces a small, empty list. (Which can of course be made a larger, blank button via a "style" tag--much as DirectGUI allows explicit sizing via the "frameSize" keyword.)

Issue #651, opened 2019-05-09
Feature suggestion (DirectGUI): DirectOptionMenus with Static Text


ArsThaumaturgis:
As it stands, DirectOptionMenu seems to be designed essentially as a means of selecting a persistent state: you select an option in the list, and that remains the selected option.

However, there may be cases in which a similar UI is desired, but in which the program's state doesn't necessarily remain that of the selected option.

For example, imagine a game in which the player can select a spell to cast in battle--the spell is cast, but (with exceptions) it's a transient effect. 

An example that I have from my own work is that of key-binding profiles: I have a menu from which these can be selected. However, while selecting one resets the bindings to those specified in the profile, the player is free to alter them thereafter, with the bindings being saved automatically to a separate file. Thus the most-recently selected profile may not reflect the actual bindings.

(In that case I worked around the issue via a custom sub-class of DirectOptionMenu that manually resets the button-text in the "set" method.)

If I'm not much mistaken, it is possible to set the button-text of a DirectOptionMenu, and even force it to keep that--until an item is selected. As soon as that happens, any prior settings seem to be overridden.

What I suggest then is that the current behaviour remain the default, but that an optional keyword-parameter be added that causes DirectOptionMenu to skip over the logic that sets the menu's button-text.

ArsThaumaturgis:
If this hasn't already been addressed, I might take this on, along with #650 and #646.

ArsThaumaturgis:
This is done, I think, with a pull request here: https://github.com/panda3d/panda3d/pull/653

Here's a short test-program that demonstrates the new functionality; press "space" to cycle through the available item-lists, noting that the button (hopefully) remains unchanged. The first item-list also has some additional styling, to test menu-placement.

```
from direct.gui.DirectGui import DirectOptionMenu

from direct.showbase import ShowBase as showBase


class Game(showBase.ShowBase):

    def __init__(self):
        showBase.ShowBase.__init__(self)

        self.accept("escape", self.userExit)

        self.accept("space", self.changeItemList)

        self.itemListIndex = 0
        self.itemLists = [
                            ["Kitten", "Cat", "Kitty"],
                            ["Mouser"],
                            ["Purr", "Meow"]
                         ]

        self.menu = DirectOptionMenu(
            scale = 0.1,
            pos = (-1, 0, 0),
            items = self.itemLists[0],
            item_pad = (0.7, 0.7),
            item_frameColor = (0, 1, 0, 1),
[...]

PR #653, opened 2019-05-13
Direct option menu static text

ArsThaumaturgis:
Changes intended to allow DirectOptionMenu to optionally display a given piece of text on its button, rather than changing its text to reflect the most recently-selected item.


ArsThaumaturgis:
1) I _think_ that I've done so--but I'm not confident in that. Please check and confirm for me.

2) Done, if I understand the means of doing so correctly.

3) It doesn't fix a bug, as such; instead it provides additional functionality. Specifically, it allows a DirectOptionMenu to have a given piece of text on its button, regardless of item-selection.

Here's the original discussion, I believe: https://github.com/panda3d/panda3d/issues/651
(Said discussion also includes a test-program demonstrating the new feature.)
(There's also [this forum thread](https://discourse.panda3d.org/t/set-fixed-text-in-directoptionmenu/25317), which includes someone else looking for similar functionality.)

4) Done, I believe; I've added the test-method to the extant "DirectOptionMenu" test-script.

PR #655, opened 2019-05-14
Direct option menu keep item attribs

ArsThaumaturgis:
Changes intended to allow a DirectOptionMenu to retain item-styling (from "item_" keyword-parameters) when the item-list is changed.

The reconstruction of the item-list in "__setitem__" is perhaps not the most efficient approach to this, in particular when multiple "item_" keyword-values are changed one after another. It might perhaps be more so to create a separate method to handle this, similar to the "resetFrameSize" method--but I hesitate to make such a change unilaterally.


rdb:
Could you report an issue with a test case to demonstrate the bug this is intended to solve?  The solution looks a bit hacky to me, but it's hard for me to evaluate it without knowing exactly what it tries to solve.

ArsThaumaturgis:
Sure!

In fact, I thought that I had posted such a program somewhere--but I don't seem to be finding it now. Ah well, it's simple enough to produce one. And so, here it is:

```python

from panda3d.core import Vec3

from direct.gui.DirectGui import DirectOptionMenu

class Game(ShowBase):
    def __init__(self):
        ShowBase.__init__(self)

        self.accept("space", self.changeItems)

        self.menu = DirectOptionMenu(scale = 0.1,
                                     items = ["mew", "meow"],
                                     item_frameColor = (0, 1, 0, 1),
                                     item_pad = (0.5, 0.5))

    def changeItems(self):
        self.menu["items"] = ["cat", "kitten"]

app = Game()
app.run()
```

To test: Open the menu, and note that the items are coloured green, and have significant padding. Now, press the space-bar to change the list of items in the menu. This causes the menu to be re-built--but since the "item_" keywords are no longer available, the applied styling is lost; the items are once again default-grey, and unpadded.

One can somewhat work around this by resetting those keywords when changing the item-list. However, a quick test seems to indicate that doing so doesn't entirely take, as clicking on an item then once again resets the style and the "item_pad" keyword doesn't seem to have an effect at all. 

Furthermore, when a menu has been given sufficiently complex styling, resetting everything whenever the item-list is changed seems like a bit of a pain--not to mention potentially-unexpected behaviour.

Moguri:

ArsThaumaturgis:
> @ArsThaumaturgis is this PR still something you'd like to pursue?

I think that that would depend on what feedback is given on it. As noted in my first post, the approach offered here is perhaps not ideal--but conversely, I do think that it's a potentially-useful feature to have.

fireclawthefox:
Since I came across this problem a few times myself and always had to create a workaround, I would really appreciate this feature. Regarding the implementation, I wouldn't know of a better way yet.

PR #656, opened 2019-05-14
DirectOptionMenu: Cleanup old items on creating the item-list

ArsThaumaturgis:
A change intended to clean up any previously-created items when creating the item-list.

This approach may not be the best, but i submit it in case it's useful!


Moguri:

PR #662, opened 2019-05-25
DirectOptionMenu bugfixes noted in: #637

fireclawthefox:
Fixes for #637
- Fix popupMarker_pos related crash
- Set initial popupMarker position to previous default by using None as default

### DIFF
--- a/direct/src/gui/DirectOptionMenu.py
+++ b/direct/src/gui/DirectOptionMenu.py
@@ -27,7 +27,7 @@ class DirectOptionMenu(DirectButton):
             # Amount of padding to place around popup button indicator
             ('popupMarkerBorder', (.1, .1), None),
             # The initial position of the popup marker
-            ('popupMarker_pos', (0, 0, 0), None),
+            ('popupMarker_pos', None, None),
             # Background color to use to highlight popup menu items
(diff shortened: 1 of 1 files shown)


rdb:
Have you tested that this produces the correct (pre-1.10.3) behaviour both with and without the option set?

fireclawthefox:
I haven't changed any of the placing behaviour and my tests showed me, that setting the popupMarker_pos to None will place it at the correct default position to the right, centered vertically. Which was AFAIK always the default. Setting it to (0,0,0) will put the marker to the lower left edge of the box as expected.
If that's what you're asking for.

rdb:
Closed in 387b43adde75184ddb9801ee05188e0cc09da994

tyteen4a03:
What's the workaround for this issue? Manually passing in a `popupMarker_pos` still causes p3d to crash the same way.

ArsThaumaturgis:
Offhand, things that occur to me that should work: 
 - You can edit the code for DirectOptionMenu; a look at the patch that fixes the issue on GitHub should give you an idea of what to do. (It's a fairly simple fix, I seem to think.)
 - You can use a pre-release version of Panda
 - You can wait for the release of 1.10.4, which I think will have the fix.

PR #664, opened 2019-05-30
CollisionParabola into CollisionBox

hecris:
Added collision test for parabola into box as well as unit tests. The algorithm iterates through the 6 faces of the box, testing their intersection with the parabola and finding the minimum 't' value aka the earliest point of intersection. Suggestions are welcomed!

### DIFF
--- a/panda/src/collide/collisionBox.cxx
+++ b/panda/src/collide/collisionBox.cxx
@@ -509,30 +509,56 @@ test_intersection_from_parabola(const CollisionEntry &entry) const {
   PN_stdfloat cur_t;
   PN_stdfloat t1, t2;
   int intersecting_face = -1;
-  for (int i = 0; i < get_num_planes(); i++){
+  for (int i = 0; i < get_num_planes(); i++) {
     LPlane face = get_plane(i);
     bool intersects = false;
-    if (!face.intersects_parabola(t1, t2, local_p)){
-      continue;
-    }
-    if (t1 >= parabola->get_t1() && t1 <= parabola->get_t2()){
-      intersects = true;
-      cur_t = t1;
-    }
-    if (t2 >= parabola->get_t1() && t2 <= parabola->get_t2()){
-      if (intersects) {
-        cur_t = std::min(t1, t2);
-      } else {
-        cur_t = t2;
-	intersects = true;
+
+    if (face.intersects_parabola(t1, t2, local_p)) {
+      // t1 and t2 are both candidates for intersection
+      // if they are not less than t, no need to check them
+      // otherwise, check if they are in our parabola
+      if (t1 < t && t1 >= parabola->get_t1() && t1 <= parabola->get_t2()){
+        intersects = true;
+        cur_t = t1;
       }
-    }
 
-    if (intersects && cur_t < t){
-      t = cur_t;
-      intersecting_face = i;
-    }
+      if (t2 < t && t2 >= parabola->get_t1() && t2 <= parabola->get_t2()){
+        if (intersects) {
+          // both t1 and t2 intersect, take the smaller of the two.
+          cur_t = std::min(t1, t2);
+        } else {
+          cur_t = t2;
+          intersects = true;
+        }
+      }
 
+      if (!intersects) {
+        // the parabola does not intersect this face, skip to next one.
+        continue;
+      }
+
+      bool behind = true;
+      // we now check if the intersection point
+      // is behind all of the other planes.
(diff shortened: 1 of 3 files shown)


Moguri:
This looks good to me. @rdb, did you want to look over this?

hecris:
Ok, got it thanks. Will make the changes soon.

hecris:
I modified the algorithm to check if the intersection point is behind all of the other planes (by checking if distance to plane < 0). Also, if the parabola intersects from inside the box, then the surface normal should go inward (simply negate the vector). Like so:
It is currently passing all tests and when using @rdb 's visualizer it works for the most part.
The part where it fails is:
image
In this example, the parabola first intersects with face 5, but the distance from the point of intersection to the second plane (face 1) is positive, indicating that the point is in front of the second plane (which visually is not true):
image

If the mouse is moved just a little bit further, it intersects:
image

This part has been troubling me for quite a while and I can't seem to find the source of the problem. I also tried to use `CollisionBox::point_is_inside()` but the result was the same. Let me know if you have any suggestions.

panda/src/collide/collisionBox.cxx
```diff
@@ -493,6 +494,102 @@ test_intersection_from_ray(const CollisionEntry &entry) const {
...
+        }
+        if (get_plane(j).dist_to_plane(local_p.calc_point(cur_t)) > 0.0f) {
+          // our point is in front of this face, so turns out
+          // it does not collide with the box at this point
+          behind = false;
+          break;
+        }
+      }
+
+      if (behind) {
```
Moguri: This probably does not solve your current problem, but shouldn't this check if `cur_t` is smaller than `t` so we get the first `intersecting_face` and `t` value?

panda/src/collide/collisionBox.cxx
```diff
@@ -493,6 +494,102 @@ test_intersection_from_ray(const CollisionEntry &entry) const {
...
+    PT(CollisionEntry) new_entry = new CollisionEntry(entry);
+
+    LPlane face = get_plane(intersecting_face);
+
+    LPoint3 into_intersection_point = local_p.calc_point(t);
+    LVector3 normal = (has_effective_normal() && parabola->get_respect_effective_normal()) ? get_effective_normal() : face.get_normal();
+
+    // consider a point just before our intersection point
+    float eps = 1.0e-6f;
+    if (get_plane(intersecting_face).dist_to_plane(local_p.calc_point(t - eps)) < 0.0f) {
```
Moguri: `get_plane(intersecting_face)` is already stored in `face`

hecris: Thanks!

panda/src/collide/collisionBox.cxx
hecris: The check is actually done beforehand on lines 520 and 525:
https://github.com/panda3d/panda3d/pull/664/files#diff-6105c95465dfdba16cebccbd91fdd51eR520

Moguri: Ah up, I see it now.

Moguri:
@rdb Any further thoughts on the issue with the corner? If there isn't something obvious, I wonder if this is "good enough." Ideally our collision tests would be "perfect," but I know that is already not the case (e.g., I seem to remember you saying something about some of the capsule tests).

hecris:
@Moguri @rdb It seems like the algorithm is working correctly. Although the parabola is appearing to collide with the box, it actually does not. Here's a visualizer:
image

- As you can see, the parabola first intersects with the vertical plane, despite appearing to intersect with the horizontal plane first. (you can verify this by looking at the surface normals). 
- When the parabola intersects with the vertical plane, the point of intersection is not behind the horizontal plane. (you can verify this by looking at the box's face)
- Similarly, when the parabola intersects with the horizontal plane, the point of intersection is not behind the vertical plane. Overall, it returns no collision.
I'll go ahead and add some more unit tests to verify this. In the meantime, let me know what you guys think. Also, I think I will move onto parabola into inverse sphere soon.

rdb:
Could this be a perspective issue?  Does setting the lens to an `OrthographicLens()` fix it?

rdb:
Also, I don't think the collision normal should point inwards.  It should always point outward, I think.

hecris:
The collision test seems to work perfectly using `OrthographicLens()` !
And OK, I will make the surface normal go outwards.

rdb:
It's working perfectly, thanks so much!  Merged.

PR #665, opened 2019-06-03
Extend BulletBodyNode::add_shapes_from_collision_solids() to accept a TransformState

PointSource-xx:
This is useful for example when the collision node is rotated, is not centered to origin, or has several parent transforms applied on it.
Code is much the same as original, so adapt original function to use the expanded one.


Moguri:
@rdb This seems like a fine, simple change to me. Do you have any objections to it?

rdb (approved):
LGTM, can't believe I missed this

PR #668, opened 2019-06-05
CMake multi-configuration fixup

treamology:
This PR makes several changes that ensure all assets, config files, and include headers get placed into the right spots.


dtool/LocalSetup.cmake
```diff
@@ -193,7 +193,14 @@ else()
   set(intdir "${CMAKE_BUILD_TYPE}")
 endif()
 
-configure_file(dtool_config.h.in "${PROJECT_BINARY_DIR}/${intdir}/include/dtool_config.h")
+configure_file(dtool_config.h.in "${PROJECT_BINARY_DIR}/include/dtool_config.h")
+
+# TODO: Add the ability to customize dtool_config.h based on configuration.
+if(IS_MULTICONFIG)
```
treamology: We need to talk more about what behavior we want when generating dtool_config.h for multi-config generators. For example, do we want to use generator expressions or have a different .in file for each config?

CFSworks: We'll want to iterate over all potential configs and generate the dtool_config.h for each. This also implies a mechanism for detecting when options that are config-specific are left at their defaults, so the per-config default can be used.

rdb: Would it be easier to generate a single `dtool_config.h` that sets the values for different configs using some ifdefs?

CFSworks:
I also have 2 requests (ordinarily I'd ignore these or do them myself, but as you're GSoC, I figure a little more rigor is in order here):

1. Could you make sure your commit messages always begin with a `topic:`? For stuff changing CMake, that would be `CMake: `
2. I usually prefer to rebase instead of merge; if you're working locally or in a personal-use-only fork, you can use `git rebase upstream/cmake` (or whatever the tracking branch for this repo's `cmake` branch happens to be) to pick up all of your local `cmake` changes and put them on top; then you can just force-push to your personal fork and/or create a brand new branch for the sake of the PR.

treamology:
@CFSworks Gotcha, thanks for the pointers. I went ahead and rewrote history to remove the merge commit and put topics in front of all the commit messages.

treamology:
I just changed the file compositing logic to separate by extension, it would be great if you all could test it.

cmake/macros/Python.cmake
```diff
@@ -51,7 +51,14 @@ function(add_python_target target)
   target_link_libraries(${target} PKG::PYTHON)
 
   if(BUILD_SHARED_LIBS)
-    set(_outdir "${PROJECT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/${slash_namespace}")
+    if(CMAKE_GENERATOR STREQUAL "Xcode")
+      # This is explained in CompilerFlags.cmake
```
CFSworks: Code duplication scares me. I'm wondering if instead `intdir` should be encouraged for global use (e.g. defined globally in CompilerFlags and given an `UPPERCASE_WITH_UNDERSCORES` name) and we just use that everywhere that `CMAKE_CFG_INTDIR` was used previously.

cmake/macros/CompositeSources.cmake
```diff
@@ -47,35 +44,69 @@ function(composite_sources target sources_var)
     return()
   endif()
 
+  # Get rid of the periods in the extensions and create each list of composite
+  # sources.
+  set(extensions "")
+  foreach(extension ${COMPOSITE_SOURCE_EXTENSIONS})
+    string(MAKE_C_IDENTIFIER extension ${extension})
+    list(APPEND extensions ${extension})
```
CFSworks: `"${extension}"` (in both cases)

cmake/macros/CompositeSources.cmake
```diff
@@ -47,35 +44,69 @@ function(composite_sources target sources_var)
     return()
   endif()
 
+  # Get rid of the periods in the extensions and create each list of composite
+  # sources.
+  set(extensions "")
+  foreach(extension ${COMPOSITE_SOURCE_EXTENSIONS})
+    string(MAKE_C_IDENTIFIER extension ${extension})
```
CFSworks: Is `MAKE_C_IDENTIFIER` necessary/appropriate here? For the extensions that are eligible for compositing, we want to keep the extension exactly the same and not make any perturbations. (The `MAKE_C_IDENTIFIER` operation is generally used for naming C symbols; it's an odd choice for filename sanitation.)

cmake/macros/CompositeSources.cmake
```diff
@@ -96,6 +127,7 @@ function(composite_sources target sources_var)
 
           # Reset for the next composite file.
```
CFSworks: This block probably needs to be moved up one if level, because flushing out the composites can now happen due to the extension changing, so the previous assumption that `if(num_composite_sources GREATER 1)` was only false at the very end of compositing is no longer true. If `composite_sources` isn't cleared out, the file will just sit around and be added to the next composite file generated.

A way to reproduce the bug I'm suspecting would be to try compositing 2 .cxx files, 1 .mm file, and 2 .c files, in that order. Unless I'm missing something, the .mm file will end up sitting around in `composite_sources` and get added to the top of the generated .c file. (The expected outcome is the .cxx files and .c files are composited while the .mm file is left bare)

cmake/macros/CompositeSources.cmake
```diff
@@ -47,35 +44,69 @@ function(composite_sources target sources_var)
...
 
+    # Get the next source file's extention
+    if(num_sources GREATER 0)
+      list(GET sorted_sources 0 next_source)
+      get_filename_component(next_extension "${next_source}" EXT)
+    else()
+      set(next_extension "")
+    endif()
+    
     if(NOT generated AND NOT is_header AND NOT skip_compositing AND
```
CFSworks: Adding this logic to the pre-sorting step may actually make the main loop easier to understand. Perhaps this refactor is a good opportunity to do that? :D

cmake/macros/CompositeSources.cmake
```diff
@@ -47,35 +44,69 @@ function(composite_sources target sources_var)
...
+
+  # Sort each source file into a list.
+  foreach(source ${sources})
+    get_filename_component(extension "${source}" EXT)
+    string(MAKE_C_IDENTIFIER extension ${extension})
+    list(APPEND sources_${extension} "${source}")
+  endforeach()
+
+  set(sorted_sources "")
+  foreach(extension ${extensions})
```
CFSworks: Come to think of it, instead of discovering `extensions` dynamically, how about just iterating `COMPOSITE_SOURCE_EXTENSIONS`?

panda/metalibs/pandagl/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ if(NOT PANDAGL_PIPE_INCLUDE)
 endif()
 
 set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME "OpenGLDevel")
-add_metalib(pandagl ${MODULE_TYPE}
```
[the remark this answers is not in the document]
treamology: Many of the other modules are also missing `${MODULE_TYPE}`, and the ones that do have it seem to be less-used ones, so I thought it was something that was already being dropped. Is it that the modules that do specify `${MODULE_TYPE}` are meant to be loaded at runtime?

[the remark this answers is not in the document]
CFSworks: Now that you mention it, things like `p3openal_audio` and `p3ffmpeg` are missing their `${MODULE_TYPE}` usages. That should probably be addressed (once we're done with this PR, at least).

The specific meaning of `${MODULE_TYPE}` is "this is code meant to be loaded by `load_dso`/`dlopen`/`LoadLibrary`/etc., unless all of Panda is being built statically, then it should be a static library." It suppresses generating [import libraries](https://docs.microsoft.com/en-us/windows/desktop/Dlls/dynamic-link-library-creation#creating-an-import-library) on Windows, skips adding versioning information to .so filenames on Linux, etc. All of this is primarily meant to indicate to a user receiving a built copy of these binaries "hey, these are plugins loaded at runtime, not libraries you have to link against" but could be relevant on some future/obscure platforms that only allow dynamic loading if the library is explicitly built for that.

cmake/scripts/CopyPattern.cmake
```diff
@@ -0,0 +1,23 @@
...
+#
+# Usage:
+#   This script is invoked via add_custom_target, like this:
+#   cmake -D SOURCE=[source directory]
+#         -D DESTINATION=[destination directory]
+#         -D FILES_MATCHING="[globbing patterns passed to file(COPY)]"
+#         -P CopyPattern.cmake
+
+if(DEFINED FILES_MATCHING)
+  string(REPLACE " " ";" FILES_MATCHING_ARGS ${FILES_MATCHING})
```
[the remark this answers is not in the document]
treamology: `add_custom_command` does not deal with semicolon separated lists well. The argument
```
-DFILES_MATCHING="PATTERN;*.rgb;PATTERN;*.png;PATTERN;*.jpg;PATTERN;*.wav"
```
yields the error
```
file called with unknown argument "PATTERN *.rgb PATTERN *.png PATTERN
  *.jpg PATTERN *.wav"
```
It seems that it processes the semicolons by turning them into spaces, but it keeps the quotes around. Omitting the quotes leads to the list being cut off after the first item. I'm sure it's possible using some trickery, but I wanted to keep the arguments easily readable within `add_custom_command`, and this hack seemed to be the best way to do it. I'll probably write a comment explaining my thinking in the script itself.

cmake/macros/Python.cmake
treamology: I actually did do something like this on my iOS branch to fix an issue with $(EFFECTIVE_PLATFORM_NAME), I must have forgotten to move it over. But the code duplication in this case comes from the fact that make would spam warnings if the output directory was set to `CMAKE_CFG_INTDIR`. Making it a generator expression fixed it in these two cases, but we can't have this generator expression being used everywhere we're using `CMAKE_CFG_INTDIR`. This code duplication doesn't bother me as much since it's only happening two times though, both in order to bypass the same warnings.

cmake/macros/CompositeSources.cmake
treamology: I'm using it here to create lists that will be used as buckets for each file extension, so I'll only have to loop over the source list once instead of however many source types there are. Then they're combined into a single, now sorted list.

dtool/Config.cmake
```diff
@@ -22,13 +22,13 @@ if(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
   set(IS_FREEBSD 1)
 endif()
 
+get_property(IS_MULTICONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
```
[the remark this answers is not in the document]
treamology: It appears to be working on Jessie without any warnings or errors.

treamology:
These two force-pushes were me trying to fix the travis builds timing out, I didn't end up changing anything with them.

PR #675, opened 2019-06-20
CollisionBox into CollisionCapsule

hecris:
Added collision test for box into capsule: splits the capsule into a cylinder and two hemispheres and tests for intersections with the box. Suggestions are welcomed!

### DIFF
--- a/panda/src/collide/collisionCapsule.cxx
+++ b/panda/src/collide/collisionCapsule.cxx
@@ -190,11 +190,9 @@ test_intersection_from_box(const CollisionEntry &entry) const {
   point_on_segment[2] = a[2];
   if (closest_point[1] < std::min(a[1], b[1])) {
     point_on_segment[1] = std::min(a[1], b[1]);
-  }
-  else if (closest_point[1] > std::max(a[1], b[1])) {
+  } else if (closest_point[1] > std::max(a[1], b[1])) {
     point_on_segment[1] = std::max(a[1], b[1]);
-  }
-  else {
+  } else {
     point_on_segment[1] = closest_point[1];
   }
 
@@ -204,7 +202,15 @@ test_intersection_from_box(const CollisionEntry &entry) const {
   }
 
   PT(CollisionEntry) new_entry = new CollisionEntry(entry);
-  LVector3 normal = (closest_point - point_on_segment);
+  LVector3 normal;
+  if (point_on_segment == closest_point) {
+    // If the box directly intersects the line segment, use
+    // the box's center to determine the surface normal.
+    normal = (box->get_center() * wrt_mat * _inv_mat) - point_on_segment;
+    if (closest_point != a && closest_point != b) normal[1] = 0;
+  } else {
+    normal = (closest_point - point_on_segment);
+  }
   normal.normalize();
   LPoint3 surface_point = point_on_segment + normal * _radius;
   new_entry->set_interior_point(_mat.xform_point(closest_point));
--- a/panda/src/collide/collisionCapsule.cxx
+++ b/panda/src/collide/collisionCapsule.cxx
@@ -149,73 +149,68 @@ test_intersection_from_box(const CollisionEntry &entry) const {
   const LMatrix4 wrt_mat = entry.get_wrt_mat();
   const LMatrix4 inv_wrt_mat = entry.get_inv_wrt_mat();
 
-  // First, we test if the box intersects the cylinder part of the 
-  // capsule. We find the point on the box closest to the center of
-  // the capsule. To do this, convert the capsule to the box's
-  // coordinate space for simpler calculations.
+  // We find the closest point on the box to the line
+  // segment defined by the endpoints of the capsule. To do
+  // this, convert the capsule into the box's coordinate space.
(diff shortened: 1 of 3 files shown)


Moguri:
One potential issue I see is if the box collides deep enough into an end cap to also collide with the cylindrical portion of the capsule. In this case, I think the collision will be reported as inside the end cap (where the end cap intersects with the cylinder) instead of on the surface of the end cap.

In other words, if the box collides with both an end cap portion and the cylinder portion, which collision do we chose to report?

rdb:
While it's working fine in the below code, if you comment out the second CollisionTube definition instead, it fails to properly register against the cylindrical section in all cases:
```python
from panda3d.core import *
#load_prc_file_data("", "notify-level-collide debug")

#base.disableMouse()
base.cam.set_y(-20)

lens = OrthographicLens()
lens.set_film_size(10 * base.getAspectRatio(), 10)
base.camNode.set_lens(lens)

tube = CollisionTube((0, 0, 0), (0, 0, 2), 2)
#tube = CollisionTube((0, 0, 0), (4, 0, 4), 0.3)

box = CollisionBox((-1, -1, -1), (1, 0.5, 1))

into_node = CollisionNode("into")
into_node.add_solid(tube)
into_node.set_into_collide_mask(1)
into_np = render.attach_new_node(into_node)
into_np.show()
#into_np.set_scale(2)
#into_np.set_hpr(51, -25, 100)
#into_np.set_hpr(0, 45, 45)

from_np_parent = render.attach_new_node("from-parent")

from_node = CollisionNode("from")
from_node.add_solid(box)
from_node.set_from_collide_mask(1)
from_np = from_np_parent.attach_new_node(from_node)
from_np.show()

#handler = CollisionHandlerPusher()
#handler.add_collider(from_np, from_np)

handler = CollisionHandlerEvent()

base.cTrav = CollisionTraverser()
base.cTrav.show_collisions(base.render)
base.cTrav.add_collider(from_np, handler)

def update(task):
    if base.mouseWatcherNode.has_mouse():
        pos = base.mouseWatcherNode.get_mouse()
        from_np_parent.set_pos(pos[0] * 10, 0, pos[1] * 10)
    return task.cont

base.taskMgr.add(update)

base.run()
```

hecris:
@Moguri that's a good point, thanks! As you can see, the cylinder is checked first so it has priority over the endcaps. So, you are correct that the reported collision will be inside the cylinder. IMO, this is an issue because the surface point should be on the endcap and not on the cylinder.

When I am testing for intersections with the endcaps, I am using a box-into-sphere test, not a box-into-hemisphere test. So, if I gave the endcaps priority over the cylinder, the box might intersect with the part of the spheres that is inside the cylinder (which would return a false collision). Now, I see that giving the capsule priority can be problematic as well.

A potential solution might be to test for intersections with the endcaps first, and then with the cylinder. However, the intersection with the endcaps would be box-into-hemisphere instead of box-into-sphere. I'm going to try this out. In the meantime let me know if you have any thoughts.

Moguri:
I am wondering if we could perform both tests and then pick a preferred one if both are hits. I don't know if there is a simple min/max type thing we can do. Maybe pick the collision with the greatest magnitude on the local primary axis of the capsule?

rdb:
@Moguri do you have a case in mind wherein a box approaches a capsule where it would result in it being pushed the wrong way if it collides with both an endcap and the cylindrical bit?

Moguri:
@rdb I think this case would only show up if the box was able to penetrate a lot in one frame. For example, if a box was approaching on the primary axis of the capsule and went from not colliding to colliding with the cylindrical part in one frame. Then the pusher would likely push the box to outside of the cylindrical part, but still within the endcap. After another frame, the box could get pushed out of the endcap since it would no longer be colliding with the cylindrical part (assuming the box is only being moved by the pusher at this point).

Moguri:
Where are we at with this?

hecris:
Hi guys, instead of testing the box with a cylinder and two spheres, I find the closest point on the box to the line segment defined by the endpoints of the capsule. So, we don't have to worry about whether the endcaps or the cylinder gets tested first.
However, I am stuck on finding the surface normal. Currently, I take this closest point on the box to be the interior point. Then, I find the point on the line segment closest to this point (might be easier to understand with code):
https://github.com/panda3d/panda3d/pull/675/files#diff-b101ff3ec8e8f1cdb11e1e1bc754ef6dR187

I get the surface normal by subtracting these two points. However, when the box directly intersects with the line segment, these two points are the same. So subtracting them would not give us a normal. Let me know if you guys have any suggestions.

rdb:
@hecris If the box intersects with the line segment, you could take the point on the line segment as interior point, and choose the direction towards the box center as surface normal (and derive the surface point from that).  Would that work?

rdb:
With that, I am suggesting that you set the Y of the normal to zero in that case, since we know that the box is colliding with the cylindrical part too.

hecris:
Thanks for the suggestion @rdb , I did as you recommended and it looks OK. Let me know what you guys think.

rdb:
Great work, thanks so much!  Merged.

PR #677, opened 2019-06-21
Add procedural geometry samples

Epihaius:
This is a collection of several samples, combined with a simple sizer-based custom GUI system, based on the creation of 3D model primitives.

The samples range from very basic to quite advanced, with a couple of demo's intended for pure entertainment thrown in.
Apart from a simple cube, five standard 3D model primitive types can be generated with the code provided in the `src/prim` package: box, sphere, cylinder, cone and torus.
Each of these can be created with a plethora of options, such that the resulting models can vary dramatically in shape and UV-mapping.


Epihaius:
Hi,

The primary reason for creating these samples is that I feel that the `procedural-cube` and `fractal-plants` samples have become quite outdated, with some function calls like `addData3f` which shouldn't be used.

But there are a couple of other reasons as well.

Mainly, to offer Panda users the possibility to create the 5 most common 3D model primitives right inside of their code, without having to resort to modelling programs, export and import just to quickly get some basic shapes into their scene.
But rather than just a plain old tired box, sphere, cylinder, cone or torus, I wanted to offer a lot of options when generating these models.
For example, a torus can not only be made to look like the typical doughnut shape, but by setting twist, rotation, smoothing and thickness properties it can turn into something completely different.
But even a simple box can be made into entirely different shapes, by omitting one or more sides and specifying an inner thickness.
These shapes would be especially useful as placeholder models, ideal for rapid prototyping. One could make objects that vaguely look like a bed, sofa, round desk or whatever very quickly, even with custom UV-mapping.

To create these primitives, a ModelMaker base class was implemented, with BoxMaker, SphereMaker, etc. classes deriving from it. These classes are more or less modelled after the CardMaker class, but instead of getter and setter methods, they have a property-based interface.
Would it be possible to one day add these classes to the code base of Panda3D? Many users would probably love to be able to access these without having to copy my code first. And poor CardMaker wouldn't feel so lonely anymore :D .

Then there is the custom GUI system I partially ported from my Panda3D Studio project, and it is entirely sizer-based, meaning that the layout of its widgets is preserved when resizing the window. These widgets themselves have nothing to do with DirectGUI, but are made up of 3D model primitives as generated by the ModelMaker classes (and as such are also an exercise on geometry manipulation).
To offer this kind of GUI will hopefully make for a nice bonus.

About the included samples:
* `basic.py` and `medium.py` both create just a square, but the former is the only sample to use GeomVertexWriters while the latter uses memoryviews to do this;
* `advanced.py` details the creation of a cube, and shows how to add and remove parts of its geometry;
* `primitives_showcase.py` allows one to see the shape generated by a ModelMaker using the property values read from a corresponding text file; having this file open in a text editor while running this sample makes it easy to see the result of a property change almost immediately;
* more demo than sample, `primitives_assembly.py` shows off the gradual construction of model primitives over time;
* and as a modernized counterpart of the `fractal-plants sample`, be sure to check out the `growing_vines.py` sample. Enjoy the show!

Oh and there's some procedural texture generation in there too :) .

gr4viton:
This is awesome! :)

gr4viton:
Would it be possible to release the BoxMaker and other procedural geometry maker classes as a separate package? I would really like to use them in my opensource project :)

Epihaius:
@gr4viton Thank you for your interest in my code. :)

You can find the repo I just created for this [here](https://github.com/Epihaius/procedural_panda3d_model_primitives).

PR #682, opened 2019-07-06
showbase: Correct functionality of `weightedChoice`

pythonengineer:
I'm not exactly sure why it was left like this, but if you call it under certain arguments it will crash because it attempts to return a variable which is out of scope. This simply corrects the functionality to what was intended according to the comments to avoid such a crash.

A test has also been added for this function.

### DIFF
--- a/direct/src/showbase/PythonUtil.py
+++ b/direct/src/showbase/PythonUtil.py
@@ -1132,7 +1132,7 @@ def weightedChoice(choiceList, rng=random.random, sum=None):
     weights, pass it in 'sum'."""
     # Throw an IndexError if we got an empty list.
     if not choiceList:
-        raise IndexError('Cannot choose from an empty sequence') from None
+        raise IndexError('Cannot choose from an empty sequence')
 
     # TODO: add support for dicts
     if sum is None:
--- a/direct/src/showbase/PythonUtil.py
+++ b/direct/src/showbase/PythonUtil.py
@@ -1138,14 +1138,16 @@ def weightedChoice(choiceList, rng=random.random, sum=None):
 
     rand = rng()
     accum = rand * sum
+    lastItem = 0.
     for weight, item in choiceList:
+        lastItem = item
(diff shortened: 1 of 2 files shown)


rdb:
Thanks for the unit test!

However, I'm not sure that the function should return `0.` in the corner case.  Shouldn't it always return an item that is part of the choice list, or raise an exception, or return a neutral value like `None`?  Can you construct a set of inputs that triggers the corner case?

Furthermore, binding to `lastItem` seems to be unnecessary as you can just do `item = 0.` before the loop.

rdb:
Looking a little more closely at the code, it looks like the "certain arguments" case that you are referring to is when you are passing it an empty list, is that correct?  If so, I'm even less sure that we should return `0.` in this case.

I would suggest that we simply check for an empty list and raise an exception in this case, similar to what Python does:
```python
>>> random.choice([])
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.7/random.py", line 261, in choice
    raise IndexError('Cannot choose from an empty sequence') from None
IndexError: Cannot choose from an empty sequence
```

pythonengineer:
@rdb You are right that it shouldn't return `0.`, upon an empty list it will now properly return `None`.

I didn't make it raise an exception because sometimes people would rather a function return `None` upon invalid information than have an exception raised which may kill the application. This should work fine.

rdb:
I think raising an exception makes more sense than returning `None`; if you're asking Panda to choose a value from a given set of no values, that is an invalid input.

There's no need for an exception to "kill the application" if it is properly caught with a try…except block or if you simply check the input such as by doing `weightedChoice(list) if list else None`.

pythonengineer:
@rdb Fair enough. It will now raise an IndexError if an empty list is passed.

rdb:
Committed, with one change: I've used `with pytest.raises(IndexError):`, which is a more idiomatic way to check for an exception being raised.

PR #688, opened 2019-07-19
shaderAttrib: improve shader flags

lachbr:
The flags in the enum are actual bit flags instead of indices. This allows for setting/clearing a mask of flags in one go, rather than having to set each flag in its own function call. Existing code shouldn't be affected.

### DIFF
--- a/panda/src/pgraph/shaderAttrib.I
+++ b/panda/src/pgraph/shaderAttrib.I
@@ -93,8 +93,8 @@ get_instance_count() const {
  *
  */
 INLINE bool ShaderAttrib::
-get_flag(int index) const {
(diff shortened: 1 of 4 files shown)


rdb:
Could you add unit tests for affected behaviours?  Thanks!

Moguri:
This looks like it could break existing applications if they were relying on the actual value of the flags, which seems like a bad idea. This could be seen as an API break, but I would also argue that the values of enumerated types should not be guaranteed.

rdb:
I personally think we shouldn't make guarantees about stability of enum values in our API (other than our usual ABI guarantees), except when the values go into .bam files (which is not the case here).

PR #691, opened 2019-07-25
New CollisionHeightfield

hecris:
A new `CollisionSolid` to deal with collisions on uneven terrains given a heightfield image (currently accepts `PNMImage`). The implementation uses a quad tree to reduce the number of triangles tested. You can use `GeoMipTerrain` or any other terrain renderer to start visualizing the collisions. It's currently under development, so the code is still messy and some bugs may come up. I am currently working on finding surface normals, writing unit tests, debugging, and finding ways to improve the sphere and box collision tests. Any suggestions are welcomed!


panda/src/collide/collisionHeightfield.I
```diff
@@ -0,0 +1,23 @@
...
+  _volume_pcollector.flush_level();
+  _test_pcollector.flush_level();
+}
+
+INLINE PNMImage &CollisionHeightfield::
+heightfield() {
+  return _heightfield;
+}
+
+INLINE double CollisionHeightfield::
```
Moguri: Should this be PN_stdfloat?

panda/src/collide/collisionHeightfield.h
```diff
@@ -0,0 +1,159 @@
+#ifndef COLLISIONHEIGHTFIELD_H
+#define COLLISIONHEIGHTFIELD_H
+
+#include "pandabase.h"
+#include "collisionSolid.h"
+#include "pnmImage.h"
+
+using std::vector;
```
Moguri: I believe we should be avoiding `using` statements in header files (and maybe in general) since it can lead to polluted namespaces. However, I will defer to @rdb's opinion on this.

rdb: I agree with this; this should not be in the header file.

hecris:
@Moguri Modifying the heightfield image probably won't require rebuilding the quadtree, but instead traversing through it and updating each node's data. The same applies for the other parameters of `CollisionHeightfield` (`max_height`, `subdivisions`). Setting a new `max_height` only requires updating the height values of each quad tree node. Setting a new `subdivisions` would require adding/deleting quad tree nodes. These methods could be convenient, so I will add them.

The use of the min/max values of `QuadTreeNode` is to construct a box as you said. I wasn't sure if `QuadTreeNode` would be useful out of `CollisionHeightfield` but if it is worth it I can definitely look into it.

panda/src/collide/collisionHeightfield.I
hecris: Changed it to PN_stdfloat :+1:

panda/src/collide/collisionHeightfield.h
hecris: OK, removed it :+1:

hecris:
@rdb @Moguri  I just pushed my unit tests. I don't have an official sample game yet but this sample program should be enough to start visualizing the collisions:
```python
from panda3d.core import *
load_prc_file_data("", "notify-level-collide debug")
base.cam.set_y(-20)
 
# Set up heightfield
img = PNMImage(33, 33, 1)
img.set_gray_val(10, 10, 255)
 
# Render terrain
terrain = GeoMipTerrain("mySimpleTerrain")
terrain.setHeightfield(img)
terrain.setBruteforce(True)
terrain.get_root().set_sz(10)
terrain.getRoot().reparentTo(render)
# texture = loader.loadTexture('ground.jpg')
# terrain.get_root().set_texture(texture)
terrain.generate()
 
sphere = CollisionSphere((10, 10, 10), 2)
heightfield = CollisionHeightfield(img, 10, 3)
 
into_node = CollisionNode("into")
into_node.add_solid(heightfield)
into_node.set_into_collide_mask(1)
into_np = render.attach_new_node(into_node)
into_np.show()
 
from_np_parent = render.attach_new_node("from-parent")
from_node = CollisionNode("from")
from_node.add_solid(sphere)
from_node.set_from_collide_mask(1)
from_np = from_np_parent.attach_new_node(from_node)
from_np.show()
 
handler = CollisionHandlerEvent()
base.cTrav = CollisionTraverser()
base.cTrav.show_collisions(base.render)
base.cTrav.add_collider(from_np, handler)
 
def move(direction):
    pos = from_np.get_pos()
    pos += direction
    from_np.set_pos(pos)
 
base.accept('arrow_up', move, [(0, 1, 0)])
base.accept('arrow_down', move, [(0, -1, 0)])
base.accept('arrow_right', move, [(1, 0, 0)])
base.accept('arrow_left', move, [(-1, 0, 0)])
base.accept('w', move, [(0, 0, 1)])
base.accept('s', move, [(0, 0, -1)])
 
base.run()
```

panda/src/collide/collisionHeightfield.I
```diff
@@ -0,0 +1,54 @@
+INLINE CollisionHeightfield::
```
rdb: There's no file header in this file.  Please follow the example of other .I files.

Similarly, please add a docstring comment above each method definition (even if it's empty at first) of the form:
```cpp
/**
 *
 */
```

panda/src/collide/collisionHeightfield.I
```diff
@@ -0,0 +1,54 @@
...
+
+INLINE PNMImage CollisionHeightfield::
+get_heightfield() {
+  return _heightfield;
+}
+
+INLINE void CollisionHeightfield::
+set_heightfield(PNMImage heightfield) {
+  int r = _heightfield.get_read_x_size();
+  int c = _heightfield.get_read_y_size();
```
rdb: Is there a reason you are using `get_read_x_size()` instead of `get_x_size()`?

panda/src/collide/collisionHeightfield.I
```diff
@@ -0,0 +1,54 @@
...
+
+INLINE void CollisionHeightfield::
+set_max_height(PN_stdfloat max_height) {
+  _max_height = max_height;
+  fill_quadtree_heights();
+}
+
+INLINE int CollisionHeightfield::
+get_subdivisions() {
+  return _subdivisions;
```
rdb: Let's call this `get_num_subdivisions()`, consistent with other Panda APIs, since we are getting the number of subdivisions and not the subdivisions themselves, right?

panda/src/collide/collisionHeightfield.cxx
```diff
@@ -0,0 +1,773 @@
+/**
+ * PANDA 3D SOFTWARE
+ * Copyright (c) Carnegie Mellon University.  All rights reserved.
+ *
+ * All use of this software is subject to the terms of the revised BSD
+ * license.  You should have received a copy of this license along
+ * with this source code in a file named "LICENSE."
+ *
+ * @file CollisionHeightfield.cxx
```
rdb: Should be `collisionHeightfield.h`

panda/src/collide/collisionHeightfield.cxx
```diff
@@ -0,0 +1,773 @@
...
+ */
+void CollisionHeightfield::
+set_subdivisions(int subdivisions) {
+  // The number of subdivisions should not be negative
+  nassertv(subdivisions >= 0 && subdivisions <= 10);
+  // Determine the number of quadtree nodes needed
+  // for the corresponding number of subdivisions.
+  int nodes_count = 0;
+  for (int i = 0; i <= subdivisions; i++) {
+    nodes_count += pow(4, i);
```
rdb: I think `pow()` is floating point.  In this case it is probably better to use this:
```cpp
    nodes_count += 1 << (i * 2);
```

panda/src/collide/collisionHeightfield.cxx
```diff
@@ -0,0 +1,773 @@
...
+  PN_stdfloat radius = csqrt(radius_2);
+
+  IntersectionParams params;
+  params.center = center;
+  params.radius = radius;
+
+  vector<QuadTreeIntersection> intersections;
+  intersections= find_intersections(sphere_intersects_box, params);
+
+  if (intersections.size() == 0) return nullptr;
```
rdb: This is slightly more readable:
```cpp
  if (intersections.empty()) {
    return nullptr;
  }
```

panda/src/collide/collisionHeightfield.cxx
```diff
@@ -0,0 +1,773 @@
...
+
+  LPoint3 point;
+  LPoint3 closest_point;
+  PN_stdfloat dist;
+  PN_stdfloat dist_min = FLT_MAX;
+
+  bool intersected = false;
+  #define in_box(x, y, node) (x >= node.area.min[0] && y >= node.area.min[1] \
+                           && x <= node.area.max[0] && y <= node.area.max[1])
+  for (unsigned i = 0; i < intersections.size(); i++) {
```
rdb: Just use `size_t` instead of `unsigned`, since that's what `size()` returns anyway

panda/src/collide/collisionHeightfield.cxx
```diff
@@ -0,0 +1,773 @@
...
+                         const Triangle &triangle) {
+  // Implementation of Möller-Trumbore algorithm
+  const PN_stdfloat EPSILON = 1.0e-7;
+  PN_stdfloat a,f,u,v;
+  LVector3 edge1, edge2, h, s, q;
+  edge1 = triangle.p2 - triangle.p1;
+  edge2 = triangle.p3 - triangle.p1;
+  h = delta.cross(edge2);
+  a = dot(edge1, h);
+  if (a > -EPSILON && a < EPSILON) {
```
rdb: We have the `IS_NEARLY_ZERO` macro for this.

panda/src/collide/collisionHeightfield.cxx
```diff
@@ -0,0 +1,773 @@
...
+void CollisionHeightfield::
+register_with_read_factory() {
+  BamReader::get_factory()->register_factory(get_class_type(), make_CollisionHeightfield);
+}
+
+PStatCollector CollisionHeightfield::_volume_pcollector(
+      "Collision Volumes:CollisionHeightfield");
+PStatCollector CollisionHeightfield::_test_pcollector(
+      "Collision Tests:CollisionHeightfield");
+TypeHandle CollisionHeightfield::_type_handle;
```
rdb: This should not be at the bottom.  Follow the example of other `.cxx` files; these go under the includes.

panda/src/collide/collisionHeightfield.h
```diff
@@ -0,0 +1,167 @@
+#ifndef COLLISIONHEIGHTFIELD_H
```
rdb: Missing file header

panda/src/collide/collisionHeightfield.h
```diff
@@ -0,0 +1,167 @@
...
+ * terrain given a heightfield image. A quad tree is implemented to
+ * significantly reduce the amount of triangles tested. Each quad
+ * tree node represents a sub-rectangle of the heightfield image
+ * and thus a box in 3D space.
+ * */
+class EXPCL_PANDA_COLLIDE CollisionHeightfield : public CollisionSolid {
+PUBLISHED:
+  CollisionHeightfield(PNMImage heightfield,
+                       PN_stdfloat max_height, int subdivisions);
+  ~CollisionHeightfield() { delete[] _nodes; }
```
rdb: → .I file

panda/src/collide/collisionHeightfield.cxx
rdb: Ugh, yes, sorry, the point was the lowercase `c`

hecris:
@rdb My intention was for users to use `CollisionHeightfield` alongside the terrain generator of their choice (`GeoMipTerrain`, etc). Do you think `CollisionHeightfield` should generate geometry?

rdb:
Yes, it is the intention that it should be used alongside an existing generator, as you say, in the regular case.

However, for debugging collisions, all other collision solids (including CollisionPolygon) can show a white viz solid when using `.show()` on a NodePath that is pointing to a CollisionNode.  It doesn't need to be particularly accurate, but it's useful to show something when debugging collisions.  You could probably just invoke GeoMipTerrain or something like that.

hecris:
Ok, I'll look into it. Would it be okay if only the triangles that are being tested for are shown?

Moguri:
I think it would be handy to show all triangles, and then maybe have the triangles being tested for in a different color.

Moguri:
What can we do to move this forward? Is the lack of visualization the only known issue at the moment?

hecris:
Hey @Moguri , I think it would also be good to separate the quad tree stuff from CollisionHeightfield, but visualization is probably the most important. I plan to revisit this when I'm done with my school work (during winter recess)

hecris:
@rdb @Moguri I'm on break so I'm picking `CollisionHeightfield` back up! I'm going to work on separating the quadtree into its own class, and also adding visualization. I'll update you guys along the way :+1:

Moguri:
I did not realize there were some commits made to this branch back in January. @hecris, is this PR ready for review again? I am not too concerned about separating out the quadtree, and I would be fine adding it  as a "good first task" to the tracker if that means we can get this PR submitted.

hecris:
@Moguri , okay sounds good. Yes, I think this PR is ready for review again.

hecris:
Hi @rdb and @Moguri .I would like to continue this hopefully when my summer break comes. I think a good first start would be decomposing this large PR into smaller ones. That being said, I was wondering if you guys are satisfied with the implementation of `collisionHeightfield`, that is, using a quad tree?

rdb:
Hey @hecris, that would be great!  Sorry for missing your last comment.  I'll take a closer look this week!

panda/src/collide/collisionHeightfield.cxx
```diff
@@ -0,0 +1,813 @@
...
+  int cols = _heightfield.get_y_size();
+  vector<Triangle> triangles;
+  if (x < 0 || y < 0 || x >= rows || y >= cols)
+    return triangles;
+
+  Triangle t;
+  int y2 = cols - 1 - y;
+  bool odd = (x + y2) % 2;
+  #define in_bounds(x, y) (x >= 0 && y >= 0 && x < rows && y < cols)
+  #define get_point(dx, dy) LPoint3(x + dx, y - dy, get_height(x + dx, y2 + dy))
```
rdb, line 696: I would rather you define this as a static inline function declared at the top of this file.  It's not great style to use inline macros for this.

panda/src/collide/collisionHeightfield.cxx
```diff
@@ -0,0 +1,813 @@
...
+  }
+
+  LPoint3 point;
+  LPoint3 closest_point;
+  PN_stdfloat dist;
+  PN_stdfloat dist_min = FLT_MAX;
+
+  bool intersected = false;
+  #define in_box(x, y, node) (x >= node.area.min[0] && y >= node.area.min[1] \
+                           && x <= node.area.max[0] && y <= node.area.max[1])
```
rdb, line 302: No inline macros please, see comment below.

rdb:
Also, now that we have CMake in the master branch, do you think you could merge in the master branch to your branch and add your files to the CMakeLists.txt?

Moguri:
@hecris Any thoughts on @rdb's style comments or updating the CMakeLists.txt? It seems like we're getting really close to getting this PR over the finish line, and I'd hate to see it fall through now. :)

hecris:
@rdb @Moguri sorry, I forgot to get back to you guys, it's been a hectic summer! I will make the style changes and update the CMakeLists.txt

hecris:
@rdb @Moguri hey! I replaced all the inline macros. I found it easier not to convert them to static inline functions, since they're not really relevant outside of the functions they're needed in + they used some local variables, so I just inserted the code into the functions. I also attempted to add the files to CMakeLists.txt, but I'm not too familiar with CMake and it seems like it's not passing the Travis builds. Could you point me in the right direction? Also yes, viz is implemented!

Moguri:
Looks like your missing some includes for other items in the `collides` library. With makepanda, your new files were added to the end of the composite file, which means it was at the end of that large, unified translation unit. The "unity" CMake builds do something similar, which is why they are working. Looks like the includes for the following classes are missing:

* CollisionRay
* CollisionNode
* CollisionSphere
* GeomVertexWriter
* GeomTriangles

hecris:
Thanks for the help @Moguri , I added the missing includes and it looks like the builds are passing!

Moguri:
@rdb have your concerns been addressed? Can we get this merged?

PR #699, opened 2019-08-02
Scrollbar width changeable after initialization

fireclawthefox:
Scrollbar width wasn't changeable after the scrollbar was initialized. This adds a function to make after initialization changes possible.


rdb:
It looks like this caused a regression in applications that explicitly set `verticalScroll_frameSize`, since that property is now being overridden.

PR #700, opened 2019-08-02
Orientation Fixes for Slider and Scrollbar

fireclawthefox:
Fixed sizing after orientation change
Added missing orientation change logic for the slider


rdb:
Hmm, what if the user has specified a custom frame size?  Would this not then reset that custom size to the default?

fireclawthefox:
Ah, yes that's true indeed. Maybe swapping the frameSize bits instead of overwriting them could work if we know the previous orientation. Will take a look into this later.

rdb:
Swapping probably makes sense.

rdb:
It looks like this might be ready to go, do you intend to add a unit test?

fireclawthefox:
Do you think this simple unittest will be sufficient or should it be extended some more?

PR #701, opened 2019-08-02
Positionable item container

fireclawthefox:
Make items position changeable (left/right/above/below)
Added positions to DirectGuiGlobals
Make item container scrollable
Enhanced checks if item frame is out of window bounds


fireclawthefox:
@rdb, as requested, here's a quick sample to demonstrate what the changes in this PR offer. It shows the use of all possible directions to open the OptionMenu as well as item scrolling if there are too many items in the menu:
https://gist.github.com/fireclawthefox/33e2f89465a5c82899855a999d6b4d1f

PR #702, opened 2019-08-02
Make passing entry to DirectEntryScroll optional

fireclawthefox:
Made passing an exiting DirectEntry while initializing optional
Added function to set an Entry

### DIFF
--- a/direct/src/gui/DirectEntryScroll.py
+++ b/direct/src/gui/DirectEntryScroll.py
@@ -7,6 +7,8 @@ from .DirectFrame import *
 from .DirectEntry import *
 
 class DirectEntryScroll(DirectFrame):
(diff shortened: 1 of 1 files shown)


rdb:
I'm just not sure that it should call `destroy()` on the old entry when assigning a new one, since this class isn't what created the entry to begin with.  Thoughts?

fireclawthefox:
Hm, yeah, probably the function name doesn't really describes this part of the function well. Maybe check if we already have an entry and output an error or warning would be better. For another way, we'd probably need to enhance the class further to be able to hold more than one entry at the same time.

rdb:
I think it's fine, but I'm just suggesting that perhaps the `.destroy()` call shouldn't be there, but I'm happy to hear arguments on why it should.

PR #717, opened 2019-08-26
The CMake buildsystem

CFSworks:
This does not (yet) handle:
- TinyDisplay
- maya tools
- max tools
- Creating an NSIS installer or other OS-specific packages
- `make wheel`
- Android
- libRocket
- PhysX
- SpeedTree

This should be testable (on all platforms) with:
- `mkdir build && cd build`
- `cmake -DTHIRDPARTY_DIRETORY=/path/to/thirdparty ..`
- `cmake --build .`

Hopefully this is without problem in most cases and should be suitable for merging to `master` even if we aren't yet telling people to drop makepanda. That is, while it's not all it can be, it should be far enough along that it doesn't create (substantial) confusion for users, and that we can even start using it ourselves primarily. Hence, this is not a draft PR.

Note there are some slight issues with the thirdparty zip currently which might require turning off certain features (HAVE_...) for now; I hope some of these are impetus for updating the thirdparty zip, though.

The situation with Travis will require attention. I consider a `-DCOMPOSITE_SOURCE_LIMIT=0` build to be an essential code quality check, but unfortunately that currently blows past the 50 minute limit on Travis jobs. Enabling ccache doesn't appear to help as Travis will even skip uploading the cache if the job limit is exceeded. A discussion about what jobs are appropriate should take place before we merge this down. In particular, we need to come up with some representative matrix of:
- OS (macOS/Windows/Linux)
- Compiler (clang/gcc/MSVC)
- Build backend (Makefiles/MSbuild/Xcode/Ninja)
- Build frontend (CMake/makepanda)
- Python version
- Composite on/off
- Metalibs on/off


dtool/Config.cmake
```diff
@@ -0,0 +1,629 @@
...
+#
+# This is the end of the PRC variable customization section.  The
+# remaining variables are of general interest to everyone.
+#
+
+
+option(HAVE_P3D_RTDIST
+  "You may define this to build or develop the Panda3D rtdist,
+the environment packaged up for distribution with the plugin."
+  OFF)
```
rdb: As far as I'm concerned we can axe the rtdist.  If CMake becomes its grave, I shall not weep.

dtool/Package.cmake
```diff
@@ -0,0 +1,703 @@
...
+      "Your version of CMake is too old; please upgrade or unset THIRDPARTY_DIRCTORY to continue.")
+  endif()
+
+  # Dig up the actual "libs" directory
+  if(APPLE)
+    set(_package_dir "${THIRDPARTY_DIRECTORY}/darwin-libs-a")
+
+  elseif(WIN32)
+    if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+      set(_package_dir "${THIRDPARTY_DIRECTORY}/win-libs-vc14-x64")
```
rdb: Probably instead of assuming `vc14` we should check the MSVC version (though, I note that 2015, 2017, and 2019 are all minor versions under "14", bless Microsoft, I don't know whether this is future-proof enough)

dtool/Package.cmake
```diff
@@ -0,0 +1,703 @@
...
+    Bullet
+    Cg
+    Eigen3
+    FCollada
+    FFMPEG
+    FMODEx
+    Freetype
+    HarfBuzz
+    JPEG
+    LibSquish
```
rdb: Why do we prefix Squish with `lib`, it seems inconsistent compared to eg. libpng

dtool/Package.cmake
```diff
@@ -0,0 +1,703 @@
...
+    set(PYTHON_LIBRARY "")
+    set(PYTHON_LIBRARIES "")
+
+    if(APPLE)
+      # macOS requires this explicit flag on the linker command line to allow the
+      # references to the Python symbols to resolve at dynamic link time
+      string(APPEND CMAKE_MODULE_LINKER_FLAGS " -undefined dynamic_lookup")
+      # TODO: p3dcparser contains some direct Python references; get rid of
+      # this once that's gone
+      string(APPEND CMAKE_SHARED_LINKER_FLAGS " -undefined dynamic_lookup")
```
rdb: Can this be removed now that we cleaned up the Python nonsense from p3dcparser?

dtool/Package.cmake
```diff
@@ -0,0 +1,703 @@
...
+    set(_ARCH_DIR "")
+
+  endif()
+
+  execute_process(
+    COMMAND ${PYTHON_EXECUTABLE}
+      -c "from sysconfig import get_config_var as g; print((g('EXT_SUFFIX') or g('SO'))[:])"
+    OUTPUT_VARIABLE _EXT_SUFFIX
+    ERROR_QUIET
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
```
rdb: We should probably not call this if we're making a cross build.

dtool/Package.cmake
```diff
@@ -0,0 +1,703 @@
...
+
+
+#
+# ------------ Image formats ------------
+#
+
+# JPEG
+find_package(JPEG QUIET)
+
+package_option(JPEG DEFAULT ON "Enable support for loading .jpg images.")
```
rdb: Since we support `stb_image` fallback for reading, we could remove the default-on status from PNG and JPEG—thoughts?

dtool/Package.cmake
```diff
@@ -0,0 +1,703 @@
...
+
+#
+# ------------ Math libraries ------------
+#
+
+# Eigen
+find_package(Eigen3 QUIET)
+
+package_option(EIGEN
+  "Enables experimental support for the Eigen linear algebra library.
```
rdb: This isn't really experimental any more

panda/src/configfiles/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+if(WIN32)
+  set(_confdir "etc")
+
+else()
+  # On Unices, ask the GNUInstallDirs module where /etc is
+  set(_confdir "${CMAKE_INSTALL_FULL_SYSCONFDIR}")
+
+endif()
+
+install(FILES "${CMAKE_BINARY_DIR}/etc/20_panda.prc" COMPONENT Core DESTINATION ${_confdir})
```
rdb: I am fine with changing the prc files to this scheme, however, if we make this step I think we should simultaneously also install the etc files to `/etc/panda3d` instead of `/etc` (and start respecting the XDG specification).

We can probably also better split them up, eg. the egg specific stuff should probably be in its own file.

pandatool/src/miscprogs/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_executable(bin2c binToC.cxx binToC.h)
+target_link_libraries(bin2c p3progbase)
+install(TARGETS bin2c EXPORT Tools COMPONENT Tools DESTINATION ${CMAKE_INSTALL_BINDIR})
```
rdb: Let's not build (or install) bin2c; we stopped doing this in makepanda because it conflicted with bin2c provided by other packages.  It doesn't seem like a Panda-specific tool anyway.

rdb:
Let's talk a bit about build types and the UX of them—is Release what is currently matched up to makepanda's `--optimize 3``, ie. neither _DEBUG nor NDEBUG is set, whereas Distribution matches optimize level 4 (`NDEBUG`, no docstrings, no bounds checks, etc.) in makepanda?

I'd like users to get a set of options that makes it obvious which option to pick for which purpose, and makes them gravitate to the right choice by default; upon further reflection, the distinction between Release and Distribution doesn't seem quite obvious.  I'm almost inclined to say that Release should correspond to optimize level 4, and we instead use an option called Normal/Standard/Development or something like that, and set that to be the default (if we can).  Any thoughts on this?  It really makes a big difference whether people use the right configuration for the right purpose, so I think we should take the time to think about this.

Furthermore, do we have/need a super-debug level that sacrifices any and all performance for the sake of thoroughly checking every thread operation, etc?

Also, one thing I'm still missing from the CMake branch is the compiler options that makepanda specifies, such as `-fno-strict-aliasing -ffast-math -fno-stack-protector -fno-finite-math-only -fno-unsafe-math-optimizations`; some of these are fairly critical.

rdb:
Also, I just noticed `panda3d/dtoolconfig.py` is missing; it's an alias for `panda3d.interrogatedb` for backward compatibility reasons.  However, I don't feel so strongly about this one if we decide that it has been deprecated for long enough.

pandatool/src/miscprogs/CMakeLists.txt
CFSworks: Should we be deleting binToC.cxx? I'd rather build the file (but perhaps not install it) if it exists. If we're straight-up not building it, we should delete it so we don't even bother maintaining it.

panda/src/configfiles/CMakeLists.txt
CFSworks: I like all points here. Installing to `/etc/panda3d` sounds great - would the search path for that be `<auto>/etc/panda3d`?

What do we use as the directory on Windows? What path do they reside under when running Panda from the build directory (i.e. not installed)?

dtool/Package.cmake
CFSworks: Oh definitely. Default-on is meant to communicate "your user experience will be severely lacking if you don't have this; please consciously opt out of it if you don't have the package"  but if it's not necessary for the samples to work, we shouldn't be requiring it so heavily.

I never knew `stb_image` was a PNG/JPEG fallback - cool!

dtool/Package.cmake
CFSworks: In a cross build, the Python executable won't be found and this won't get called. (This also works if you don't have a Python executable at all, and you only provide the headers.)

cmake/install/Panda3DConfig.cmake
```diff
@@ -0,0 +1,153 @@
...
+#
+#
+#   ODE       - Support for the ODE physics engine.
+#
+#               Panda3D::ODE::p3ode
+#
+#
+#   FFmpeg    - Support for FFmpeg media format loading.
+#
+#               Panda3D::FFmpeg::p3ffmpeg
```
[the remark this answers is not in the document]
CFSworks: When doing a fully-static build, they're static libraries rather than modules, and linking against them is necessary to use them.

The `FFmpeg` (and other modules') component should _probably not_ be available if Panda was not built statically though.

cmake/macros/BuildMetalib.cmake
```diff
@@ -0,0 +1,426 @@
...
+
+    else()
+      list(APPEND sources "${source}")
+
+    endif()
+  endforeach()
+
+  if(BUILD_METALIBS)
+    # CMake 3.0.2 doesn't like .I/.N/.T files!  We let it know that they're only
+    # headers.
```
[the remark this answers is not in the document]
CFSworks: CMake treats headers as things that are part of the project (and will appear in e.g. an MSVC dropdown) but are not passed to the compiler. Interrogate does indeed get the .N files passed (as well as all headers).

cmake/modules/FindTar.cmake
```diff
@@ -0,0 +1,39 @@
+# Filename: FindTar.cmake
```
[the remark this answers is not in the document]
CFSworks: Given the stupidly unfriendly name to be installing to a user's global command namespace, yes please, let's remove patching support.

If someone's interested in patching we should instead guide them toward either bsdiff or VCDIFF. Panda supporting binary patching seems like more Imagineering NIH syndrome cruft.

cmake/scripts/ConcatenateToCXX.cmake
```diff
@@ -0,0 +1,70 @@
...
+      break()
+    endif()
+
+    # Count the bytes we're adding
+    string(LENGTH "${data}" strlen)
+    math(EXPR byte_count "${byte_count} + (${strlen} / 2)")
+
+    # Format runs of up to 32 hex chars by indenting and giving a newline
+    string(REGEX REPLACE
+      "(...?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?.?)" " \\1\n"
```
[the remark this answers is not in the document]
CFSworks: I did try that initially. Sadly, [CMake's regex support is scant](https://cmake.org/cmake/help/v3.4/command/string.html#regex-specification) and this was the best I could do. I could perhaps use a for loop to generate the `.?` repetitions if that would help?

cmake/templates/win32_python/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+    relpath = os.path.relpath(path, __path__[-1])
+    dll_path = os.path.abspath(os.path.join(__path__[-1], '../bin', relpath))
+    if not os.path.isdir(dll_path):
+        return
+
+    os_path = os.environ.get('PATH', '')
+    os_path = os_path.split(os.pathsep) if os_path else []
+    os_path.insert(0, dll_path)
+    os.environ['PATH'] = os.pathsep.join(os_path)
```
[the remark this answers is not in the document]
CFSworks: The function here is to ensure that `bin` is in the search path for DLLs. Does Python 3 somehow prevent the DLL loader from seeing changes to the environment? At what version should I begin to use `os.add_dll_directory` instead of touching PATH?

dtool/CompilerFlags.cmake
```diff
@@ -0,0 +1,170 @@
...
+
+# Panda3D is now a C++11 project. Newer versions of CMake support this out of
+# the box; for older versions we take a shot in the dark:
+if(CMAKE_VERSION VERSION_LESS "3.1")
+  check_cxx_compiler_flag("-std=gnu++11" COMPILER_SUPPORTS_CXX11)
+  if(COMPILER_SUPPORTS_CXX11)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
+  else()
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++0x")
+  endif()
```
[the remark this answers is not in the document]
CFSworks: We definitely don't support MSVC on CMake 3.0.2 then.

We should make a checklist of every OS/distro/release we target out-of-the-box and, if none of them are still shipping 3.0.2, just bump the minimum to 3.1 (which doesn't expect you to play flag games to get C++11 support).

dtool/CompilerFlags.cmake
```diff
@@ -0,0 +1,170 @@
...
+    set(cxx_exceptions_on "-fexceptions")
+    set(cxx_exceptions_off "-fno-exceptions")
+
+  else()
+    set(cxx_exceptions_on)
+    set(cxx_exceptions_off)
+
+  endif()
+
+endif()
```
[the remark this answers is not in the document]
CFSworks: The style I chose for most of the CMake files is to have no empty line after a block opener but one empty line before closing that block iff that block contains 2+ lines. It does look like there are some cases in `CompilerFlags.cmake` that don't follow that style, but it's used in most other places.

We can change the style if it's unsightly, although I found the spacing to help my own eyes tremendously when I first introduced it.

dtool/Config.cmake
```diff
@@ -0,0 +1,629 @@
...
+    set(CMAKE_BUILD_TYPE RelWithDebInfo)
+  endif()
+  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+    Release RelWithDebInfo Debug MinSizeRel Distribution)
+endif()
+
+# Provide convenient boolean expression based on build type
+if(CMAKE_BUILD_TYPE MATCHES "Debug")
+  set(IS_DEBUG_BUILD True)
+  set(IS_NOT_DEBUG_BUILD False)
```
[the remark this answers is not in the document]
CFSworks: ON/OFF is what I usually use. This is actually old stuff that I thought was phased out (we should not be checking `CMAKE_BUILD_TYPE` at all; it may not even be known at configuration time)

dtool/Config.cmake
CFSworks: Good because that isn't tested at all.

dtool/Package.cmake
CFSworks: I figure after makepanda dies, the thirdparty directory should be structured in a way that's more CMake-specific (unless thirdparty zips go down with the ship). This bit is just meant to help people hit the ground running in the transition to CMake; it'll definitely evolve in the near future.

dtool/Package.cmake
CFSworks: This is to [avoid conflicts with the other Squish](https://cmake.org/cmake/help/v3.4/module/FindSquish.html). I don't like it either.

CMakeLists.txt
```diff
@@ -0,0 +1,141 @@
...
+set(CMAKE_DISABLE_IN_SOURCE_BUILD ON) # Must go before project() below
+
+if(CMAKE_VERSION VERSION_GREATER "3.12" OR POLICY CMP0074)
+  # Needed for THIRDPARTY_DIRECTORY support; this will be enabled by default
+  # once the minimum CMake version is at least 3.12.
+  cmake_policy(SET CMP0074 NEW)
+endif()
+
+# Figure out the version
+file(STRINGS "setup.cfg" _version REGEX "^version = ")
```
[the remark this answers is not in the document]
CFSworks: Best I can do is `[\\t ]*` since `\s` [isn't included in CMake's regex dialect](https://cmake.org/cmake/help/v3.4/command/string.html#regex-specification).

cmake/README.md
```diff
@@ -0,0 +1,52 @@
...
+```sh
+mkdir build && cd build
+cmake ..
+make
+[sudo] make install
+```
+
+To configure CMake, it is recommended to use cmake-gui (`cmake-gui .`),
+however it is also possible to configure it entirely through CMake's
+command-line interface; see `man cmake` for more details.
```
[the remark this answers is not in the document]
CFSworks: I'll probably postpone this particular concern. Some kind of instruction on how to use the new CMake buildsystem is extremely important; however it might belong in the root `README.md` instead, and we should hold off until we ourselves know what the best practices are before we tell people what new habits to form.

To that end I'm considering just deleting this file. Thoughts?

cmake/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+Panda subpackage building is handled by:
+```
+	BUILD_<SUBPACKAGE>=True/False   # Example: BUILD_DTOOL, BUILD_PANDA
+```
+Other configuration settings use their historical names (same names as in-source):
+```
+	# Examples
+	PANDA_DISTRIBUTOR="MyDistributor"
+	LINMATH_ALIGN=On
```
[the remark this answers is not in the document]
CFSworks: We should, but see above that I'm probably going to kill this file.

cmake/install/Panda3DConfig.cmake
```diff
@@ -0,0 +1,153 @@
...
+#
+#   DX9       - Support for Direct3D 9 rendering.
+#
+#               Panda3D::DX9::pandadx9
+#
+#
+#   GLES      - Support for OpenGL ES rendering.
+#
+#               Panda3D::GLES::pandagles
+#               Panda3D::GLES::pandagles2
```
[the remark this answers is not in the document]
CFSworks: Components should correspond mostly 1:1 to checkboxes in the NSIS installer. I could see someone wanting desktop OpenGL but not GLES, so it makes sense not to put them under `OpenGL`. However you may have a good point that they should be split up - I grouped them because they're so heavily grouped in the codebase.

Actually, perhaps it's best if you take a crack at the components list from a mile up: if these are options for what can/can't be installed independently, do they make sense? What would be your ideal component listing (off the top of your head, without even looking at code)?

cmake/macros/Interrogate.cmake
```diff
@@ -0,0 +1,388 @@
...
+  list(APPEND IGATE_FLAGS -D_X86_ -D__STDC__=1 -DWIN32_VC -D "_declspec(param)=" -D "__declspec(param)=" -D_near -D_far -D__near -D__far -D_WIN32 -D__stdcall -DWIN32)
+endif()
+if(MSVC_VERSION)
+  list(APPEND IGATE_FLAGS "-D_MSC_VER=${MSVC_VERSION}")
+endif()
+if(INTERROGATE_VERBOSE)
+  list(APPEND IGATE_FLAGS "-v")
+endif()
+
+set(IMOD_FLAGS -python-native)
```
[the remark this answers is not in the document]
CFSworks: I would love to tackle that. I'd like most of the team to be familiar with how `Interrogate.cmake` works for the time being before I change it so I can get good feedback on how to expand to other interpreters/languages, so I think we should put a pin in this and come back to it.

CFSworks:
> Also, one thing I'm still missing from the CMake branch is the compiler options that makepanda specifies, such as `-fno-strict-aliasing -ffast-math -fno-stack-protector -fno-finite-math-only -fno-unsafe-math-optimizations`; some of these are fairly critical.

Thanks for the flags! Is it as easy as pasting this exact string in the `CompilerFlags.cmake`? Do you have the MSVC equivalents handy as well? Are these the same across all build types?

As for build configurations: This is a wide-open field and we get to choose exactly how we set it up and I'm glad we're discussing this.

Should we stick to the 4 CMake core build types and assign them useful roles, consistent with what CMake users expect? For reference these are:
- `Release`: I am building this to ship to my users.
- `MinSizeRel`: I am building this to ship to my users, but only want strictly necessary features
- `RelWithDebInfo`: I want a release configuration build, but I don't trust it to be bug-free. I would like debugging symbols. (I probably also would like most of the less-expensive asserts.)
- `Debug`: I am a developer. Save me from my own mistakes.

The thing that makes my head spin about Panda is that there are really two types of "release":
- an SDK release, which may be downloaded and installed by someone who then makes a bunch of mistakes in Python land (so sanity-checking asserts on Python-facing functions should be intact)
- an end-user (i.e. player) release, where someone (probably reasonably more comfortable compiling and shipping software) wants every assert stripped out so everything runs the fastest possible

To that end I have no idea what `Distribution` means. Is this our SDK configuration or our end-user configuration? Which one is `Release`?

.travis.yml
```diff
@@ -22,30 +32,85 @@ addons:
...
+- pip install pytest
+
+before_script:
+- mkdir built
+- cd built
+
 script:
-    - $PYTHONV makepanda/makepanda.py --everything --git-commit $TRAVIS_COMMIT $FLAGS --threads 4
-    - test -n "$SKIP_TESTS" || LD_LIBRARY_PATH=built/lib PYTHONPATH=built $PYTHONV makepanda/test_imports.py
-    - test -n "$SKIP_TESTS" || LD_LIBRARY_PATH=built/lib PYTHONPATH=built $PYTHONV -m pytest -v tests
```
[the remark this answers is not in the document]
CFSworks: I'm actually in favor of throwing out the `.travis.yml` file and starting over. We need to have a broader discussion about what combinations of build settings make for a representative CI test.

pandatool/src/miscprogs/CMakeLists.txt
kamgha: I'd like this being switchable (the default being off), like pview. This tool is useful for compiling in bams (e. g. default font), while the source files can be kept in an editable format and be converted to the then current bam version.

cmake/README.md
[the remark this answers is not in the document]
rdb: That's fine.  I do think we need to acknowledge the CMake stuff in some file somewhere, maybe just put a note in README.md that it's possible to experimentally build with CMake but that documentation is forthcoming.

cmake/install/Panda3DConfig.cmake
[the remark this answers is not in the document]
rdb: Off the top of my head?  One for every `panda3d.*` module, and then one for every plug-in.  So, `Core`, `Direct`, `Vision`, `Assimp`, `FFMpeg`, `OpenGL`, `OpenGLES2`, `OpenAL`, etc.

Upon further thought I think I could go either way on splitting up GLES.  They're somewhat related in that a target platform probably supports both.  But, one will probably choose one or the other.  (I could actually see us deprecating OpenGL ES 1 at some point in the future.)

panda/src/configfiles/CMakeLists.txt
rdb: I think for system-wide installations, we search `$XDG_CONFIG_HOME/panda3d` and `$XDG_CONFIG_DIRS/panda3d` without going up the filesystem specifically.  Although, looking at the [spec](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html), the latter appears to be `/etc/xdg` by default, which I'm not 100% sure people are familiar with.

On Windows and non-systemwide installations we can use the current behaviour of having it under an `etc` directory in the build directory (or under `panda3d/etc` for a pip installation).

pandatool/src/miscprogs/CMakeLists.txt
rdb: We could build it but not install it.  Then people who compile Panda from source (the kind who are likely to compile-in default fonts) can still use it from the build directory.

However, are there really no third-party tools that can do the same thing?  It seems like a very generic thing.  I'd even be open to just ripping this out into its own repository and putting it up on PyPI so those who need it can `pip install bin2c` or something.

CFSworks: Is the question whether something like this is needed at build time? Because the `ConcatenateToCXX.cmake` script that I whipped up for adding the Interrogate preamble does all this and more. If we want to keep default fonts in non-C format (ttf and bam) the concatenate scrpit would work for those too and we can remove bin2c.

rdb:
I should note we have two kinds of developers: the ones that are working with Python, want useful asserts and don't want Panda to crash if they pass in the wrong parameters to a function, and the ones who actually want to debug Panda's internals.

We build our public SDKs particularly optimized for the former; noting that because it's the public SDK, people are going to expect reasonable performance from it, so it should still be reasonably efficient.  So it's actually a Release build with asserts (we define neither `_DEBUG` nor `NDEBUG`).  It doesn't sound like any of the existing CMake options; maybe `RelWithDebInfo` comes close except that it doesn't actually have debugging symbols, which would make the SDK builds massive.

I also am not clear what the difference is between "Distribution" and "Release" from how it is worded, hence my suggestion to have a "Standard" build type for the public SDKs / Python developers, so we can use "Release" for the things that actually go out to end-users.  The advantage of that is also that C++ people would reasonably expect "Release" to define `NDEBUG`, which would be preserved.

As for the flags, this is the bit in makepanda we should copy:
https://github.com/panda3d/panda3d/blob/35135ea0e1d9cd8bc24bec9a60fab259baeba740/makepanda/makepanda.py#L1428-L1438
Furthermore, we should define `-fno-rtti` in Release builds (and on Android we define it for all builds).
I'm not sure that we define particularly important flags for MSVC except `/DWINVER=0x501 /Zc:threadSafeInit- /D_USING_V110_SDK71_` to preserve Windows XP compatibility and `/GR-` to disable RTTI.

pandatool/src/miscprogs/CMakeLists.txt
rdb: Oo, I like the idea of being able to specify a font file to CMake and having it automatically implement the bin2c step using a CMake script.

dtool/Package.cmake
rdb: Is CMake clever enough not to try to find things in the system when doing a cross build?  Then it's fine, I suppose (though I assume there's still a way to get at a host Python executable for running scripts like makewheel or so).

dtool/Package.cmake
rdb: Yeah; stb_image lacks support for writing images, and for some of the fancier features, but it does fine in the vast majority of use cases.

dtool/Package.cmake
CFSworks: The appropriate thing to do when cross-compiling using CMake is to define a "toolchain file" which contains info about where to find the cross buildroot, what special compile flags are needed, what compiler to use, and such a file usually contains this line:
`set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)`

...that causes CMake to look at the host when searching for programs (like Python; though I don't know if FindPython guarantees the interpreter is part of the same installation as the library, in which case it would not find it)

If you're feeling particularly hardcore you can use QEMU + binfmts on Linux to make non-native executables work as if they were native, then you can use `ONLY` instead of `NEVER` in the above. But I do want to make sure it works without having to do that.

.travis.yml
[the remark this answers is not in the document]
rdb: That's fine, but unless you want to conclude that discussion before we merge this, let's at least keep the current makepanda matrix plus at *least* one configuration that tests CMake.

cmake/install/Panda3DConfig.cmake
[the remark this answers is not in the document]
rdb: Oh right, I forgot about static linking of plug-ins.  Fair enough.  I agree on making it unavailable in dynamic builds, but it's probably not a high priority.

cmake/modules/FindTar.cmake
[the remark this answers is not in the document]
rdb: image
aef81aceab51dee73b94dd48a1a30190b6d432d4 8b8796e9df6a302b7ec076b51e326f7f78c17560
Please do the needful and remove the tar stuff from CMake.  I haven't yet removed the Patchfile class since it's a public API, but we should definitely deprecate it.

cmake/templates/win32_python/__init__.py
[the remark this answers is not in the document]
rdb: That would be Python 3.8 (you can simply check whether that function exists).  See the `panda3d/__init__.py` file that makepanda generates on Windows (I posted an excerpt of the generation logic in another comment here), which has proven to work so I see no reason to replace it with new logic.

dtool/CompilerFlags.cmake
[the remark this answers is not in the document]
rdb: I am not about to start bikeshedding about our favourite styles, it just seemed like it might have been unintentional at first glance.  We can resolve this discussion without action as far as I'm concerned.

dtool/Config.cmake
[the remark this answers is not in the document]
rdb: In that case, if it's obsolete, it should be removed with prejudice.

dtool/CompilerFlags.cmake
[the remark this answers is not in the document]
CFSworks: Fair enough; was just trying to emphasize the arbitrariness of my choice and would not oppose a change in style before this lands (but I would want all other discussions to reach a conclusion before that, if we do).

I'll mark this resolved.

dtool/Package.cmake
rdb: Fine with me

dtool/Package.cmake
rdb: I might prefer a check for this just in case the user *hasn't* set this, but I don't feel strongly and you can resolve this as you please.

CFSworks: I'll choose to wait and see if anyone's bitten by forgetting to set it before adding a check myself. I feel like the users who cross-compile will either use an existing toolchain file or (if they're setting up the environment themselves) they know what they're getting into and can debug it if they make a mistake.

But if someone does complain, we can revisit.

rdb:
Another consideration, by the way, is that we currently promise that third-party libraries compiled against an SDK Panda build (with optimize level 3) can still be used with Panda binaries that are built with optimize level 4—this is necessary to have third-party C++ extension modules linked against the Panda API still work when the optimized Panda3D wheels are used during deployment.  (Though, this is not actually currently working due to a bug, which I shall fix shortly).

We should preserve this in CMake by reviewing the flags and ensuring that nothing that breaks the ABI in such a manner (such as DEBUG_THREADS) is being set in a build with default configuration, whether that ends up being Debug or something else.

We should probably add some automated tests for this at some point, but that would not be a blocker.

CFSworks:
Okay: `Distribution` renamed to `Standard`. The intended purpose of `Standard` is that it's what we cut our SDK installers from. `Release` hence means "Release _to end users specifically_" (i.e. is what I will be using most heavily)

I got most of those flags in. I'm waiting to see what the deal is with Windows XP support post-3.4 before worrying too much about supporting XP in the CMake buildsystem. It shouldn't be hard to add, but I don't want to add it and pull it immediately back out if XP is on its last legs.

So it looks like the remaining outstanding issues are:
- [ ] Test ABi compatibility for downstream Interrogate/etc. extensions to Panda across `Standard`/`Release` configurations
- [ ] Note in main README.md
- [ ] Get the CI situation sorted out
- [ ] Adjust Unix PRC install path to `/etc/panda3d` and make sure all of the reasonable defaults a la makepanda are addressed in the autoinstalled PRC(s).

Unless I'm missing anything these look like fairly low-impact bullet points, which is great news for me.

dtool/Config.cmake
```diff
@@ -36,15 +36,15 @@ endif()
...
+  set(CMAKE_CONFIGURATION_TYPES Release RelWithDebInfo Debug MinSizeRel Standard)
 else()
   # CMAKE_BUILD_TYPE can't just be set using the usual set(CACHE) method since
   # it's an empty string by default.
   if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE RelWithDebInfo)
   endif()
   set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
-    Release RelWithDebInfo Debug MinSizeRel Distribution)
+    Release RelWithDebInfo Debug MinSizeRel Standard)
```
rdb: Let's put Standard first in the list, and make it the default as opposed to RelWithDebInfo.  Standard corresponds to what is currently the makepanda default, and what we want the majority of developers building Panda from source for the first time will want to use.

CFSworks:
> Another consideration, by the way, is that we currently promise that third-party libraries compiled against an SDK Panda build (with optimize level 3) can still be used with Panda binaries that are built with optimize level 4

> makepanda's --optimize 3, ie. neither _DEBUG nor NDEBUG is set, ... optimize level 4 (NDEBUG, no docstrings, no bounds checks, etc.) in makepanda

From this I gather that there's supposed to be ABI compatibility between `NDEBUG` being set/not set, but while trying to verify this, I've found [this in PandaNode](https://github.com/panda3d/panda3d/blob/7790f84/panda/src/pgraph/pandaNode.h#L535) which causes `NDEBUG` to make `PandaNode` 8 bytes smaller when enabled.

How should we proceed? Do we add `NDEBUG` to `Standard`, do we drop the guarantee of ABI compatibility across `Release`/`Standard`, or do we fix all cases of ABI depending on `NDEBUG` (and add a CI job to run tests with the build types mixed to guard against regressions)?

rdb:
@CFSworks I'm aware that there are parts where the ABI still differs from NDEBUG.  (Not sure how this particularly egregious case has been missed.)  These cases should be fixed.  I will remove the NDEBUG guard.  Automating ABI testing has been on my mind for a long time, but I've never gotten around to doing it.

I suppose we're not really selling it as a "guarantee" as long as we don't actually test that it's the case, but the reason why this ABI compatibility is important is so that people can build their own third-party extensions and still use our deployment system, which downloads a Release version of the Panda binaries.

Furthermore, the reason why Standard shouldn't have NDEBUG is because NDEBUG removes a lot of things that people really do expect from a "Standard" build of Panda3D.  Panda will often simply crash if you pass an out-of-bounds value to a function.  This is not acceptable behaviour for Python users, who have no way to debug segmentation faults.

dtool/Config.cmake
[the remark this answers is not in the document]
rdb: This one seems unresolved as of yet.

PR #733, opened 2019-09-07
build_apps: Create desktop entries for Linux binaries

Moguri:
This includes icons. I am unsure of a great way to deal with relative versus absolute paths other than just letting packagers deal with it (see PKGBUILD script below). I am also not sure how to specify multiple icon sizes, so I just picked the largest one for now. We should also investigate allowing SVG files.

Generated .desktop files pass desktop-file-validate, but I ran into difficulties actually testing icons with i3 since none of the menus I found displayed icons. I used the following PKGBUILD script to test out installing the .desktop file:

```bash
pkgname=asteroids-bin
pkgver=0.0.0
pkgrel=1
pkgdesc=""
arch=('x86_64')
license=('custom:commercial')
provides=('asteroids-bin')
options=('!strip')
source=("asteroids-${pkgver}_manylinux1_x86_64.tar.gz")
md5sums=('SKIP')

prepare() {
    sed -i 's/Exec.*/Exec=\/opt\/asteroids\/asteroids/' "${srcdir}/asteroids/asteroids.desktop"
    sed -i 's/Icon.*/Icon=\/opt\/asteroids\/asteroids.png/' "${srcdir}/asteroids/asteroids.desktop"
}

package() {
    pushd "${srcdir}/asteroids/"
    find . -type f -exec install -Dm755 {} "${pkgdir}"/opt/asteroids/{} \;
    popd
    install -Dm644 "${srcdir}/asteroids/asteroids.desktop" "${pkgdir}/usr/share/applications/asteroids.desktop"
}
```

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -617,6 +617,35 @@ class build_apps(setuptools.Command):
             if filename.endswith('.whl')
         ]
 
+    def create_desktop_entry(self, builddir, appname, use_terminal):
+        """Create .desktop file for Linux"""
+        fname = os.path.join(builddir, '{}.desktop'.format(appname))
+
+        iconobj = self.icon_objects.get(
+            appname,
+            self.icon_objects.get('*', None)
+        )
+        iconname = None
+
+        if iconobj is not None:
+            # Just grab the largest icon for now
+            largest_size = sorted(iconobj.images.keys(), reverse=True)[0]
+            iconpnmi = iconobj.images[largest_size]
+            iconname = '{}.png'.format(appname)
+            icondst = os.path.join(builddir, iconname)
+            iconpnmi.write(icondst)
+
+        with open(fname, 'w') as desktop_file:
+            desktop_file.write('[Desktop Entry]\n')
+            desktop_file.write('Version=1.0\n')
+            desktop_file.write('Type=Application\n')
+            desktop_file.write('Encoding=UTF-8\n')
+            desktop_file.write('Name={}\n'.format(appname.title()))
+            desktop_file.write('Terminal={}\n'.format('true' if use_terminal else 'false'))
+            desktop_file.write('Exec={}\n'.format(appname))
+            if iconobj is not None:
+                desktop_file.write('Icon={}'.format(appname))
+
     def update_pe_resources(self, appname, runtime):
         """Update resources (e.g., icons) in windows PE file"""
 
@@ -1028,6 +1057,10 @@ class build_apps(setuptools.Command):
             for suffix in freezer.mf.suffixes:
                 if suffix[2] == imp.C_EXTENSION:
                     ext_suffixes.add(suffix[0])
+
+            # Create a desktop entry for Linux
+            if 'linux' in platform:
+                self.create_desktop_entry(builddir, appname, use_console)
 
         for appname, scriptname in self.gui_apps.items():
             create_runtime(platform, appname, scriptname, False)


Moguri:
@rdb What can we do to move this along? Are there some tests you would like to see?

ArsThaumaturgis:
> Just wondering, what is the effect of having the .desktop file? What does it do? Is the .desktop file clickable in some file managers?

I'm no expert, but my understanding is that a "desktop file" is essentially the Linux equivalent of a launcher shortcut: double click on it and the indicated program is launched. It also allows the specification of an icon, as well as a variety of bits of metadata.

el-dee:
Also, application can register their .desktop to the Window Manager using _NET_WM_DESKTOP_FILE hint (Mainly Ubuntu Unity), the WM will then use the information found in te file to display the app icon and tailor the window appearance

rdb:
Can someone test that this PR is actually having this effect?

Should we be setting _NET_WM_DESKTOP_FILE?

Moguri:
This is not completely untested. I was able to have both rofi and dmenu pick up on the desktop file. However, neither of these display an icon. I inspected the resulting desktop file and compared it to one from Blender, and the icon stuff looks right, but I agree that having someone confirm it's working would be good.

I have not heard about the _NET_WM_DESKTOP_FILE stuff before; I just placed the desktop file in `/usr/share/applications`.

I'll fix the whitespace and debug print issues. With such a small patch, you'd think I would have noticed that...

ArsThaumaturgis:
What would be called for to test it? Would it be enough to take a desktop file and sample program generated by it, place it in the correct location, and see whether it shows an icon and runs the program? Or does this it building this branch and using it to generate a desktop file and distributable on the test-machine?

I have Ubuntu 18.04.5, and should thus be able to run a test or two.

Moguri:
I have updated the branch (rebased onto current `master`, removed the extra whitespace, and removed the debug print).

A desktop file generated with this patch should work with the `build_apps` results from the current `master`. It will likely also work with 1.10.6. However, the asteroids sample doesn't have an icon, so I'll need to set something up for testing.

Moguri:
The failing test seems unrelated to this patch:
```
_________________________ test_filename_ctor_pathlib __________________________

    def test_filename_ctor_pathlib():
        pathlib = pytest.importorskip('pathlib')
    
        path = pathlib.Path(__file__)
        fn = Filename(path)
>       assert fn.to_os_specific_w() == str(path)
E       AssertionError: assert 'D:\\a\\panda...t_filename.py' == 'd:\\a\\panda...t_filename.py'
E         - d:\a\panda3d\panda3d\tests\dtoolutil\test_filename.py
E         ? ^
E         + D:\a\panda3d\panda3d\tests\dtoolutil\test_filename.py
E         ? ^

..\tests\dtoolutil\test_filename.py:23: AssertionError
```

rdb:
> Would it be enough to take a desktop file and sample program generated by it, place it in the correct location, and see whether it shows an icon and runs the program? Or does this it building this branch and using it to generate a desktop file and distributable on the test-machine?

Someone can just copy the modified commands.py file in this PR to their local copy of Panda, run build_apps on a game with an icon defined, and check that the .desktop file does indeed behave as advertised in the file manager (ie. it shows up with the full name of the game, double-clicking it launches the game, and it shows the icon).

ArsThaumaturgis:
Hmm... when I simply replace my "commands.py" (found in "/usr/local/lib/python3.6/dist-packages/direct/dist") and attempt to build a modified version of the Asteroids sample (using the command "python3 setup.py bdist_apps"), I get the following error:

```text
  File "/usr/local/lib/python3.6/dist-packages/direct/dist/commands.py", line 27, in <module>
    from .icon import Icon
ModuleNotFoundError: No module named 'direct.dist.icon'
```

Full error-text:

```text
  File "setup.py", line 22, in <module>
    "asteroids" : ["ship.png"
  File "/usr/lib/python3/dist-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python3.6/distutils/core.py", line 134, in setup
    ok = dist.parse_command_line()
  File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 499, in parse_command_line
    result = _Distribution.parse_command_line(self)
  File "/usr/lib/python3.6/distutils/dist.py", line 472, in parse_command_line
    args = self._parse_command_opts(parser, args)
  File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 814, in _parse_command_opts
    nargs = _Distribution._parse_command_opts(self, parser, args)
  File "/usr/lib/python3.6/distutils/dist.py", line 528, in _parse_command_opts
    cmd_class = self.get_command_class(command)
  File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 634, in get_command_class
    self.cmdclass[command] = cmdclass = ep.load()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2324, in load
    return self.resolve()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2330, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib/python3.6/dist-packages/direct/dist/commands.py", line 27, in <module>
    from .icon import Icon
ModuleNotFoundError: No module named 'direct.dist.icon'
```

Attempting to build with the original "commands.py" doesn't seem to produce this error.

Is there perhaps some other change in Moguri's branch that's not found in the standard distribution of Panda, and that these changes rely upon?

Moguri:
You'll need to modify a 1.11 build since it has the icon changes. You can grab a 1.11 build with:

`pip install --pre --extra-index-url https://archive.panda3d.org/branches/master panda3d`

ArsThaumaturgis:
All right, I've installed 1.11 using the above pip-command (well, with "pip3" instead of "pip" and with "sudo -H" to give the appropriate permissions), and replaced its "commands.py" with the new one, I believe.

Applied to the Asteroids example, this successfully produces a build (and the executable runs)--but the resultant desktop file doesn't seem to work on my system.

Specifically, here are the contents of the file as generated:
```
[Desktop Entry]
Version=1.0
Type=Application
Encoding=UTF-8
Name=Asteroids
Terminal=false
Exec=asteroids
Icon=asteroids
```

To start with, note that the value for the "icon" field lacks an extension (in this case ".png").

But more importantly, and perhaps more problematically, it looks like relative paths may not work--for icons or executables--in desktop files. (With some exceptions for certain directories and paths, I think--for example, it seems that executables located within the values "$PATH" should be picked up.)

And indeed, looking at a desktop file associated with another program that I have installed (Divinity: Original Sin), I do note that it uses full, absolute paths for all of its "icon", "exec", and "path" values. I'm guessing that these were derived and set by the installer, as they are, of course, specific to my machine.

See this (admittedly old) discussion for more:
https://askubuntu.com/questions/722179/icon-path-in-desktop-file

This looks to be the relevant spec. for Ubuntu:
https://developer.gnome.org/desktop-entry-spec/

Now, the discussion linked-to above is old, and I don't state with confidence that the spec. linked-to above is the one in use. What they say does seem to track on my system, but I may be missing something--this isn't a field in which I'm particularly learned. Thus take the above--save for the generated desktop file not working on my machine--with some salt.

Moguri:
Ah, right, I think I ran into simular issues with absolute paths, which is why I was using that PKGBUILD script in the original post. PKGBUILD is for creating packages for ArchLinux. It seems like it might not be uncommon to let packagers deal with this? For example, Blender does not seem to use absolute paths: https://developer.blender.org/diffusion/B/browse/master/release/freedesktop/blender.desktop.

ArsThaumaturgis:
> PKGBUILD is for creating packages for ArchLinux. It seems like it might not be uncommon to let packagers deal with this?

I'll confess that this is something that I'm unfamiliar with. I take it that a "packager" is something like an installer, or part thereof?



> EDIT: To add on to this, I suspect packagers/packages either install files to locations where "relative" paths works (e.g., into a location on the `PATH` like `/usr/bin`) or modify the desktop file to use an absolute path (as I did above).

I imagine that this is true indeed. As I mentioned in my post, at least one desktop file that I have that was installed with an application uses an absolute path, presumably generated as part of the installation process, which to some degree at least supports this idea.

I would suggest that it might be preferable to not install to system locations, as I imagine that the relative-path-enabled locations generally are, and thus to rather install to some user location and then modify the desktop file to use absolute paths.

Moguri:
A "packager" in this case is referring to someone who is preparing a package for a Linux distribution's package repo. For example, someone who creates deb files for Debian or Ubuntu.

ArsThaumaturgis:
Ah, I see--thank you for the explanation. So I take it that the script that you provided above might be incorporated into a deb-file, to be run on the end-user's system? Is it feasible for such deb-files to be generated by our build-system?

rdb:
Where are we with this PR?  From my understanding, the current status is that the .desktop files do not work when extracting the tarball, because they do not support relative paths (ie. they require the binaries to be added to $PATH), correct?

From the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html), could we solve this problem with the %k substitution?

Moguri:
The two main hold-ups are:
1. The current implementation does not support portable binaries and requires icons and the binary to be "installed"
2. Icons have not been confirmed to work (I do not have an application launcher/menu that shows icons)

It seems that the desktop file spec does not really support "portable" desktop files (e.g., via relative paths). The two issues I see with `%k` are:

1. This is a path to a file and not a directory, so we'd need to do some shell trickery to grab the directory
2. I do not believe this works for icons

I am inclined to leave the generated desktop file as assuming that the binary is on the `$PATH`, and that icons are similarly "installed" into an appropriate directory.

rdb:
I'm reluctant to ship a "broken" desktop file, ie. one that looks inviting to click on after extracting the tarball, but actually doesn't work.  It seems that desktop files are clearly meant to be used only when the binary is installed to the system, so I would be inclined to only ship the desktop file when we are producing an installer.

ArsThaumaturgis:
The above seems wise to me.

Issue #516, opened 2019-01-07
Remove deprecated p3d/AppRunner code


Moguri:
With the new deployment code in 1.10, we no longer need the old p3d/AppRunner code. I am aggregating a list of code that can be removed in this issue and will follow up with a Pull Request.

- [ ] direct/src/p3d
- [ ] direct/src/plugin*
- [ ] direct/src/showbase/AppRunnerGlobal.py

CFSworks:
A thousand times yes!

There's some code in `panda/src/iphone` that references runp3d and/or the plugin, and I believe `panda/src/display/subprocess*` also exists to facilitate the plugin. We might want to look into removing those as well.

rdb:
Ideally, we'd still have a way to run old .p3d files for compatibility; there's a lot of old .p3d files floating around the web that would become unplayable.  We could create a little runp3d script that installs Panda3D 1.9 into a virtualenv, mounts the .p3d to the VFS, registers the VFSImporter, and registers a shim AppRunner class.  It doesn't have very high priority for me personally, but maybe someone in the community could work on such a thing?

Moguri:
Maybe it's better to leave this stuff in and then remove it in a major version change?

rdb:
I think it's fine to remove this in master, and consider a possible compatibility p3d runner as a separate issue if someone feels it is needed.

Moguri:
Should the runtime and rtdist build options be removed as part of this?

rdb:
@Moguri Yes, they should.

Moguri:
This should have been closed after #734 was merged; closing now.

rdb:
I noticed that panda/src/dxml (containing TinyXML) isn't used by anything, because it was only used by the plug-in code.

Should we remove it?  There are plenty of alternative XML parsing libraries out there for people that depend on this.

PR #734, opened 2019-09-08
Remove p3d/AppRunner/plugin system

Moguri:
This is for #516 

Tested with:
  * Running test suite
  * Running `bdist_apps` on the asteroids sample with an icon defined


dtool/PandaVersion.pp
```diff
@@ -48,4 +48,4 @@
 // integers, but by convention it will consist of four integers, with
 // the first three matching the plugin version, and the fourth integer
 // being incremented with each new Core API revision.
-#define P3D_COREAPI_VERSION $[P3D_PLUGIN_VERSION] 2
+#define P3D_COREAPI_VERSION
```
rdb: What's the rationale for not removing these lines altogether?

I think the only thing using this is `ParsePluginVersion` in `makepandacore.py`, which is used by `makepanda/getversion.py --runtime`, but its `--runtime` handling should be removed, so then we can also remove `ParsePluginVersion` and `ParseCoreapiVersion` from `makepandacore.py`.

direct/src/showbase/AppRunnerGlobal.py
```diff
@@ -1,11 +0,0 @@
...
-object, which only exists when we are running a .p3d file via
-runp3d.py or via the Panda3D plugin or standalone executable.
-
-This is needed for apps that start themselves by importing
-DirectStart; it provides a place for these apps to look for
-the AppRunner at startup. """
-
-#: Contains the global AppRunner instance, or None if this application
-#: was not run from the runtime environment.
-appRunner = None
```
rdb: Could we leave this module in existence for now, but print a deprecation warning when trying to import it?

dtool/PandaVersion.pp
Moguri: I was thinking it was used by PandaSystem, but it's not so I'l remove them.

direct/src/showbase/AppRunnerGlobal.py
Moguri: Do we have a mechanism for doing deprecation, or should I just print a warning if this module is imported?

rdb: Let's just `print()` a warning, protected by `if __debug__:`

dtool/PandaVersion.pp
rdb: We can change the methods on PandaSystem to do whatever they do if the symbols aren't defined.  We should also mark those methods as deprecated.

Moguri: How do we deprecate C++ code?

rdb: At the very least put a `@deprecated` tag in the docstring like so:
```cpp
/**
 * Method that does something.
 *
 * @deprecated The runtime system has been removed, this now does nothing.
 */
void PandaSystem::
do_something() {
```

Adding a warning using `std::cerr << "Using deprecated PandaSystem.do_something()\n";` under an `#ifndef NDEBUG` would be appropriate if we want to discourage people even more strongly.  (The prc/logging stuff hasn't been defined yet in dtoolbase, so we can't rely on `nassert_raise` et al.)

dtool/PandaVersion.pp
```diff
@@ -31,21 +31,3 @@
 // build which you will be using to produce a distributable Panda3D
 // package, you should set this string appropriately.
 #define PANDA_PACKAGE_VERSION
```
rdb: This can also be removed

dtool/PandaVersion.pp
Moguri: I have the `@deprecated` tag in the docstrings. I can also add the `cerr` prints if you'd like.

Moguri:
Any other changes or testing you'd like to see?

rdb:
There are still references to the runtime in installpanda.  Also, the docstrings should be wrapped along the ruler per our coding style.  However, I'll fix these before merging.  Thanks!

PR #740, opened 2019-09-16
Use GitHub Actions for continuous integration

rdb:
This adds CI using the GitHub Actions beta for Ubuntu, macOS and Windows.  I didn't spend much effort on this, but it's a stub that we can use to replace Travis (and eventually, BuildBot); it's already better than Travis in that it can build on the major three platforms (and also supports Docker containers).  We can add CMake as a separate job.

The main downside appears to be that GitHub Actions doesn't have IRC or Discord notifiers yet, but this could be done using custom action steps (or by using `curl` to poke a webhook).

We can actually upload artifacts for a build, but this is not yet done.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,57 @@
+name: Continuous Integration
+on: [push, pull_request]
+jobs:
+  makepanda:
+    strategy:
+      matrix:
+        os: [ubuntu-16.04, windows-2016, macOS-10.14]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install dependencies (Ubuntu)
+      if: matrix.os == 'ubuntu-16.04'
+      run: |
+        sudo apt-get install build-essential bison flex libfreetype6-dev libgl1-mesa-dev libjpeg-dev libode-dev libopenal-dev libpng-dev libssl-dev libvorbis-dev libx11-dev libxcursor-dev libxrandr-dev nvidia-cg-toolkit zlib1g-dev
+    - name: Get thirdparty packages (Windows)
+      if: runner.os == 'Windows'
+      shell: powershell
+      run: |
+        $wc = New-Object System.Net.WebClient
+        $wc.DownloadFile("https://www.panda3d.org/download/panda3d-1.10.4.1/panda3d-1.10.4.1-tools-win64.zip", "thirdparty-tools.zip")
+        Expand-Archive -Path thirdparty-tools.zip
+        Move-Item -Path thirdparty-tools/panda3d-1.10.4.1/thirdparty -Destination .
+    - name: Get thirdparty packages (macOS)
+      if: runner.os == 'macOS'
+      run: |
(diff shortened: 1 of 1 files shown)


rdb:
Hmm, shame, it won't show the checks in this PR (presumably because the workflow hasn't been merged to master yet); but the run can be seen here:
https://github.com/rdb/panda3d/runs/224277491

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+      run: |
+        python -m pip install pytest
+        PYTHONPATH=built LD_LIBRARY_PATH=built/lib DYLD_LIBRARY_PATH=built/lib python -m pytest
+    - name: Set up Python 2.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 2.7
+    - name: Build Python 2.7
+      run: |
+        python makepanda/makepanda.py --no-copy-python --git-commit=${{github.sha}} --outputdir=built --everything --no-eigen --python-incdir=$pythonLocation/include --python-libdir=$pythonLocation/lib --verbose --threads=4
```
Moguri: Is this actually running with Python 2.7? I am probably just missing something, but I am not seeing anything where `python` would point to anything other than the system `python`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+        Move-Item -Path thirdparty-tools/panda3d-1.10.4.1/thirdparty -Destination .
+    - name: Get thirdparty packages (macOS)
+      if: runner.os == 'macOS'
+      run: |
+        curl -O https://www.panda3d.org/download/panda3d-1.10.4.1/panda3d-1.10.4.1-tools-mac.tar.gz
+        tar -xf panda3d-1.10.4.1-tools-mac.tar.gz
+        mv panda3d-1.10.4.1/thirdparty thirdparty
+        rmdir panda3d-1.10.4.1
+        (cd thirdparty/darwin-libs-a && rm -rf ode openal vrpn openexr assimp rocket)
+    - name: Set up Python 3.7
```
Moguri: I'm sure you're already thinking about this, but we'll want to setup a matrix of Python versions too.

.github/workflows/ci.yml
rdb: The setup-python action above is putting the correct Python version on the PATH.  The makepanda build log confirms that the correct version is chosen.

If I use `python2.7` explicitly it won't work on Windows, since version-specific symlinks don't exist there.

.github/workflows/ci.yml
rdb: The reason I haven't put the Python versions in a matrix is that by building each Python version sequentially, makepanda can reuse the majority of the build for the other Python version; ie. it doesn't need to rebuild the C++ libraries or command-line tools, only the Python bindings.  This makes a dramatic difference in build time.

Furthermore, when we want to build the installers, we actually want to ship multiple Python versions in the same .exe (perhaps this could alternatively be done using a dependent job or so; but this seemed easier for now).

It does mean we have a more bloated workflow file, of course, but I think it's worth it as CI builds should be fast.

.github/workflows/ci.yml
Moguri: Ah, I see. That makes sense. I was thinking the action just made sure it was available.

PR #742, opened 2019-09-20
Added DirectSpinBox

fireclawthefox:
New GUI widget spinbox for numeric user input
Added Mousewheel up and down button events to Direct GUI globals

### DIFF
--- a/direct/src/gui/DirectGuiGlobals.py
+++ b/direct/src/gui/DirectGuiGlobals.py
@@ -69,6 +69,8 @@ B3PRESS = PGButton.getPressPrefix() + MouseButton.three().getName() + '-'
 B1RELEASE = PGButton.getReleasePrefix() + MouseButton.one().getName() + '-'
 B2RELEASE = PGButton.getReleasePrefix() + MouseButton.two().getName() + '-'
 B3RELEASE = PGButton.getReleasePrefix() + MouseButton.three().getName() + '-'
+MWUP = PGButton.getPressPrefix() + MouseButton.wheel_up().getName() + '-'
+MWDOWN = PGButton.getPressPrefix() + MouseButton.wheel_down().getName() + '-'
 # For DirectEntry widgets
 OVERFLOW = PGEntry.getOverflowPrefix()
 ACCEPT = PGEntry.getAcceptPrefix() + KeyboardButton.enter().getName() + '-'
--- a/direct/src/gui/DirectSpinBox.py
+++ b/direct/src/gui/DirectSpinBox.py
@@ -1,0 +1,316 @@
+'''This module contains the DirectSpinBox class.
+
+A DirectSpinBox is a type of Entry specialized for number selection.
+It contains two buttons to its right for increasing and decreasing the
+numeric value given in the entry field.  Values can also be directly
+typed in the entry and will be checked against the type given to the
+spiner box which is *int* by default.
+
+Options available for this Widget are:
+
+*   value
+    The value selected by the spiner. This will always return the number
+    in the format given in the valueType option.  You can also pass a
+    string or other type which the valueType function is possible to
+    convert from.
+
+*   textFormat
+    This is a format string which will be used to write the value into
+    the entry field
+
+*   stepSize
+    When the up and down arrows are pressed, stepSize will be used to
+    determine how much the value should be in-/decreased
+
+*   minValue
+    The minimum value that can be entered in the spiner value. If values
+    given to the spiner are lower than the min value, the spiner will
+    set the minValue as new value.
+
+*   maxValue
+    The maximum value that can be entered in the spiner value. If values
+    given to the spiner are larger than the max value, the spiner will
+    set the maxValue as new value.
+
+*   valueType
+    A function like float, int or similar which will convert the value
+    given to the spiner.  It's return value if it was able to convert
+    the value correctly will then be converted using the textFormat to
+    a string which is entered into the textfield.
+
+*   repeatdelay
+    Delay in seconds at which the value should be in-/decreased when the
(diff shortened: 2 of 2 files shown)


direct/src/gui/DirectGuiGlobals.py
```diff
@@ -69,6 +69,8 @@
 B1RELEASE = PGButton.getReleasePrefix() + MouseButton.one().getName() + '-'
 B2RELEASE = PGButton.getReleasePrefix() + MouseButton.two().getName() + '-'
 B3RELEASE = PGButton.getReleasePrefix() + MouseButton.three().getName() + '-'
+MWUP = PGButton.getPressPrefix() + MouseButton.wheel_up().getName() + '-'
+MWDOWN = PGButton.getPressPrefix() + MouseButton.wheel_down().getName() + '-'
```
rdb: I think this might be too obscure of a name.  How about `WHEELUP`?

fireclawthefox: Sure, will add unit tests for it. And yeah, WHEELUP and WHEELDOWN are fine too. I just tried to keep it similar to the other event names in the first place.

tests/gui/test_DirectSpinBox.py
```diff
@@ -0,0 +1,117 @@
...
+        global test_value_a
+        test_value_a = extraArg
+
+    spinner = DirectSpinBox(command=func_a, extraArgs=[True])
+
+    messenger.send(DGG.ACCEPT + spinner.valueEntry.guiId, [None])
+
+    assert test_value_a
+
+pytest.main()
```
rdb: I think this line may be causing the tests to fail.

tests/gui/test_DirectSpinBox.py
```diff
@@ -0,0 +1,117 @@
...
+if sys.version_info >= (3, 0):
+    import builtins
+else:
+    import __builtin__ as builtins
+builtins.loader = Loader(base=None)
+
+from direct.task.TaskManagerGlobal import taskMgr
+builtins.taskMgr = taskMgr
+from direct.showbase.MessengerGlobal import messenger
+builtins.messenger = messenger
```
rdb: I would rather change DirectSpinBox to rely on as few builtins as possible.  This can be done by adding the following to `DirectSpinBox.py`:
```
from direct.showbase.MessengerGlobal import messenger
from direct.task.TaskManagerGlobal import taskMgr
```

I realise this may not be possible (yet) for Loader.

tests/gui/test_DirectSpinBox.py
fireclawthefox: oh, I missed to remove that line again :P

tests/gui/test_DirectSpinBox.py
fireclawthefox: Makes sense, will change the class according to your proposal

el-dee:
I've tested (and started to integrate) this into my app, and it works great!

The only problem is that the callbacks as they are now are almost useless, at least for my usage. Instead I modified the code to fire the callback in the setValue() method. By doing so any change of the value trigger the callback, whether it's a change of the value in the entry box, press or repeated press on a button or even external change. By doing so the DirectSpinBox behaves like the DirectSlider.

fireclawthefox:
> I've tested (and started to integrate) this into my app, and it works great!
> 
> The only problem is that the callbacks as they are now are almost useless, at least for my usage. Instead I modified the code to fire the callback in the setValue() method. By doing so any change of the value trigger the callback, whether it's a change of the value in the entry box, press or repeated press on a button or even external change. By doing so the DirectSpinBox behaves like the DirectSlider.

That's exactly what I noticed and had to add too some days ago and which I finally just got to commit :P

el-dee:
> That's exactly what I noticed and had to add too some days ago and which I finally just got to commit :P

Coincidence or not, thank you for the update :)

For my usage, I modified the callback call to also pass extraArgs as parameters (and also the actual value like DirectEntry), otherwise you need one callback per DirectSpinBox, which can quickly become bothersome when you have a lot of them :)

fireclawthefox:
> Coincidence or not, thank you for the update :)
> 
> For my usage, I modified the callback call to also pass extraArgs as parameters (and also the actual value like DirectEntry), otherwise you need one callback per DirectSpinBox, which can quickly become bothersome when you have a lot of them :)

Ah, yes, you're right, I've totally forgot to pass the extra args. In my case it didn't matter much as each SpinBox was encapsulated separately and hence there was no need to differentiate all the spinboxes.
I'll add the extra args and entered value to my version too, probably by tomorrow.

Moguri:
@fireclawthefox Are you still interested in pursuing this PR? If not, or if I don't here from you in two weeks, I intend to close this PR. It can always be re-opened.

el-dee:
Small bug I discovered recently, suppressKeys and suppressMouse are not propagated to the inner components (the entry and the buttons).

Simply adding this in the createcomponent() seems to do the trick : 

    suppressKeys = self['suppressKeys'],
    suppressMouse = self['suppressMouse']

fireclawthefox:
@el-dee I'm passing the suppress* options now as it seems to be quite convenient, though for other options I guess it'll be better to pass them as valueEntry_[optionName], incButton_[optionName] or decButton_[optionName]. So, similar to what you'd do in other directGui elements that have sub components.

@rdb, the name is changed to the one with the uppercase 'D' now as well as the font being loaded with the FontPool. Also extended the tests with the suppress* changes el-dee proposed. I think it's good enough for a new review.

Moguri:
@rdb looks like all previous review items have been addressed. Any further thoughts on this?

PR #748, opened 2019-09-30
MacOS: Wrong CFBundleIconFile in Info.plist

el-dee:
CFBundleIconFile should contain the actual icon file name : iconfile.icns, not iconfile. Otherwise the icon is not used for the app.

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -514,7 +514,7 @@ class build_apps(setuptools.Command):
             self.icon_objects.get('*', None)
         )
         if icon is not None:
-            plist['CFBundleIconFile'] = 'iconfile'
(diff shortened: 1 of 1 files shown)


rdb:
According to the [macOS documentation](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102043):
> `CFBundleIconFile` (`String` - iOS, macOS) identifies the file containing the icon for the bundle. The filename you specify does not need to include the extension, although it may.

In my own testing it also didn't make a difference.  Did your testing indicate otherwise?

el-dee:
I reproduced the bug, but actually the problem is in Finder which seems to sometimes miss the update of the application bundle and keep the old icon (or absence of icon) of the previous version of the app.

PR #751, opened 2019-10-02
dcparser: Fix a typo where an unpacked uint64 or int64 might shave off a few bits

TheRandomDog:
Just was rummaging through Panda3D source after I encountered an issue in my own project that ended up giving me different numbers after packing and unpacking a large uint64 value. I believe this is the culprit.

### DIFF
--- a/direct/src/dcparser/dcPackerInterface.I
+++ b/direct/src/dcparser/dcPackerInterface.I
@@ -253,7 +253,7 @@ do_unpack_int64(const char *buffer) {
                     ((uint64_t)(unsigned char)buffer[4] << 32) |
                     ((uint64_t)(unsigned char)buffer[5] << 40) |
                     ((uint64_t)(unsigned char)buffer[6] << 48) |
-                    ((int64_t)(signed char)buffer[7] << 54));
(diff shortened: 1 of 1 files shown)


rdb:
Are you willing to add a unit test for DCPacker that just packs and unpacks various values for uint64?

CFSworks:
@rdb The change itself is definitely correct; should I get the unit test?

rdb:
@CFSworks I'd appreciate your opinion on whether this could cause compatibility issues.  Anyone can do the unit test

CFSworks:
@rdb Highly unlikely. This only affects integers larger than 56-bit, and as-is the packing is mismatched with the unpacking code - if somebody else had encountered this, we'd have heard about it sooner. The users of this code are also only using it for networked purposes (afaik nobody serializes to nonvolatile storage using DCPacker) so any compatibility issues would be temporary anyway.

The only compatibility issue I could see is if there's a networked game that sends some 64-bit random seed and there are mixed clients running Panda versions before/after this change lands which end up disagreeing on the seed. My stance is that's a case that's unlikely enough, and easily resolved enough, that we can merge this without fear of inconvenience to our users.

PR #752, opened 2019-10-04
Update DirectScrolledList string type checking

Hawkheart:
The previous implementation (which used type checking via `__class__.__name__`) was causing issues while porting my project from Python 2 to be Python 2/3 compatible.

I've replaced these calls; additionally, I've replaced the more acceptable `type(item) == type('')` checking to match other string-type-checks done in the direct.gui module.

### DIFF
--- a/direct/src/gui/DirectScrolledList.py
+++ b/direct/src/gui/DirectScrolledList.py
@@ -9,6 +9,12 @@ from direct.directnotify import DirectNotifyGlobal
 from direct.task.Task import Task
 from .DirectFrame import *
 from .DirectButton import *
+import sys
+
+if sys.version_info >= (3,0):
+    stringType = str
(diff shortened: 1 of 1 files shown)


rdb:
Doesn't this break passing in a `unicode` object in Python 2?  What kind of issues were you running into with the existing approach of checking against `basestring`?

Hawkheart:
The existing approach used in this file was not checking against basestring (it was checking type or class name for exact equality to `str`), so Unicode objects were already causing crashes due to accidental `reparentTo` calls.

I don't believe the current patch will work with `bytes` on Python 3, however, if that is required.

rdb:
Cherry-picked as 984d70e8e0e84dec1848965fd336648672b7731d.  Thanks!

PR #753, opened 2019-10-05
Make Socket Send methods take vector_uchar instead of string

Hawkheart:
This changes the version of these functions as exposed to Python to take `bytes`, rather than `string`s, which was causing issues when sending raw, binary data over a SocketUDP in my application when running on Python 3.

Note that as at this time, I've removed the prior implementation of `Send` and `SendTo` that take strings, compatibility with C++ applications depending on those is broken. I can re-add the `string` versions of these functions if necessary. Additionally, I have not yet touched `SocketTCP`, which also exposes `string` based functions to Python.

(Apologies if anything isn't correct here - I haven't touched the C++ side of the Panda codebase previously.)

### DIFF
--- a/panda/src/nativenet/socket_tcp.h
+++ b/panda/src/nativenet/socket_tcp.h
@@ -3,6 +3,7 @@
 
 #include "pandabase.h"
 #include "socket_ip.h"
+#include "vector_uchar.h"
 
 /**
  * Base functionality for a TCP connected socket This class is pretty useless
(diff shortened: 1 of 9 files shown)


rdb:
I endorse this change.

I don't think we need `std::string` overloads if there is nothing in Panda using it with `std::string`.

Hawkheart:
I've pushed an additional commit switching TCP sockets over to use `vector`s instead.

There were several uses of the string overloads within Panda's codebase -- all within `net`'s `connection` class. I've migrated all of these uses over to the `vector_uchar` interface.

Moguri:
@rdb any issues with this PR or requested changes?

PR #763, opened 2019-10-29
basic warnings for nonsense functions

BMaxV:
There are issues with DirectGUI, to fix them, let's introduce some warnings and remove code later.

### DIFF
--- a/direct/src/gui/OnscreenGeom.py
+++ b/direct/src/gui/OnscreenGeom.py
@@ -6,6 +6,8 @@ from panda3d.core import *
 from direct.showbase.DirectObject import DirectObject
 import sys
 
+import warnings
+
 if sys.version_info >= (3, 0):
     stringType = str
 else:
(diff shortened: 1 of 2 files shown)


direct/src/gui/OnscreenGeom.py
```diff
@@ -111,6 +113,7 @@ def setGeom(self, geom,
                 self.setColor(color[0], color[1], color[2], color[3])
 
     def getGeom(self):
+        warnings.warn("stop using this function",FutureWarning)
```
rdb: Style-wise, there should be a space after the comma.  We could probably improve the wording of the warning message, like "This method is vestigial and will be removed shortly."

Moguri:
@rdb were you waiting on CI (i.e., existing tests) to verify that these methods are vestigial or are there new tests that should be created?

rdb:
Hmm, how can we automate testing that effectively—perhaps it is possible to configure pytest that it sees warnings as errors?

Moguri:
@BMaxV, Are you still interested in pursuing this PR? If not, or if I don't here from you in two weeks, I intend to close this PR. It can always be re-opened.

BMaxV:
Yes I am still interested in this PR.

Moguri:
@BMaxV glad to hear that! Do we have some next steps for this PR?

Moguri:
@rdb and @BMaxV since these functions are actually being used, should we close this PR?

PR #765, opened 2019-10-30
move DGG.INITOPT into DirectGuiWidget, which is parent to all others

BMaxV:
Basically, globals are bad and sometimes even unnecessary. In this case, INITOPT is just some default that's always used for the weird construction mechanism in place with DirectGUI.

All I've done here is move the global to the DirectGuiWidget class, which is parent and then refered to it by self. in all children. 

So this should just work and not really change anything.

### DIFF
--- a/direct/src/gui/DirectGuiGlobals.py
+++ b/direct/src/gui/DirectGuiGlobals.py
@@ -17,6 +17,8 @@ drawOrder = 100
 panel = None
 
 # USEFUL GUI CONSTANTS
+#still needed for some other internal options AND THE TESTS yay.
+INITOPT = ['initopt']
 
 # Mouse buttons
 LMB = 0


BMaxV:
To be blunt, the DirectGui stuff is messed up to it's core.

We can't keep stuff defineoptions, configure, initializeoptions and write clean, understandable, functional code.

We can also not keep the old DGG, write a new interface and have both work in concert. At least I'm not sure how to do it.

We can keep DGG, if you want, but it's only used internally by utility functions.

I'm also not sure why you would write tests for the internal stuff. These are the tests that are failing: 

>test_isinitoption():
>test_set_options():

Why test them explicitly? The interface needs to work, gui has to be constructed and react to events properly.

I know it's bad style to argue against tests, but I'd really like to know why you added them...

PR #766, opened 2019-10-31
Particle offset plus "_tics_since_birth" getter

ArsThaumaturgis:
The primary purpose of this pull request is to add a means of offsetting a particle system's first birth. 

To this end, a new variable, "_initial_birth_offset", has been added, along with a getter and setter for it. Where "_tics_since_birth" was previously set to "0.0f", it's now set to the value of "_initial_birth_offset" instead, causing the first particle-birth of the system to be shifted by that amount.

There is a complication: the "start" method of "ParticleEffect" just enables the relevant particle systems; it doesn't reset "_tics_since_birth". (Although "softStart" _does_ reset "_tics_since_birth".) Furthermore, there seems to be no post-setup initialisation.

As a result, having the setter-method for "_initial_birth_offset" just setting the value of the variable's isn't quite enough: it's essentially too late, as "_tics_since_birth" has already been assigned the default value of "0.0f".

Thus, I currently have the setter for "_initial_birth_offset" also reset "_tics_since_birth"; while perhaps not ideal, it does solve the problem. I'm very open to better solutions, however!

I intend to add unit-tests to this pull-request at a later point; I want to get this much uploaded now, at least!

This pull-request also adds a method to query the value of "_tics_since_birth", which may be useful for timing other objects. (I recall that it's generally preferred to split separate changes, but this one seems so small that I'm hoping that it's okay to just include it with the main request. ^^; )

And finally, a quick program that demonstrates the new functionality (aside from getting "_tics_since_birth"):

```python
from direct.particles.ParticleEffect import ParticleEffect
from panda3d.core import Vec4

class Game(ShowBase):

    def __init__(self):
        ShowBase.__init__(self)
        self.enableParticles()
        self.win.setClearColor(Vec4(0, 0, 0, 1))

        self.basicEffect = ParticleEffect()
        self.basicEffect.loadConfig("testParticles.ptf")
        self.basicEffect.setPos(0.75, 5, -0.3)

        self.offsetEffect = ParticleEffect()
        self.offsetEffect.loadConfig("testParticles.ptf")
        self.offsetEffect.getParticlesList()[0].setInitialBirthOffset(-0.25)
        self.offsetEffect.setPos(-0.75, 5, -0.3)

        self.basicEffect.start(parent = render, renderParent = render)
        self.offsetEffect.start(parent = render, renderParent = render)

app = Game()
app.run()
```

"testParticles.ptf" is a simple point-particle effect:
```python

self.reset()
self.setPos(0.000, 0.000, 0.000)
self.setHpr(0.000, 0.000, 0.000)
self.setScale(1.000, 1.000, 1.000)
p0 = Particles.Particles('particles-1')
# Particles parameters
p0.setFactory("PointParticleFactory")
p0.setRenderer("PointParticleRenderer")
p0.setEmitter("SphereSurfaceEmitter")
p0.setPoolSize(1024)
p0.setBirthRate(2.0000)
p0.setLitterSize(100)
p0.setLitterSpread(0)
p0.setSystemLifespan(0.0000)
p0.setLocalVelocityFlag(1)
p0.setSystemGrowsOlderFlag(0)
# Factory parameters
p0.factory.setLifespanBase(0.5000)
p0.factory.setLifespanSpread(0.0000)
p0.factory.setMassBase(1.0000)
p0.factory.setMassSpread(0.0000)
p0.factory.setTerminalVelocityBase(400.0000)
p0.factory.setTerminalVelocitySpread(0.0000)
# Point factory parameters
# Renderer parameters
p0.renderer.setAlphaMode(BaseParticleRenderer.PRALPHAOUT)
p0.renderer.setUserAlpha(1.00)
# Point parameters
p0.renderer.setPointSize(1.00)
p0.renderer.setStartColor(Vec4(1.00, 1.00, 1.00, 1.00))
p0.renderer.setEndColor(Vec4(1.00, 1.00, 1.00, 1.00))
p0.renderer.setBlendType(PointParticleRenderer.PPONECOLOR)
p0.renderer.setBlendMethod(BaseParticleRenderer.PPNOBLEND)
# Emitter parameters
p0.emitter.setEmissionType(BaseParticleEmitter.ETRADIATE)
p0.emitter.setAmplitude(1.0000)
p0.emitter.setAmplitudeSpread(0.0000)
p0.emitter.setOffsetForce(Vec3(0.0000, 0.0000, 1.0000))
p0.emitter.setExplicitLaunchVector(Vec3(1.0000, 0.0000, 0.0000))
p0.emitter.setRadiateOrigin(Point3(0.0000, 0.0000, 0.0000))
# Sphere Surface parameters
p0.emitter.setRadius(0.1000)
self.addParticles(p0)
```
Screenshot from 2019-10-31 22-42-07

### DIFF
--- a/panda/src/particlesystem/particleSystem.I
+++ b/panda/src/particlesystem/particleSystem.I
@@ -157,6 +157,15 @@ set_floor_z(PN_stdfloat z) {
 
  */
 INLINE void ParticleSystem::
+set_inital_birth_offset(PN_stdfloat offset) {
+  _inital_birth_offset = offset;
+}
+
+
+/**
+
+ */
+INLINE void ParticleSystem::
 set_active_system_flag(bool a) {
   _active_system_flag = a;
 }
@@ -340,6 +349,14 @@ get_factory() const {
 INLINE PN_stdfloat ParticleSystem::
 get_floor_z() const {
   return _floor_z;
+}
+
+/**
+
+*/
+INLINE PN_stdfloat ParticleSystem::
+get_inital_birth_offset() const {
+  return _inital_birth_offset;
 }
 
 /**
--- a/panda/src/particlesystem/particleSystem.I
+++ b/panda/src/particlesystem/particleSystem.I
@@ -44,7 +44,7 @@ clear_to_initial() {
       kill_particle(i);
     }
   }
-  _tics_since_birth = _inital_birth_offset;
+  _tics_since_birth = _initial_birth_offset;
 }
 
 /**
@@ -55,7 +55,7 @@ soft_start(PN_stdfloat br) {
   if (br > 0.0)
     set_birth_rate(br);
   _cur_birth_rate = _birth_rate;
-  _tics_since_birth = _inital_birth_offset;
+  _tics_since_birth = _initial_birth_offset;
 }
 
 /**
@@ -66,7 +66,7 @@ soft_stop(PN_stdfloat br) {
   if (br > 0.0)
     set_soft_birth_rate(br);
   _cur_birth_rate = _soft_birth_rate;
-  _tics_since_birth = _inital_birth_offset;
+  _tics_since_birth = _initial_birth_offset;
 }
 
 /**
@@ -157,8 +157,9 @@ set_floor_z(PN_stdfloat z) {
 
  */
 INLINE void ParticleSystem::
-set_inital_birth_offset(PN_stdfloat offset) {
-  _inital_birth_offset = offset;
+set_initial_birth_offset(PN_stdfloat offset) {
+  _initial_birth_offset = offset;
+  _tics_since_birth = offset;
 }
 
 
@@ -355,8 +356,8 @@ get_floor_z() const {
 
 */
 INLINE PN_stdfloat ParticleSystem::
-get_inital_birth_offset() const {
-  return _inital_birth_offset;
+get_initial_birth_offset() const {
+  return _initial_birth_offset;
 }
 
 /**
--- a/panda/src/particlesystem/particleSystem.I
+++ b/panda/src/particlesystem/particleSystem.I
@@ -44,7 +44,7 @@ clear_to_initial() {
       kill_particle(i);
     }
   }
-  _tics_since_birth = 0.0f;
+  _tics_since_birth = _inital_birth_offset;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


rdb:
Thank you.  I'll review this later, but I just wanted to remark that the fact that a new field is added to the PandaSystem class unfortunately means that it cannot be picked up for 1.10.5, since changing the size of PandaSystem breaks ABI compatibility.  I will consider this for master but that means that you would need to point your requirements.txt to our package index for the development branch.  If you can implement this change in a way that doesn't change the memory layout of the ParticleSystem class, it could be considered for 1.10.5.

For the record, the e-mail address you used to make the commits has not been added to your GitHub account.  This is not a problem per se, it just means that your commits aren't linked up to your GitHub account.

ArsThaumaturgis:
> Thank you. 

It's my pleasure. ^_^

> ... I just wanted to remark that the fact that a new field is added to the PandaSystem class unfortunately means that it cannot be picked up for 1.10.5, since changing the size of PandaSystem breaks ABI compatibility. 

Ah, I see. That's a pity!

> If you can implement this change in a way that doesn't change the memory layout of the ParticleSystem class, it could be considered for 1.10.5.

Hmm... I can see a way to do much the same thing without the extra variable, although it would be less robust: instead of having a value to which the tic-count is set, I could make a sort of "special offseted restart" method that essentially calls soft-start and then offsets the tic-count.

Perhaps that version for 1.10.5 (via a new pull-request), and this version for master?

(One advantage of that version is that it should be even less likely to cause problems: it would add one new method that does exactly two things: call an extant method, and set an extant variable.)

Or would the addition of a new method cause the same problem?

If so, I suppose that I could add a new parameter to soft-start...

> I will consider this for master but that means that you would need to point your requirements.txt to our package index for the development branch. 

Fair enough, and noted! Thank you. ^_^

> For the record, the e-mail address you used to make the commits has not been added to your GitHub account. This is not a problem per se, it just means that your commits aren't linked up to your GitHub account.

Hmm... I'm not terribly bothered about my commits being linked to my account, unless it's inconvenient for you in some way.

If it is, then where would I go to add that address to my GitHub account? And is the address being used for commits simply the primary email address that I set in my GitHub settings, or is it set somewhere else?

ArsThaumaturgis:
I've added a pull request for a simpler implementation, which doesn't introduce a new variable to Panda: #769

rdb:
Closing this in favour of #769, which addresses the use case in question (in an IMHO more elegant way).  We can reconsider if a new use case pops up that requires the solution implemented by this PR.

PR #768, opened 2019-11-01
Fix reset text scale on item unhighlight

fireclawthefox:
Set text scale to previous unhighlighted scale on unhighlight to keep custom scales

### DIFF
--- a/tests/gui/test_DirectOptionMenu.py
+++ b/tests/gui/test_DirectOptionMenu.py
@@ -1,0 +1,70 @@
+from direct.gui.DirectOptionMenu import DirectOptionMenu
+
+def test_menu_destroy():
+    menu = DirectOptionMenu(items=["item1", "item2"])
+    menu.destroy()
+
+def test_showPopupMenu():
+    menu = DirectOptionMenu()
+    exceptionThrown = False
+    try:
+        menu.showPopupMenu()
+    except:
+        # Showing an option menu without items will raise an exception
+        exceptionThrown = True
+    assert exceptionThrown
+
+    menu["items"] = ["item1", "item2"]
+    #TODO: This call requires render2d to be set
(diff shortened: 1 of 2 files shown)


rdb:
Would you be willing to show some code demonstrating the bug, or better yet, add a unit test?  That will allow me to merge this into the next release with confidence, thanks!

fireclawthefox:
Added unittests for DirectOptionMenu. The test for the text scale behavior is at the bottom.

tests/gui/test_DirectOptionMenu.py
```diff
@@ -0,0 +1,70 @@
...
+
+def test_showPopupMenu():
+    menu = DirectOptionMenu()
+    exceptionThrown = False
+    try:
+        menu.showPopupMenu()
+    except:
+        # Showing an option menu without items will raise an exception
+        exceptionThrown = True
+    assert exceptionThrown
```
rdb: I'll fix this for now, but for future reference, pytest has a more idiomatic way to do this:
```python
with pytest.raises(Exception):
    menu.showPopupMenu()
```

rdb:
Thank you, committed to 1.10.5 branch in eb3b45ebb15d996ed0e155c087628fea9ea1c700 and e6fc6c695dbb9c61dc316c8fc3b17e2d41983245.

I've fixed the "TODO" in your unit test by allowing DirectOptionMenu to be used without ShowBase, in 69e8b4ed3c2156e2e9015bd8c048a80bb4269a18 and 2e3b350510cb91aa0e88400434d371621cb6bb44.

fireclawthefox:
Thanks for the hint with the pytest exception test and fixing the todo. Changing the variable is fine, I don't know of a good reason for it to be public either.

PR #772, opened 2019-11-03
discussion about what direct gui property access should look like

BMaxV:
Basically something we talked about on IRC.

Direct gui property access doesn't always work as expected.

This is meant as discussion about what the access should look like. Nothing about how it works concerning states or which properties exist should change.


Moguri:
Would this be better as an issue than as a PR?

Moguri:
Closing due to inactivity and that Issues are better for discussion.

PR #776, opened 2019-11-09
new setter for text, image and geom for DirectFrame

BMaxV:
Same as before, setImage setText and setGeom have a lot of duplicate code and some constants where they shouldn't be. The new_set combines common code. There are still some defaults that could be move elsewhere.

### DIFF
--- a/direct/src/gui/DirectFrame.py
+++ b/direct/src/gui/DirectFrame.py
@@ -68,15 +68,26 @@ class DirectFrame(DirectGuiWidget):
         self.setText()
 
     def setText(self, text=None):
-        
+        if text is not None:
+            self["text"] = text
+
+        text = self["text"]
         component_class = OnscreenText
-        
         component_kwargs ={
                 "scale": 1,
                 "mayChange": self['textMayChange'],
                 "sort": DGG.TEXT_SORT_INDEX, # this is "30"
         }
-        self.new_set("text", text, component_kwargs, component_class)
+        
+        if text is None:
+            text_list = (None,) * self['numStates']
+        else:
+            if self.inputTypeValidation(name, text):
+                text_list = (text,) * self['numStates']
+            else:
+                text_list = text
+        
+        self.reinitComponent("text", text_list, component_kwargs, component_class)
 
     def clearGeom(self):
         self['geom'] = None
@@ -98,23 +109,15 @@ class DirectFrame(DirectGuiWidget):
                 return True
         return False
 
-    def new_set(self, name, object_, component_kwargs,component_class):
+    def reinitComponent(self, name, object_list, component_kwargs,component_class):
         """set function common code"""
         assert name in ("geom", "image", "text")
-        if object_ is not None:
-            self[name] = object_
-
-        object_ = self[name]
+        
         
         #object list from here...        
-        if object_ is None:
-            object_list = (None,) * self['numStates']
-        else:
-            if self.inputTypeValidation(name, object_):
-                object_list = (object_,) * self['numStates']
-            else:
-                object_list = object_
-
+        
+        if len(object_list) != self['numStates']:
+            raise ValueError
         
         # constants should be local to or default arguments of constructors
 
@@ -144,14 +147,24 @@ class DirectFrame(DirectGuiWidget):
             #c += 1
 
     def setGeom(self, geom=None):
-        
+        if geom is not None:
+            self["geom"] = geom
+
+        geom = self["name"]
         component_class = OnscreenGeom
-        
         component_kwargs = {
                 "scale": 1,
                 "sort": DGG.GEOM_SORT_INDEX, # this is "20"
             }
-        self.new_set("geom", geom, component_kwargs, component_class)
+        
+        if geom is None:
+            geom_list = (None,) * self['numStates']
+        else:
+            if self.inputTypeValidation(name, geom):
+                geom_list = (geom,) * self['numStates']
+            else:
+                geom_list = geom
+        self.reinitComponent("geom", geom_list, component_kwargs, component_class)
 
     def clearImage(self):
         self['image'] = None
@@ -159,11 +172,22 @@ class DirectFrame(DirectGuiWidget):
 
     def setImage(self, image=None):
         
+        if image is not None:
+            self["image"] = image
+
+        image = self["image"]
         component_class=OnscreenImage,
-        
         component_kwargs = {
                 "scale": 1,
                 "sort": DGG.IMAGE_SORT_INDEX, # this is "10"
             }
             
-        self.new_set("image", image, component_kwargs, component_class)
+        if image is None:
+            image_list = (None,) * self['numStates']
+        else:
+            if self.inputTypeValidation(name, image):
+                image_list = (image,) * self['numStates']
+            else:
+                image_list = image
+                
+        self.reinitComponent("image", image_list, component_kwargs, component_class)
--- a/direct/src/gui/DirectFrame.py
+++ b/direct/src/gui/DirectFrame.py
@@ -68,6 +68,7 @@ class DirectFrame(DirectGuiWidget):
         self.setText()
 
     def setText(self, text=None):
+        #TODO remove type checks
         if text is not None:
             self["text"] = text
 
@@ -82,34 +83,18 @@ class DirectFrame(DirectGuiWidget):
         if text is None:
             text_list = (None,) * self['numStates']
         else:
-            if self.inputTypeValidation("text", text):#used by button and directmenu
+            if isinstance(text, stringType):#used by button and directmenu
                 text_list = (text,) * self['numStates']
             else:
                 text_list = text
         
-        self.reinitComponent("text", text_list, component_kwargs, component_class)
+        self.__reinitComponent("text", text_list, component_kwargs, component_class)
 
     def clearGeom(self):
         self['geom'] = None
         self.setGeom()
 
-    def inputTypeValidation(self, name, object_):
-        if isinstance(object_, stringType):
-            return True
-        if name == "image":
-            if ((isinstance(object_, NodePath) or
-                isinstance(object_, Texture)) 
-                or
-                ((len(object_) == 2) and# Its a model/node pair of strings
-                isinstance(object_[0], stringType) and
-                isinstance(object_[1], stringType))):
-                return True
-        elif name == "geom":
-            if (isinstance(object_, NodePath)):
-                return True
-        return False
-
-    def reinitComponent(self, name, object_list, component_kwargs,component_class):
+    def __reinitComponent(self, name, object_list, component_kwargs,component_class):
         """set function common code"""
         assert name in ("geom", "image", "text")
         
@@ -141,6 +126,7 @@ class DirectFrame(DirectGuiWidget):
                 )
 
     def setGeom(self, geom=None):
+        #TODO remove type checks
         if geom is not None:
             self["geom"] = geom
 
@@ -154,18 +140,18 @@ class DirectFrame(DirectGuiWidget):
         if geom is None:
             geom_list = (None,) * self['numStates']
         else:
-            if self.inputTypeValidation(name, geom):
+            if isinstance(object_, NodePath):
                 geom_list = (geom,) * self['numStates']
             else:
                 geom_list = geom
-        self.reinitComponent("geom", geom_list, component_kwargs, component_class)
+        self.__reinitComponent("geom", geom_list, component_kwargs, component_class)
 
     def clearImage(self):
         self['image'] = None
         self.setImage()
 
     def setImage(self, image=None):
-        
+        #TODO remove type checks
         if image is not None:
             self["image"] = image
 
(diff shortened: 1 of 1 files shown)


direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+                isinstance(object_, Texture)) 
+                or
+                ((len(object_) == 2) and# Its a model/node pair of strings
+                isinstance(object_[0], stringType) and
+                isinstance(object_[1], stringType))):
+                return True
+        elif name == "geom":
+            if (isinstance(object_, NodePath)):
+                return True
+
```
darktohka: I think it would be good to have a "return False" here, as this function right now would return None instead of False.

BMaxV: Sure why not, it's only helping cleaning up input...

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+                ((len(object_) == 2) and# Its a model/node pair of strings
+                isinstance(object_[0], stringType) and
+                isinstance(object_[1], stringType))):
+                return True
+        elif name == "geom":
+            if (isinstance(object_, NodePath)):
+                return True
+        return False
+
+    def new_set(self, name, object_=None):
```
rdb: You should name this method something descriptive.  It should probably also a private method, or were you imagining this to be part of the public interface?

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+        """my new code golf par is 1"""
```
rdb: I'm not sure what this docstring is intended to communicate to a peruser of the API reference.

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+        assert name in ["geom", "image", "text"]
+        if object_ is not None:
+            self[name] = object_
+
+        object_ = self[name]
+
+        if object_ == None:
```
rdb: Use `is None`

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+    def long_type_check(self, name, object_):
+        if isinstance(object_, stringType):
+            return True
+        if name == "image":
+            if ((isinstance(object_, NodePath) or
+                isinstance(object_, Texture)) 
+                or
+                ((len(object_) == 2) and# Its a model/node pair of strings
+                isinstance(object_[0], stringType) and
+                isinstance(object_[1], stringType))):
```
rdb: This if check is hard to read, it is better broken up.

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
-        geom = self['geom']
-
-        if geom == None:
-            # Passed in None
-            geomList = (None,) * self['numStates']
-        elif isinstance(geom, NodePath) or \
-             isinstance(geom, stringType):
-            # Passed in a single node path, make a tuple out of it
-            geomList = (geom,) * self['numStates']
+    def long_type_check(self, name, object_):
```
rdb: This should probably be private, and named more descriptively.

More generally, I'm not entirely convinced that having a bunch of if checks for each individual property in here is the best way to do it.  Is there a way to move the type check back to the individual setters, while still keeping common code in `new_set`?  Or to pass the permitted types as an argument to `new_set`, or otherwise breaking up `new_set` into multiple methods?

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
-                    self.destroycomponent(component)
+            if self.long_type_check(name, object_):
+                object_list = (object_,) * self['numStates']
+            else:
+                object_list = object_
+
+        name_based_classes = {
+            "geom": OnscreenGeom,
+            "image": OnscreenImage,
+            "text": OnscreenText, }
```
rdb: I don't like these lookups.  It's a lot more elegant to simply pass this as argument to the `new_set` helper.  Same for `name_based_constants`, below.

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+            "text": {
+                "scale": 1,
+                "mayChange": self['textMayChange'],
+                "sort": DGG.TEXT_SORT_INDEX, # this is "30"
+            },
+        }
+        # constants should be local to or default arguments of constructors
+
+        m = self["numStates"]
+        for c in range(self['numStates']):
```
rdb: You're querying this twice in a row, maybe you meant `for c in range(m):`?  That said, the variable should also probably be named more descriptively than a single letter.

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+            component_name = name+str(c)
```
rdb: Spaces around +

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+            comp_input = object_list[c]
+            name_based_constants[name][name] = comp_input
+
+            if self.hascomponent(component_name):
+                if comp_input == None:
```
rdb: `is None`

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+                    self.destroycomponent(component_name)
                 else:
-                    self[component + '_geom'] = geom
+                    self[component_name+"_"+name] = comp_input
             else:
-                if geom == None:
+                if comp_input == None:
```
rdb: `is None`

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
+                isinstance(object_[1], stringType))):
+                return True
+        elif name == "geom":
+            if (isinstance(object_, NodePath)):
+                return True
+        return False
+
+    def new_set(self, name, object_=None):
+        """my new code golf par is 1"""
+        assert name in ["geom", "image", "text"]
```
rdb: Prefer tuple over list here

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,113 @@ def __init__(self, parent = None, **kw):
...
-            try:
-                geom = geomList[i]
-            except IndexError:
-                geom = geomList[-1]
-
-            if self.hascomponent(component):
-                if geom == None:
-                    # Destroy component
-                    self.destroycomponent(component)
+            if self.long_type_check(name, object_):
```
rdb: It's not obvious reading this code what `long_type_check` does.  It should probably be named in such a way that I don't have to peek inside it to understand what this code does.

Furthermore, while snake_case is undoubtedly the right PEP8 style, consistency in the same file is more important.

direct/src/gui/DirectFrame.py
BMaxV: Generally yes, specially here it's not that useful.

The reason long_type_check exists, is because the loop in new_set, always has one of three inputs:

 * a tuple of (None,) with the length of self["numstates"], 
 * a tuple of (object,) copies, with the length of self["numstates"]
 * or it's assumed that whatever was given to the function already is a tuple of (object,) with length of self["numstates"].

We can move that around, but I don't think it would improve stuff.

I will move the type specific variables to the old setters, so it's more obvious which variables come from where and what stays constant during new_set.

As a side node, the tuple of (None,) is only used for destruction of components, so not even that is "cleanly" following the purpose of what a setter should do.

That whole part should be done better anyway. Like, maybe there should be a set_all_states_XXX function that takes the tuple and a set_single_state_XXX that takes the single values.

direct/src/gui/DirectFrame.py
BMaxV: People should be able to set components of the interface elements with it or something like it eventually. So "yes". I can rename it to "common_set" or something, we can talk about a good name when everything else is fine as well.

direct/src/gui/DirectFrame.py
BMaxV: long_type_check is the helper that "solves" that only inputs of correct types get processed.

It's one of the things the old functions do that I don't like. Input arguments should be well defined enough to avoid a mess like that. When everything is finished, the entire function should not be necessary. For now it was meant to remove the question of "Is this ok input" from the main body of new_set, since it doesn't have anything to do with actual setting.

To make it more obvious what the rest of the function does.

Not sure what to do here.

direct/src/gui/DirectFrame.py
BMaxV: Hm it's the state counter. Sure we can name it state_i, but I'm not sure how necessary that is. It's only used 3 times within like 20 lines  and the loop starts with "for i in self["numstates"]". I think it's pretty obvious that that is the state number.

I will change it if you insist though.

rdb: I'm talking about the assignment to `m`.  It makes sense to assign it to a temporary variable, but only if you're going to then use that variable. :)

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,136 @@ def __init__(self, parent = None, **kw):
...
         self.setGeom()
 
+    def inputTypeValidation(self, name, object_):
+        if isinstance(object_, stringType):
+            return True
+        if name == "image":
+            if ((isinstance(object_, NodePath) or
+                isinstance(object_, Texture)) 
+                or
+                ((len(object_) == 2) and# Its a model/node pair of strings
```
rdb: Please remove unnecessary parentheses in this if.

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,136 @@ def __init__(self, parent = None, **kw):
...
-                        )
+                text_list = text
+        
+        self.reinitComponent("text", text_list, component_kwargs, component_class)
 
     def clearGeom(self):
         self['geom'] = None
         self.setGeom()
 
+    def inputTypeValidation(self, name, object_):
```
rdb: Should this be made private?  Also, try to name methods as a verb, so `validateInputType` would be better

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,136 @@ def __init__(self, parent = None, **kw):
...
-            textList = (self['text'],) * self['numStates']
+            self["text"] = text
+
+        text = self["text"]
+        component_class = OnscreenText
+        component_kwargs ={
+                "scale": 1,
+                "mayChange": self['textMayChange'],
+                "sort": DGG.TEXT_SORT_INDEX, # this is "30"
+        }
```
rdb: This dict has spacing issues; but then, I think it would be cleaner to just move this down as keyword arguments to the reinitComponent call.

direct/src/gui/DirectFrame.py
```diff
@@ -50,166 +52,136 @@ def __init__(self, parent = None, **kw):
...
+            return True
+        if name == "image":
+            if ((isinstance(object_, NodePath) or
+                isinstance(object_, Texture)) 
+                or
+                ((len(object_) == 2) and# Its a model/node pair of strings
+                isinstance(object_[0], stringType) and
+                isinstance(object_[1], stringType))):
+                return True
+        elif name == "geom":
```
rdb: If we're doing if checks for this, it would be better to just do this check in setGeom, don't you think?

direct/src/gui/DirectFrame.py
BMaxV: So, reninitComponent(scale=1)? I disagree.

The point of all this in the first place is that createcomponent is superfluous. Everything that setImage and now __reinitComponent and createcomponent do, is collect inputs to call the init's of Onscreen[Stuff].

That is where these defaults belong.

But I can't replace createcomponent in one go, so collecting inputs like this will make the next PR where I do remove create component more clear. 

Then I can move the default values to the inits and just remove the dict here.

direct/src/gui/DirectFrame.py
BMaxV: We shouldn't be doing if checks, because the input type should be clear, as long as the type checks are removed and the inits are adapted at some point in the future, I don't mind moving the ifs.

BMaxV:
Eh noticed a mistake the tests didn't catch, don't merge...

rdb:
Thanks!  Your PR looked pretty close to perfect; I've made a few minor changes before merging, mostly some minor style changes.  I also moved component_kwargs into the actual keyword arguments, which makes it a little bit more compact, and I also realised that the `None` handling could be folded into the other `if` case.

You missed the ability to pass in a string into setGeom and setImage, so I fixed that too.  Furthermore, there was a reference to an `object_` in `setGeom` that didn't actually exist.  We really need exhaustive unit tests to make sure we are aware when breakage occurs; I've added a couple of unit tests in 2801165db32fee94c5f4d2e0a80da828508300c0 to account for this.

For future reference: please configure your editor to use spaces for indentation and to strip spaces at the end of lines.

Thanks again!  This is a great improvement.

PR #777, opened 2019-11-10
Create CONTRIBUTING.md

BMaxV:
Open for discussion and suggestions. Having it written down may help.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,21 @@
+
+
+some self given rules to make project management easier:
+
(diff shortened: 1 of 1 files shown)


rdb:
I think we should indeed have documentation for contributors.  It should also be linked to from [here](https://www.panda3d.org/get-involved/).  I'm not sure whether this fits best in the manual or in the source code repository, though.

rdb:
Closed in 8ae4c8eee8ff06aa952340abf8929751bab0c97a

PR #782, opened 2019-11-11
Update macOS path to /Library/Developer

amj100:
I changed the mac path to /Library/Developer and updated the macOS xcode-tools path. This should fix issue #760

### DIFF
--- a/makepanda/makepackage.py
+++ b/makepanda/makepackage.py
@@ -376,40 +376,40 @@ def MakeInstallerOSX(version, python_versions=[], **kwargs):
     if (os.path.exists("dstroot")): oscmd("rm -rf dstroot")
     if (os.path.exists("Panda3D-rw.dmg")): oscmd('rm -f Panda3D-rw.dmg')
 
-    oscmd("mkdir -p dstroot/base/Developer/Panda3D/lib")
-    oscmd("mkdir -p dstroot/base/Developer/Panda3D/etc")
-    oscmd("cp %s/etc/Config.prc           dstroot/base/Developer/Panda3D/etc/Config.prc" % outputdir)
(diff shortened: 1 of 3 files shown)


amj100:
You're right about that.
Also, here is documentation on Apple's file system if anyone is wondering: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html
The /Library/Developer directory seems to be the new corresponding place for the previous /Developer directory.

el-dee:
Moving to /Library/Developer is a good short term compromise, but I still believe the long term solution would be to make Panda3d a Framework, /Library/Developer is not really documented and mainly used by the command line tools of Xcode. On the other hand, Frameworks are documented properly by Apple and many if not all libraires and SDKs are shipped as frameworks.

That said, there is much more work to do to transform the Panda SDK into a Framework as the layout will need to be changed to conform to Apple requirements. And there might be a heavier impact on existing application (but that is to be confirmed)

rdb:
I see that this was closed and I cannot reopen it because the original branch was deleted; can we still apply this for 1.10.5, since we likely will not be ready to apply the framework solution in the short term?

el-dee:
Since there will be two builds, one for older MacOS and one for Catalina, wouldn't it be better to still target /Developer if we are doing a build for Mac <= Mojave and target /Library/Developer for Catalina ?

Doing so would avoid to try and detect a previous installation in /Developer and delete it on older version, and on Catalina there would be no migration as /Developer can not exist.

rdb:
My reasoning for making a second build that targeted 10.9+ (instead of 10.15+) is that we would kill two birds with one stone, by making the second build compiled with libc++.

However, your reasoning does make sense as well, and it's easier to separate out the two cases in makepanda.

PR #783, opened 2019-11-11
build_apps: Fix defaults for package_data_dirs

### LINKED ISSUE #779: setuptools: PACKAGE_DATA_DIRS in dist/commands.py is not used
In dist/commands.y a variable PACKAGE_DATA_DIRS contains extra dirs for cefpython3 matplotlib, ...  

https://github.com/panda3d/panda3d/blob/f1b235e06594f4c0a2c028adfe2ef2241e4091e0/direct/src/dist/commands.py#L102

However it is never used, so when including cefpython3 one has to copy the entries in their own setup.py file.

Moguri:
Fixes #779

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -365,7 +365,7 @@ class build_apps(setuptools.Command):
         tmp.update(self.file_handlers)
         self.file_handlers = tmp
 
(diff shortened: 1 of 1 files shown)


Moguri:
I don't see an incoming branch for the 1.10.x branch, so I'm making a PR instead.

Issue #757, opened 2019-10-16
BulletVehicle.getWheel(0).getSteering() is in radians


davidrusu:
From the docs it should really be returning values in degrees:
https://www.panda3d.org/reference/python/classpanda3d_1_1bullet_1_1BulletWheel.html#a3b74c45b79964079f3921417a26be64c

Here's a test program where you can see the behaviour:
[vehicle_wheel_steering_bug.zip](https://github.com/panda3d/panda3d/files/3732198/vehicle_wheel_steering_bug.zip)
And my output from running this program:
```
Set steering to -45.000000 but reading back as -0.785398 (converted to degrees is -45.000001)
```

PR #784, opened 2019-11-13
bullet: fix wheel.getSteering() to return in degrees #757

leotada:
This should fix issue #757

### DIFF
--- a/panda/src/bullet/bulletWheel.cxx
+++ b/panda/src/bullet/bulletWheel.cxx
@@ -223,7 +223,7 @@ void BulletWheel::
 set_steering(PN_stdfloat value) {
   LightMutexHolder holder(BulletWorld::get_global_lock());
 
-  _info.m_steering = (btScalar)value;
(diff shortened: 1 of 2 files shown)


Moguri:
This looks good to me. What testing have you done? Some unit tests for this would be nice, but I don't think they are required to accept this PR.

rdb:
Thanks, can you add unit tests?  I think unit tests should be a requirement for all PRs that change existing behaviours.

leotada:
Ok, I will try to add the tests.

leotada:
Is this ok?

rdb:
getSteering() and setSteering() should use the same units.  Your unit test seems to indicate that setSteering takes a value in radians.  Could you fix setSteering to also use degrees?

leotada:
Of course. The documentation don't mention the unit then I didn't know If I change.

leotada:
@rdb please review! I can't resolve the change request status in Github.

PR #786, opened 2019-11-13
egldisplay: EGL or X11 is valid for GLES1/2

pmp-p:


CFSworks:
@rdb Is `egldisplay` able to operate without X11? There's a link directive in there for `p3x11display` which isn't built if `HAVE_X11=NO`, so I'm not sure what to make of this PR.

rdb:
@CFSworks at the moment, egldisplay requires X11.  It would not be much work to adapt it to work with other backends (I vaguely recall someone did effort to get it to work on Windows), but this is the current reality.

CFSworks:
@pmp-p I think I need to close this PR, then. Should the EGL-depends-on-X11 situation change (or even if your current work ends up improving that situation!) please don't hesitate to open a fresh one! :)

PR #792, opened 2019-11-26
Added EXPCLs to fix linking errors with clang in windows

Cerberus1746:
I am dividing the pull request #791 and adding into the main branch as requested

Those changes add EXPCLs to sub classes in some classes. Because without then, linking errors ocurred with clang in windows while using mingw32 in Msys2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ __pycache__/
 .tox/
 .cache/
 .pytest_cache/
+/.settings/
+/.cproject
+/.project
--- a/dtool/src/dtoolutil/textEncoder.h
+++ b/dtool/src/dtoolutil/textEncoder.h
@@ -140,6 +140,7 @@ private:
 
   static Encoding _default_encoding;
 };
+
 EXPCL_DTOOL_DTOOLUTIL std::ostream &
 operator << (std::ostream &out, TextEncoder::Encoding encoding);
 EXPCL_DTOOL_DTOOLUTIL std::istream &
--- a/dtool/src/dtoolutil/textEncoder.h
+++ b/dtool/src/dtoolutil/textEncoder.h
@@ -140,7 +140,6 @@ private:
 
   static Encoding _default_encoding;
 };
-
 EXPCL_DTOOL_DTOOLUTIL std::ostream &
 operator << (std::ostream &out, TextEncoder::Encoding encoding);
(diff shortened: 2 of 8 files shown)


CFSworks:
Looks good to me so far, though if we're updating the EXPCLs, it would be prudent to silence the warnings too:

- Drop EXPCL from `operator <<` in `textEncoder.h` and `lens.h`
- `EXPCL_PANDA_PNMIMAGE INLINE` -> `INLINE` in `convert_srgb.h` (realigning the continuation lines as necessary)

As a nitpick, the commit message should probably be something more like `general: Update EXPCLs to be MinGW-friendly` (since technically Clang can follow the MSVC ABI, where the EXPCLs are fine, or the MinGW ABI, where the EXPCLs aren't precise enough yet).

You can use `git commit --amend` (or the "amend commit" button on your Git GUI) and do a force-push to revise this PR without having to add commits to it. (Note that force-push is *not* something you should use on a branch you're using to collaborate with others, but for a PR it's fine.)

Cerberus1746:
Oh, I forgot to mention here that I made the changes @CFSworks asked.

dtool/src/dtoolutil/textEncoder.h
```diff
@@ -140,7 +140,6 @@ class EXPCL_DTOOL_DTOOLUTIL TextEncoder {
 
   static Encoding _default_encoding;
 };
-
```
CFSworks: This space should probably remain

panda/src/gobj/lens.h
```diff
@@ -399,7 +399,7 @@ class EXPCL_PANDA_GOBJ Lens : public TypedWritableReferenceCount {
   static TypeHandle _type_handle;
 };
 
-EXPCL_PANDA_GOBJ INLINE std::ostream &operator << (std::ostream &out, const Lens &lens);
+std::ostream &operator << (std::ostream &out, const Lens &lens);
```
CFSworks: `INLINE` deleted here

.gitignore
```diff
@@ -63,3 +63,6 @@ __pycache__/
 .tox/
 .cache/
 .pytest_cache/
+/.settings/
+/.cproject
+/.project
```
rdb: In the future, please don't put unrelated changes in the same commit.

rdb:
Could you clarify why these functions don't need to be exported from the DLL?

PR #793, opened 2019-11-26
Changed macros and definition for the engine to compile with MingW

leandro-benedet-garcia:
This is the minimum necessary changes together with #792 to be able to compile the engine with mingw

Some thing to note is the file panda/src/device/xInputDevice.cxx:
```cpp
#ifdef _MSC_VER
typedef struct _XINPUT_BATTERY_INFORMATION {
  BYTE BatteryType;
  BYTE BatteryLevel;
} XINPUT_BATTERY_INFORMATION;
#endif
```
Which I had to add the `#ifdef` because in msys it raised a error saying `redefinition of * '_XINPUT_BATTERY_INFORMATION'`

And the file panda/src/downloader/config_downloader.h:
```cpp
#ifndef AI_ADDRCONFIG
#define AI_ADDRCONFIG 0x00000400
#endif

#ifndef AI_V4MAPPED
#define AI_V4MAPPED 0x00000800
#endif
```
Where I had to add those defs because for some reason, the file `ws2tcpip.h` had conditions which weren't triggering and wasn't adding them correctly.

### DIFF
--- a/dtool/src/dtoolutil/textEncoder.h
+++ b/dtool/src/dtoolutil/textEncoder.h
@@ -141,6 +141,7 @@ private:
   static Encoding _default_encoding;
 };
 
+
 EXPCL_DTOOL_DTOOLUTIL std::ostream &
 operator << (std::ostream &out, TextEncoder::Encoding encoding);
 EXPCL_DTOOL_DTOOLUTIL std::istream &
--- a/dtool/src/dtoolutil/textEncoder.h
+++ b/dtool/src/dtoolutil/textEncoder.h
@@ -152,8 +152,7 @@ operator >> (std::istream &in, TextEncoder::Encoding &encoding);
 
 // This function is declared inline to minimize the risk of link conflicts
 // should another third-party module also define the same output operator.
-INLINE EXPCL_DTOOL_DTOOLUTIL std::ostream &
-operator << (std::ostream &out, const std::wstring &str);
+INLINE std::ostream & operator << (std::ostream &out, const std::wstring &str);
 
 #include "textEncoder.I"
 
--- a/dtool/src/dtoolutil/textEncoder.h
+++ b/dtool/src/dtoolutil/textEncoder.h
@@ -141,7 +141,6 @@ private:
   static Encoding _default_encoding;
 };
 
-
 EXPCL_DTOOL_DTOOLUTIL std::ostream &
 operator << (std::ostream &out, TextEncoder::Encoding encoding);
 EXPCL_DTOOL_DTOOLUTIL std::istream &
--- a/panda/src/device/xInputDevice.cxx
+++ b/panda/src/device/xInputDevice.cxx
@@ -68,10 +68,16 @@
 #define BATTERY_LEVEL_FULL 0x03
 #endif
 
+/**
+ * With MingW32 this raises the error:
+ * redefinition of * '_XINPUT_BATTERY_INFORMATION'
+ */
+#ifdef _MSC_VER
 typedef struct _XINPUT_BATTERY_INFORMATION {
   BYTE BatteryType;
   BYTE BatteryLevel;
 } XINPUT_BATTERY_INFORMATION;
+#endif
 
 // Undocumented, I figured out how this looks by trial and error.
 typedef struct _XINPUT_BUSINFO {
--- a/panda/src/device/xInputDevice.cxx
+++ b/panda/src/device/xInputDevice.cxx
@@ -68,10 +68,8 @@
 #define BATTERY_LEVEL_FULL 0x03
 #endif
 
-/**
- * With MingW32 this raises the error:
- * redefinition of * '_XINPUT_BATTERY_INFORMATION'
- */
+// With MingW32 this raises the error:
+// Redefinition of '_XINPUT_BATTERY_INFORMATION'
 #ifdef _MSC_VER
 typedef struct _XINPUT_BATTERY_INFORMATION {
   BYTE BatteryType;
--- a/panda/src/downloader/bioPtr.cxx
+++ b/panda/src/downloader/bioPtr.cxx
@@ -31,6 +31,18 @@
 #include <fcntl.h>
 #endif
 
+// For some reason in msys those two macros are not defined correctly in the
+// header file ws2tcpip.h
+// Also, those lines will be removed when the engine change
+// _WIN32_WINNT to 0x0600
+#ifndef AI_ADDRCONFIG
+#define AI_ADDRCONFIG 0x00000400
+#endif
+
+#ifndef AI_V4MAPPED
+#define AI_V4MAPPED 0x00000800
+#endif
+
 using std::string;
 
 #ifdef _WIN32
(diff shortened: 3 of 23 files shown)


panda/src/downloader/httpDate.cxx
```diff
@@ -226,7 +226,7 @@ HTTPDate(const string &format) {
     time_t now = time(nullptr);
     struct tm *tp = localtime(&now);
     _time -= tp->tm_gmtoff;
-#elif defined(_WIN32)
+#elif defined(_MSC_VER)
     long int timezone;
     _get_timezone(&timezone);
```
CFSworks: This is tricky. `_get_timezone` is actually part of the Microsoft C runtime library, so protecting it with `_MSC_VER` (which checks for the compiler) isn't *quite* right.

panda/src/device/xInputDevice.cxx
```diff
@@ -68,10 +68,16 @@
 #define BATTERY_LEVEL_FULL 0x03
 #endif
 
+/**
```
CFSworks: This style of comment is for the documentation generator. A `//` comment is probably more appropriate here instead.

panda/src/downloader/config_downloader.h
```diff
@@ -24,6 +24,18 @@
 #include "configVariableFilename.h"
 #include "configVariableList.h"
 
+/**
+ * For some reason in msys those two macros are not defined correctly in the
+ * header file ws2tcpip.h
+ */
+#ifndef AI_ADDRCONFIG
+#define AI_ADDRCONFIG 0x00000400
```
CFSworks: Might not want to duplicate this ourselves, and especially not in the .h (should do it in a .cxx so we don't affect downstream projects); I see from the MinGW32-w64 headers that this is wrapped in a `(_WIN32_WINNT >= 0x0600)`. We may need to see what @rdb thinks about this.

CFSworks (changes requested):
If the goal is to eliminate `WIN32_VC` and `WIN32`, this commit should really be deleting the block in `dtoolbase.h` that defines them. This would force us to get rid of *all* uses of those macros.

This is going to be a somewhat impactful change. We might want to see if Panda currently compiles/works in Cygwin and (if it does), make sure it continues to work in Cygwin even after the changes are made. (If it doesn't currently work in Cygwin then let's not worry about it, but we might want to decide if bringing back Cygwin support would be a good idea.)

panda/src/downloader/httpDate.cxx
Cerberus1746: Allrighty, I will check if there's a macro that checks if C runtime library is available.

panda/src/downloader/config_downloader.h
Cerberus1746: Ye, the only reason I changed in a .h file is because the two files that were raising errors had references to that file. Changing that sounds easy. But do you want to wait for @rdb opinion before I change it?

Cerberus1746:
> If the goal is to eliminate `WIN32_VC` and `WIN32`, this commit should really be deleting the block in `dtoolbase.h` that defines them. This would force us to get rid of _all_ uses of those macros.

The other PR does that, the big clumping mess one that is #791, for now this PR to make the engine compile in mingw I think it would make a good first step to purge the unnecessary definitions later. 

> This is going to be a somewhat impactful change. We might want to see if Panda currently compiles/works in Cygwin and (if it does), make sure it continues to work in Cygwin even after the changes are made. (If it doesn't currently work in Cygwin then let's not worry about it, but we might want to decide if bringing back Cygwin support would be a good idea.)

I can test if it works with Cygwin. But Msys was the first in my list because it was already installed in my computer. For some reason. I think Ruby installed it...?

Cerberus1746:
Well, I tried to compile panda using cmake and makepanda. Both of them fail.

So or it needs some fixes within cygwin or it doesn't work at all.

panda/src/downloader/config_downloader.h
rdb: I'm fine with putting this in the .cxx file if this is necessary.  However, if MinGW does not define this, are we sure that its getaddrinfo actually supports this flag?

As CFS pointed out in a different comment, don't use `/**` comments here, which are only for docstrings of public interfaces.

Cerberus1746: > I'm fine with putting this in the .cxx file if this is necessary. However, if MinGW does not define this, are we sure that its getaddrinfo actually supports this flag?

It's not getting defined because of `ws2tcpip.h` has this line `_WIN32_WINNT >= 0x0600` which doesn't trigger in my computer. The tests work fine adding the lines above so I assume mingw works with fine if `AI_ADDRCONFIG` and `AI_V4MAPPED` gets defined.

rdb: Ah, okay.  We will be setting `_WIN32_WINNT` to `0x0600` in Panda in about a month, so I don't really mind what hack we use in the meantime.

panda/src/downloader/bioPtr.cxx
```diff
@@ -95,6 +95,17 @@ BioPtr(const URLSpec &url) : _connecting(false) {
     // These hints tell getaddrinfo what kind of address to return.
     struct addrinfo hints, *res = nullptr;
     memset(&hints, 0, sizeof(hints));
+
+// For some reason in msys those two macros are not defined correctly in the
+// header file ws2tcpip.h
```
rdb: Please update comment to imply that these should be removed when _WINNT_VER (or whatever it was called) is changed, so we don't forget about it.

panda/src/nativenet/socket_address.cxx
```diff
@@ -28,6 +28,13 @@ set_host(const std::string &hostname, unsigned short port) {
 
   struct addrinfo hints, *res = nullptr;
   memset(&hints, 0, sizeof(hints));
+
+// For some reason in msys this macro are not defined correctly in the
+// header file ws2tcpip.h
+#ifndef AI_ADDRCONFIG
+#define AI_ADDRCONFIG 0x00000400
+#endif
```
rdb: Please move defines to the top of the file (under the includes), and not mid-code.  Same with the other block.

rdb (changes requested):
I'm happy with this (after requested changes are made) if @CFSworks is.

panda/src/downloader/bioPtr.cxx
kamgha: Could also encapsulate this with `#if _WIN32_WINNT < 0x0600` similar to here: https://github.com/panda3d/panda3d/blob/95f3e30d376a8c3c744cc20d68cbe8b8254db187/dtool/src/dtoolbase/mutexWin32Impl.cxx#L24

Cerberus1746: I don't think it's necessary because the `#ifndef` already makes sure those definitions won't be redefined, also, with other compilers the definitions works fine regardless of the `_WIN32_WINNT` macro and these definitions would be redefined again.

rdb (approved):

dtool/src/dtoolutil/textEncoder.h
```diff
@@ -141,6 +141,7 @@ class EXPCL_DTOOL_DTOOLUTIL TextEncoder {
   static Encoding _default_encoding;
 };
 
+
```
CFSworks: PR should avoid any edits to `textEncoder.h` at all if it's not changing anything of substance. Could you delete this extra newline?

panda/src/downloader/config_downloader.h
```diff
@@ -24,6 +24,18 @@
 #include "configVariableFilename.h"
 #include "configVariableList.h"
 
+/**
```
CFSworks: Docstring-type comment where `//` is more appropriate

CFSworks (changes requested):
These are just nitpicks; the PR itself is good-to-go by me once these are resolved!

CFSworks (approved):

Issue #794, opened 2019-11-27
MacOS: Catalina: Actual application occupies a quarter of the window


el-dee:
As seen on the screenshot below, Panda application on Catalina are confined to the lower left quarter of the window. I guess this is related to the HiDPI scaling of the Retina monitor and that somehow it's no longer applied properly.

<img width="912" alt="Screenshot" src="https://user-images.githubusercontent.com/44778133/69766431-3750a080-1178-11ea-90ff-a0709bcd29f0.png">

el-dee:
Problem is linked with the default value of setWantsBestResolutionOpenGLSurface which seems to have been changed in Catalina, the default mode is HiDPI ON now. Godot and SDL had the same problem, their fix are:

SDL : https://hg.libsdl.org/SDL/rev/46b094f7d20e

Godot : https://github.com/godotengine/godot/pull/32809

rdb:
Could it be that Cocoa is giving the OpenGL window a smaller window size than is actually in existence?  In other words, would it be fixed if we doubled the window size that Cocoa reports?

el-dee:
Not exactly, from what I learned, until Catalina, an app had to specify that it support scaling on the OpenGL frame buffer (backing surface in the Apple lingo). If not activated (the default mode), a pixel in the frame buffer is rendered as a pixel on the screen. If scaling is activated, the frame buffer is magnified or scaled down according to the scaling factor, see https://developer.apple.com/documentation/appkit/nsview/1414938-wantsbestresolutionopenglsurface#declarations

From Catalina onwards, OpenGL app have the scaling automatically activated, even if the app does not support it. Sadly, there is no information from Apple about this change, but you see many OpenGL based application having that change performed the last month due to Catalina.

el-dee:
Even with the fix, the problem occurs if you switch your application to fullscreen. Worse, if you switch back to windowed mode, the window is moved out of the screen and so no longer reachable, though still visible if you ask the desktop to show all the windows of the app.

I also tested a panda application not compiled on Catalina; in windowed mode there is no problem even without the workaround, but the fullscreen issue is present too.

el-dee:
[...]

PR #796, opened 2019-11-28
MacOS: Force disable HiDPI on NSView

el-dee:
Simple fix for #794, HiDPI mode is always disabled when creating the NSView for the current window.

(Note: only tested on Catalina I want to see if it passes the CI, I will test later on Mojave)

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -549,6 +549,11 @@ open_window() {
     _parent_window_handle->attach_child(_window_handle);
   }
 
+  // Always disable application HiDPI support, Cocoa will do the eventual upscaling for us.
+  // Note: setWantsBestResolutionOpenGLSurface method is supported from MacOS 10.7 onwards
+  if ([_view respondsToSelector:@selector(setWantsBestResolutionOpenGLSurface:)]) {
+    [_view setWantsBestResolutionOpenGLSurface:NO];
+  }
   if (_properties.has_icon_filename()) {
     NSImage *image = load_image(_properties.get_icon_filename());
     if (image != nil) {


rdb:
What does this do?  Will the compositor perform upscaling this way or do we get the full, high-res window coordinates and framebuffer this way?

el-dee:
No, this fix keeps the current behaviour, i.e. the framebuffer is pixel exact, one pixel in the buffer is one pixel on screen (except when there is magnification, in that case the rendered window is scaled accordingly).

To enable HiDPI we need to add conversion from screen units to buffer units and vice-versa, which is much more complicated.

rdb:
I'm asking because on Windows we render at double resolution if a double DPI monitor is used (ie. all the coordinates returned by window size are doubled, etc.).  It sounds like you're saying that this is not what happens on macOS, but rather that Cocoa doesn't report the window size in dpi-independent pixels rather than framebuffer pixels and upscales if a HiDPI display is used?  If so, while keeping the current behaviour is probably the right fix for 1.10.5, we may want to ensure that we are consistent between platforms in the future.

el-dee:
> It sounds like you're saying that this is not what happens on macOS, but rather that Cocoa doesn't report the window size in dpi-independent pixels rather than framebuffer pixels and upscales if a HiDPI display is used?

Yes, That's the current behaviour, the app gives its size, but the renderer upscale the buffer according to the actual baking scale factor (but it never not downscale).

What panda does on windows is what would be required if we set wantsBestResolutionOpenGLSurface to ON, but instead of using a hardcoded scaling factor of 2, we have to request to Cocoa the actual scaling factor to use (which depend if we are using a HiDPI display or a standard display)

el-dee:
Tested on Mojave with 10.6 SDK, no regression seen, upscaling is still correctly performed by Cocoa

PR #807, opened 2019-12-11
Eliminate the use of RTTI

rdb:
We currently use RTTI for one purpose alone: giving type names to dynamically registered types in the type system, so that we can define `init_type()` for template types.  Because there is a cost to having RTTI enabled that is not worth the minimal benefit it gives us, we disable this for NDEBUG builds.  This is a problem for ABI compatibility, since it means that the typeinfo symbols are available in the Standard builds, but not in Release.

C++ does not have a standard way to get the type names without RTTI.  There is however a way to get this by using compiler-specific diagnostic information provided in internal variables.  Every compiler makes this available, including MSVC (although we don't need it there, because MSVC is the only compiler that does not require RTTI to be enabled to use `typeid(T).name()`).  The compilers are fairly unlikely to change the format, since they know that a fair number of projects (including, Boost) now rely on this.

This is hacky enough that I want a second opinion on whether this is worth it, or whether we should go for another solution (such as simply enabling RTTI even if we only use it for something so silly).

### DIFF
--- a/dtool/src/dtoolbase/register_type.I
+++ b/dtool/src/dtoolbase/register_type.I
@@ -112,3 +112,44 @@ register_dynamic_type(const std::string &name,
   registry->record_derivation(type_handle, parent4);
   return type_handle;
 }
+
+/**
+ * This is a helper that returns the type name for any given type.
+ */
+template<class T>
+INLINE std::basic_string<char>
+_get_type_name() {
+#if defined(__clang__)
+  static const size_t entire_size = sizeof(__PRETTY_FUNCTION__) - 1;
+  static const size_t prefix_size = sizeof("std::basic_string<char> _get_type_name() [T = ") - 1;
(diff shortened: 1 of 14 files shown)


Moguri:
@CFSworks any thoughts on this?

@rdb any new thoughts on this? Have we characterized the cost of RTTI for NDEBUG builds?

rdb:
IIRC we were both somewhat unhappy with the ugliness of this approach, yet enabling RTTI for such a minor function seems not worth it.  Not exactly sure how to proceed.

Maybe a cleaner way to circumvent this is to just give `init_type()` for templates (like RefCountObj) an argument requiring the full name of the contained type to be given explicitly.

Moguri:
Should we create a discussion issue for this an close this PR until we're ready for it? I would prefer only keep active PRs open. We can still reference this PR from an issue, and we can also re-open the PR if desired.

PR #809, opened 2019-12-12
Window maximize

fireclawthefox:
Integrated the feature of maximizing Panda3D windows from source and config property


rdb:
This looks good, how much testing did you do?  I see you modified the Cocoa bits as well, did you test on macOS?

fireclawthefox:
I mainly tested the Linux build for everything that came to my mind how one could change the maximize state. There was some testing for Windows from another user. I may be able to test some more on Windows later too.
For now there were no one who could have tested the MacOS Version. I don't own this system and back when I asked on irc/discord no Mac users seemed to be there.

rdb:
Thanks!  Merged with fixes.

Issue #811, opened 2019-12-13
Particle Panel: stack trace when cancelling Load Params


xordspar0:
I'm running Panda3D 1.10.4.1 with Python 3.7 on macOS.

When I start to load a file in the particle panel, but then change my mind and click "cancel" in the native file selection dialog that pops up, I get a stack trace:

```
OSError Exception in Tk callback
  Function: <bound method ParticlePanel.loadParticleEffectFromFile of <direct.tkpanels.ParticlePanel.ParticlePanel object at 0x113d83350>> (type: <class 'method'>)
  Args: ()
Traceback (innermost last):
  File "/usr/local/lib/python3.7/site-packages/Pmw/Pmw_2_0_1/lib/PmwBase.py", line 1776, in __call__
    return self.func(*args)
  File "/usr/local/lib/python3.7/site-packages/direct/tkpanels/ParticlePanel.py", line 1278, in loadParticleEffectFromFile
    Filename.fromOsSpecific(particleFilename))
  File "/usr/local/lib/python3.7/site-packages/direct/particles/ParticleEffect.py", line 203, in loadConfig
    data = vfs.readFile(filename, 1)
OSError: Failed to read file: 'None'
```

I edited ParticlePanel.py to add the following line after getting the filename:

```python
        print(repr(particleFilename))
```

askopenfilename is indeed returning the string `'None'`.

For context, here's where the problem is occurring in Panda3d's code:

https://github.com/panda3d/panda3d/blob/2c9d16f62e2a60ae1437d2b725beefcb27c3f55a/direct/src/tkpanels/ParticlePanel.py#L1269-L1281

I draw two conclusions from this:

1. There might be a bug in tkinter and/or the Mac file dialog.
2. The particle panel should check if the file that gets returned by the file picker actually exists. This would also fix race conditions where real files got deleted before we have a chance to read them.

rdb:
I went to the Particle Panel, clicked File > Load Params, and then clicked Cancel, and no error was displayed anywhere.  Did you follow different steps?  Perhaps the Mac version of tkinter is doing something odd by returning "Null" instead of None?

rdb:
Your error message says that it's trying to load "None" instead of "Null", which one is it?

xordspar0:
Oops, my mistake. Replace all places where I said Null with None.

xordspar0:
> no error was displayed anywhere. Did you follow different steps?

[...]

PR #812, opened 2019-12-13
particle panel: catch file not found errors

xordspar0:
tkinter.filedialog.askopenfilename sometimes returns a string 'None'
when the user click cancel in the file picker dialog. This catches
errors like this, where the file is not found.

No error is displayed in the GUI, but the text of the error is printed
to stdout.

This stems from #811.

### DIFF
--- a/direct/src/tkpanels/ParticlePanel.py
+++ b/direct/src/tkpanels/ParticlePanel.py
@@ -1274,11 +1274,14 @@ class ParticlePanel(AppShell):
             parent = self.parent)
         if particleFilename:
             # Delete existing particles and forces
-            self.particleEffect.loadConfig(
-                Filename.fromOsSpecific(particleFilename))
-            self.selectEffectNamed(self.particleEffect.getName())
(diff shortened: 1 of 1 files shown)


rdb:
This seems like the wrong solution; if tk reports a filename of "Null", shouldn't we just check whether it returned that rather than first trying to load a file called Null before giving up?

xordspar0:
1. My mistake, it was actually returning 'None', not Null.
2. How do we know? Maybe there really is a file called None.

rdb:
>     2\. How do we know? Maybe there really is a file called None.

And you think that if such a file were to exist, clicking Cancel on the "open file" dialog should open it?

If askopenfilename is returning None for clicking Cancel, I would wager a bet that if a file called None exists, it would be because someone clicked Cancel on the "Save Params" dialog.

rdb:
I think this should be addressed by 7ca66bfe6f1634abc3f73d843e3318f98d09e278.

PR #814, opened 2019-12-15
Code golf3

BMaxV:
Ok, hi.

This PR has the main goal of removing DirectGuiBase.createcomponent
because it is nothing other than a badly named, badly structured 
__init__/constructor function.

The outcome should be equivalent to the old function, but more
readable.

### how it works

When createcomponent is called, 
it is called by passing a component class,
fetching saved values from a parent/dict saved in a parent,
using those, overwriting the saved args with the new given args
and then in the end using the class' own constructor anyway.

The improvement is sort of obvious, I think:

I have moved some code of createcomponent into a getcomponentargs function
which will fetch previously saved constructor arguments.

I have then replaced all(* see below for details) uses of createcomponent in Direct* definitions
with a call to getcomponentargs and explicit use of the class' constructors

-------------

createcomponent itself doesn't have to be removed. Since all this
PR is doing is shuffling code and changing a few names, but not touching
the logic or outcome, we can keep createcomponent as it is right now. so
we should be 100% backwards compatible.

The only change is that it too uses getcomponentargs to fetch parent/
saved values now.

-------------

Reading the code, some things popped up:

The second and third arguments of createcomponent, component alias and 
component are never not an empty tuple or None. 
It is possible someone has used this in projects, although I doubt it,
because introducing aliases for objects instead of just passing a reference
is just... not very useful I think.

I am still not sure how the parent/old arguments are stored. Which is bad
because that should be more obvious.

--------------

I have not understood how DirectEntry and DirectOptionMenu use createcomponent
and where the difference to my code is. All I know is that they break
when I replace the code. I suspect some nasty side effects somewhere.
So in a way this is a bugreport/issue too.

So this is a TODO that has to be resolved before this change is good for 
merging. If you have the time, please help figure this out.

### DIFF
--- a/direct/src/gui/DirectCheckButton.py
+++ b/direct/src/gui/DirectCheckButton.py
@@ -64,8 +64,8 @@ class DirectCheckButton(DirectButton):
         self.indicator=DirectLabel(self,*args,**kwargs)
         #this is something createcomponent did, not sure why, 
         #let's keep it in place for now.
-        self.__componentInfo[name] = (self.indicator, self.indicator.configure,
-                DirectLabel.__name__, self.indicator.cget, componentGroup)
+        self.__componentInfo[name] = self.indicator
+        self.indicator.group=componentGroup
 
         # Call option initialization functions
         self.initialiseoptions(DirectCheckButton)
--- a/direct/src/gui/DirectCheckButton.py
+++ b/direct/src/gui/DirectCheckButton.py
@@ -47,16 +47,25 @@ class DirectCheckButton(DirectButton):
         self.defineoptions(kw, optiondefs)
         # Initialize superclasses
         DirectButton.__init__(self, parent)
-        self.indicator = self.createcomponent("indicator", (), None,
-                                              DirectLabel, (self,),
-                                              numStates = 2,
-                                              image = self['boxImage'],
-                                              image_scale = self['boxImageScale'],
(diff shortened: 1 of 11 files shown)


Moguri:
@rdb Not looking for a full review here, but any initial thoughts on this change/approach?

@BMaxV Are you still interested in pursuing this PR? If not, or if I don't hear from you in two weeks, I intend to close this PR. It can always be re-opened.

BMaxV:
I'm still interested in pursuing this eventually. I didn't pursue it because other easier to resolve PRs were not merged yet and I didn't want to be pushy with giving other people more work. The code in master is still difficult to read and understand right now, *so it's still a problem* and I would ask you to not close it. 

The problem I see is one of methodology, so it's very much open to discussion and I would like to hear input. I think my comments and changes so far still make sense in the context, so I have little to add for now.

It's higher hanging fruit though. Can you maybe give it a lower priority tag?

Moguri:
I would prefer to close PRs that are inactive to keep the mental inventory down. Perhaps it would be better to close this PR and create a discussion issue that links back to this PR?

rdb:
In principle I don't mind replacing createcomponent with explicit calls, but I actually find the new code in this case harder to read and understand.  I think that refactoring of DirectGUI code should generally have the aim to make the codebase more elegant and 
easier to understand.

PR #816, opened 2019-12-17
Fix compilation warning on macOS 10.6 related to setWantsBestResolutionOpenGLSurface

el-dee:
#797 introduced a (harmless but annoying) compilation warning on macOS 10.6 as setWantsBestResolutionOpenGLSurface is unknown on macOS 10.6.

This replaces the respondsToSelector with a good old conditional compilation activating the code on mac0S 10.7+

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -549,11 +549,11 @@ open_window() {
     _parent_window_handle->attach_child(_window_handle);
   }
 
-  // Always disable application HiDPI support, Cocoa will do the eventual upscaling for us.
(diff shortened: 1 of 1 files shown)


rdb:
Won't this disable the fix for builds that are compiled for 10.6 but run on Catalina?

el-dee:
Yes, but isn't the intent to only use SDK10.9 based build on Catalina ? If not, indeed this PR should be trashed.

rdb:
The SDK build, yes, but the wheel builds for 10.6 are still functional on Catalina.  Even if we made separate 10.9+ wheels, if someone used deploy-ng with the 10.6 wheels, their packaged application would still need to work with Catalina.

If it's just a warning, it's harmless to keep this in, but I would still appreciate having this change in master, though we don't need the `#if` check there (since master does not support 10.6).

el-dee:
Fair point, I forgot about the wheels :( I'll make a quick PR for master then!

Issue #781, opened 2019-11-10
Feature Request: setuptools: allow per platform requirements file


el-dee:
Pip currently does not respect the platform and py_version when checking environment markers (see https://github.com/pypa/pip/issues/6117)

So when including a platform package is needed (like pywin32 or pycocoa), a separate requirements.txt per platform is needed as well as as separate setup.py as one can't specify thee target platform on the command line.

to solve, it should be possible to specify in the setup.py either a requirements.txt per platform (optionally of course) or even be able to specify the content of the requirements.txt file in the setup.y to avoid having several requirements.txt files.

rdb:
I'd rather not add another way to specify requirements if there is already is an obvious way to do it, even if it's bugged, unless we cannot think of any workaround.  Until pip fixes this upstream, maybe we could monkey patch something to have the right thing injected into the `sys.platform` when parsing the `requirements.txt`?

el-dee:
If that works it would be a good workaround.  
There is another use case when setuptools requires a different set of dependencies: when you are compiling c++ modules you need the Panda header files and so the SDK installed ,  so the user requirements.txt file does not contain Panda. But for making a binary you need Panda to be specified in the requirements file. But this case is less problematic,  you only need two requirements.txt and one setup file.

el-dee:
I had a quick look at dist/commands.py to test the workaround but pip is used not as a module but as an external executable so it's a bit less trivial...

rdb:
I'm not sure if there's a particular reason why it's being invoked as executable, but it's probably not hard to fix that.  Another approach is to parse the requirements.txt file ourselves (possiby with help of the pip module that does this) and spit out one for the current platform.

I had a quick look at the pip source and it looks like this might be what we need to override:
https://github.com/pypa/pip/blob/d9e27f0030cfd56451efaf320ad75867be01fd47/src/pip/_vendor/packaging/markers.py#L244

el-dee:
That solution would be much better as we would have control not only on sys_platform but also on all the other environment markers. (and monkey-patching the platform module is I think much harder)

rdb:
(I was proposing monkey-patching the `packaging.markers` module, not sys/platform)

el-dee:
Ok I misunderstand your first proposal :)

Anyway, overriding default_environment() could bee the solution, but we need to check several versions of pip, here I have looked at two pip installation, 18.1 and 19.3.1

[...]

PR #817, opened 2019-12-17
Add pip wrapper to emulate target platform when downloading wheels (WIP)

el-dee:
This is a proof of concept of a solution for #781  I choose the pip wrapper approach as cleaning up pip state each time seems to much of a hurdle (unless there is an easy way to unload a bunch of modules and everything they have created).

Also, it seems there isn't an easy way to extract all the information used by the environment markers from the pep425 platform, so I have to extract and guess most of the information manually.

Lastly, I've done this against release/1.10.x but if it's too dangerous for a maintenance version, I will rebase on master.


rdb:
I don't personally really like a solution this complex but perhaps there is no better solution for this.

Pinging @Moguri for thoughts.

direct/src/dist/pip_wrapper.py
```diff
@@ -0,0 +1,86 @@
...
+
+for (index, arg) in enumerate(sys.argv):
+    if arg == '--platform':
+        target_platform = sys.argv[index + 1]
+
+if target_platform is not None:
+    if target_platform.startswith("win"):
+        sys_platform = 'win32'
+        os_name = 'nt'
+    elif target_platform.startswith("manylinux") or target_platform.startswith("linux"):
```
Moguri: It would probably be safe enough to just do `'linux' in target_platform`

direct/src/dist/pip_wrapper.py
```diff
@@ -0,0 +1,86 @@
...
+    elif target_platform.startswith('macosx'):
+        sys_platform = 'darwin'
+        os_name = 'posix'
+    else:
+        print("Unsupported target platform")
+    for machine in ("x86_64", "amd64", "i386", "i586", "i686", "ppc"):
+        if target_platform.endswith(machine):
+            platform_machine = machine
+            break
+    if platform_machine is None:
```
Moguri: This construct is fine and may want to be kept for readability, but this type of pattern is what for-else is for.

el-dee:
I am personally not also a fan of this workaround,  but I could not find a way to transform back a pip425 tag into platform information,  hence the if-elif hell... 
If the workaround is acceptable,  I will add tests for bsd and also big warning when the code can not infer the platform.

direct/src/dist/pip_wrapper.py
el-dee: Indeed,  will be much clearer :-)

direct/src/dist/pip_wrapper.py
el-dee: I'm not a big fan of for-else but I can use it there if preferred

el-dee:
The code is updated according to the remarks, I also added FreeBSD and OpenBSD platforms as well as a few more exotic machine types.

Moguri:
I don't suppose anything has changed upstream regarding this?

I am okay with giving this work-around a shot (merge it in and see if anyone complains) if we can get this PR updated. I think we can also remove support for very old pip versions (there is some branching around the pip internals change that happened ages ago).

rdb:
I'm on the fence.  It feels fragile and possibly a maintenance time sink.

PR #819, opened 2019-12-21
Add DirectGui events for wheel up and down mouse events

el-dee:
Trivial PR to add WHEELUP and WHEELDOWN events in DirectGuiGlobals to help using the wheel with DirectGui widgets, see https://discourse.panda3d.org/t/directgui-mouse-wheel-scrolling/25429


rdb:
Ah, oops, I feel silly for not noticing that the checks failed before merging.  It looks like the `p3d.` prefix is not meant to be here.

el-dee:
My bad, I had it removed for PGButton but I forgot to remove it for MouseButton... Thanks for the fix :-)

PR #825, opened 2019-12-27
makepackage: Change macOS install location

treamology:
Changes the install location to /Library/Developer/Panda3D. More information on this is in the commit message. I'm pretty sure I caught all of the instances of the old path!

If anyone's interested, discussion on this is in #760.

### DIFF
--- a/makepanda/makepackage.py
+++ b/makepanda/makepackage.py
@@ -484,40 +484,38 @@ def MakeInstallerOSX(version, runtime=False, python_versions=[], **kwargs):
     if (os.path.exists("dstroot")): oscmd("rm -rf dstroot")
     if (os.path.exists("Panda3D-rw.dmg")): oscmd('rm -f Panda3D-rw.dmg')
 
-    oscmd("mkdir -p dstroot/base/Developer/Panda3D/lib")
-    oscmd("mkdir -p dstroot/base/Developer/Panda3D/etc")
-    oscmd("cp %s/etc/Config.prc           dstroot/base/Developer/Panda3D/etc/Config.prc" % outputdir)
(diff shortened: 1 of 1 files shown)


treamology:
Just realized that upgrading from a previous version pre-Catalina will probably cause the installation to fail. I will get a fix pushed for this ASAP. I’m assuming we want to handle this by simply deleting the installation at the old location.

rdb:
Thanks for picking this up.

I was thinking, maybe it would be good to allow setting the install dir in makepackage with an argument?  (Not super urgent to have this done now, though.)

Deleting the installation at the old location sounds OK, but I hope that this doesn't end up deleting things that people may have placed there themselves (which does, unfortunately, occur more than one'd imagine—I've seen users who had their game source code deleted this way because they placed it there).  Is moving to trash an option?

How do you feel about installing a symlink in the old location on pre-Catalina?

treamology:
Moving to trash would probably require an external package. Files on the boot volume that are moved to the trash get moved to `~/.Trash`, while any files on an external disk are stored in `/.Trashes` on that disk, in a subdirectory whose name is some sort of numerical ID that I don't know how is generated. Since Panda allows installations to an external disk, it would probably be more trouble than its worth.

The commit I've already pushed creates symlinks in the old locations.

treamology:
Just added a new commit that removes the old installation, if any. Tested on 10.14 and 10.15, both with and without existing installs.

rdb:
Thanks!  Since the NSIS also accepts an "installdir" parameter, we could probably just use a generic (non-macOS-specific) argument that can be hooked up to the NSIS setting as well later.

I think we should not be writing the script to remove /Developer/Panda3D if someone is passing in `--installdir /Developer/Panda3D` to get the old behaviour, is that a change you can make?

PR #826, opened 2019-12-29
Created some Vector tests

Cerberus1746:
Hello, I created some unit tests for vectors. Mostly are basic ones.


Cerberus1746:
> Looks great, thank you! But, the `sys` and `pytest` imports appear to be unused, so they should be removed.

Done

tests/linmath/test_lvector4.py
```diff
@@ -0,0 +1,69 @@
...
+def test_vec4_swizzle_mask():
+    original_vector = Vec4(3, 5, 1, 0)
+
+    assert original_vector.xy == Vec2(3, 5)
+    assert original_vector.zxy == Vec3(1, 3, 5)
+    assert original_vector.zxyw == Vec4(1, 3, 5, 0)
+
+
+def test_vec4_repr():
+    assert str(Vec4(2, 3, 1, 9)) == "LVector4f(2, 3, 1, 9)"
```
rdb: Ah, this won't work for double-precision builds, where it's LVector4d.

You could instead make two explicit tests, like so:
```python
    assert str(Vec4F(2, 3, 1, 9)) == "LVector4f(2, 3, 1, 9)"
    assert str(Vec4D(2, 3, 1, 9)) == "LVector4d(2, 3, 1, 9)"
```

Also, a test named _repr should probably test repr() and not str(), unless you called it test_vec4_str. ;-)

Cerberus1746: It's test repr because Vector defines `__repr__` not `__str__` :-P

rdb:
LGTM, but this now has conflicts since I merged your other PR.  If you could merge this with master (or rebase), that'd be great, thanks.

Cerberus1746:
> LGTM, but this now has conflicts since I merged your other PR. If you could merge this with master (or rebase), that'd be great, thanks.

Will do

PR #830, opened 2020-01-06
macOS: Add warning in mpkg installer about the change of installation path

el-dee:
Following #760 I added a trivial Readme resource in the distribution.dist file to show a warning before the installation proceeds (and the legacy installation is removed).

Fee free to amend or reword the text :)


el-dee:
This is how it looks in the installer :

Installer README

rdb:
This is great, thank you!

PR #831, opened 2020-01-06
winRawInputDevice: add dualshock support

NathanX-S:


rdb:
Thanks!
Have you tested whether the mapping is all correct with this?
The cross should map to A, the circle should map to B, the triangle should map to Y, and the square should map to X, based on the positions compared to the Xbox controller.

The select button should furthermore map to "back" and the start button should map to, well, "start".

rdb:
By the way, "touchpad" is not currently a supported event name, so I would rather leave that one to 0 until we add a proper event for that.

NathanX-S:
Yep! I compared the index of the pressed button accordingly.

PR #837, opened 2020-01-07
dist: Also use copy_with_dependencies() for executables to flatten their dependencies

### LINKED ISSUE #833: setuptools: Can not integrate CEFPython on macOS
I'm trying to add CEFPython in an app bundle on macOS, but there is one piece of CEF that is not copied properly ```Chromium Embedded Framework.framework/Chromium Embedded Framework```

I added it explicitly in PACKAGE_DATA_DIRS (and discovered a small bug, see #832), 

            ('cefpython3/Chromium Embedded Framework.framework/Chromium Embedded Framework', '', {'PKG_DATA_MAKE_EXECUTABLE'}),

However the file ```Chromium Embedded Framework``` is directly copied under Frameworks/ in the bundle, not as ```Frameworks/Chromium Embedded Framework.framework/Chromium Embedded Framework```

And the wrapping library looks for that relative path :

    cefpython3.cefpython_py37.so:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	/usr/lib/libc++abi.dylib (compatibility version 1.0.0, current version 49.1.0)
	@rpath/Chromium Embedded Framework.framework/Chromium Embedded Framework (compatibility version 31774.13.31, current version 31774.13.31)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)

I'm not sure how to solve this (and hopefully still for 1.10.5)

el-dee:
This is the last modification needed to fix #833. With this modification every file flagged as 'PKG_DATA_MAKE_EXECUTABLE' will be copied using copy_with_dependencies() to detect and ship their dependencies (and flatten their reference in the binary).

With this fix CEF is correctly shipped in the app bundle and works out of the box.

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -642,6 +642,28 @@ class build_apps(setuptools.Command):
         freezer_modules = set()
         freezer_modpaths = set()
         ext_suffixes = set()
+
+        def get_search_path_for(source_path):
+            search_path = [os.path.dirname(source_path)]
+            if use_wheels:
+                search_path.append(os.path.join(p3dwhlfn, 'deploy_libs'))
+
+                # If the .whl containing this file has a .libs directory, add
+                # it to the path.  This is an auditwheel/numpy convention.
+                if '.whl' + os.sep in source_path:
+                    whl, wf = source_path.split('.whl' + os.path.sep)
+                    whl += '.whl'
+                    rootdir = wf.split(os.path.sep, 1)[0]
+                    search_path.append(os.path.join(whl, rootdir, '.libs'))
+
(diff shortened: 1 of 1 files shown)


el-dee:
Oops, corrected !

el-dee:
That's my fear too, though, on the other hand, until now the standalone executables were ignored so their eventual dependencies were not copied and they weren't updated to the new location of the libraries in the app bundle. So the probability that something else gets broken is (hopefully) low.

And I agree with you, there should be a list of reference libraries to be tested to validate deploy-ng

PR #841, opened 2020-01-10
Drop support for Windows XP

### LINKED ISSUE #548: When can we drop Windows XP support?
I realise that there are several people in the community who may still be using and targeting Windows XP, and that there is still a non-negligible audience for Windows XP.  I am creating this issue to collect feedback and get a sense of how long people still expect Panda3D to support Windows XP.

Windows XP does not properly support thread-safe magic statics.  Also, Windows Vista provides proper thread primitives that could really benefit the Panda codebase if we could drop support for Windows XP.  We can probably provide some compatibility layer, but I would want to know that it's worth the effort.

rdb:
This is in anticipation of the imminent dropping of support for Python 2.7 and 3.4 (which are the last Python releases to support Windows XP).  If anyone has questions or objections about this, now is the time to make this known.

The minimum version now becomes Vista.  I can see us moving to Windows 7 as minimum in the foreseeable future feature, since Vista is no longer under extended support, but Vista is apparently still supported by Python 3.8.

Closes #548.


rdb:
Closed in 1bb4a032aa7228dedf381bca6e720f687be419b7, with related changes 4c1373b7213cb1c6921b9848b158f46c63f188be and 2bf60b4361f8c4324e9b39396f51bd214ccd74d5.

PR #847, opened 2020-01-15
device: Fix macOS 10.15 segfault on USB hotplug

treamology:
For some reason, `IOHIDDeviceRegisterRemovalCallback()` no longer works on 10.15+, so an app will crash once trying to poll a device that doesn't exist anymore.  Thankfully, there is the alternative solution of using `IOHIDManagerRegisterDeviceRemovalCallback()`. This just required a little rearranging of the callback code, as well as keeping track of the connection between `IOHIDDeviceRefs` and` IOKitInputDevices` so we actually know which device to remove.

### DIFF
--- a/panda/src/device/ioKitInputDeviceManager.cxx
+++ b/panda/src/device/ioKitInputDeviceManager.cxx
@@ -83,7 +83,7 @@ on_match_device(void *ctx, IOReturn result, void *sender, IOHIDDeviceRef device)
   nassertv(mgr != nullptr);
   nassertv(device);
 
-  PT(IOKitInputDevice) input_device = new IOKitInputDevice(device);
+  IOKitInputDevice *input_device = new IOKitInputDevice(device);
   if (device_cat.is_debug()) {
     device_cat.debug()
       << "Discovered input device " << *input_device << "\n";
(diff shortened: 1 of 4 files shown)


panda/src/device/ioKitInputDeviceManager.cxx
```diff
@@ -78,16 +79,33 @@ IOKitInputDeviceManager::
...
+ * Called by IOKit when an input device has disappeared.
+ */
+void IOKitInputDeviceManager::
+on_remove_device(void *ctx, IOReturn result, void *sender, IOHIDDeviceRef device) {
+  IOKitInputDeviceManager *mgr = (IOKitInputDeviceManager *)ctx;
+  nassertv(mgr != nullptr);
+  nassertv(device);
+
+  nassertv(mgr->_devices_by_hidref.find(device) != mgr->_devices_by_hidref.end());
+  PT(IOKitInputDevice) input_device = mgr->_devices_by_hidref[device];
```
rdb: We don't need to increase the reference count here (we've already asserted that there exists a reference to it, and will be removing it at the end of this method so the PT doesn't serve to extend the lifetime of any reference).  So a regular pointer will suffice.

This is pretty insignificant, but I figured I might mention it while I'm at it: we're also unnecessarily accessing the map twice here.  It'd be more efficient to replace the two lines above with this:
```
auto it = mgr->_devices_by_hidref.find(device);
nassertv(it != mgr->_devices_by_hidref.end());
IOKitInputDevice *input_device = *it;
```

treamology:
> Does Apple guarantee that the add/remove callbacks are only ever called from one thread? If it does, then that keeps things simple. If not, we may have to think about protecting _devices_by_hidref from mutual access.

Since `IOHIDManagerScheduleWithRunLoop()` is being passed `CFRunLoopGetMain()`, the callbacks will only ever come from the main thread.

PR #849, opened 2020-01-18
cocoadisplay: Use hotspot read from .cur files

### LINKED ISSUE #845: OSX mouse "hotspot" in .cur file not being updated
When `.setCursorFilename("cursorFile.cur")` is assigned, the value of the "hotspot" on Windows is carried properly, but in OSX is not updated, leaving the "hotspot" in the top right position.

Have not tested it in Linux.

```python
wp = WindowProperties()
wp.setCursorFilename("crossHair.cur")
base.win.requestProperties(wp)
```

added .cur file with modified hotspot in zip file.

[crossHair.zip](https://github.com/panda3d/panda3d/files/4056632/crossHair.zip)

treamology:
Tested this with setting the cursor through WindowProperties and through a .prc file. Any non .cur file should default to the origin.

Fixes #845

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -568,13 +568,36 @@ open_window() {
   }
 
   if (_properties.has_cursor_filename()) {
-    NSImage *image = load_image(_properties.get_cursor_filename());
+    NSData *image_data = load_image_data(_properties.get_cursor_filename());
+
+    // Read the metadata from the image, which should contain hotspotX and
+    // hotspotY properties.
+    CGImageSourceRef cg_image = CGImageSourceCreateWithData((CFDataRef)image_data, nullptr);
+    NSDictionary *image_props = (NSDictionary *)CGImageSourceCopyPropertiesAtIndex(cg_image, 0, nil);
+
+    CGFloat hotspot_x = 0.0f;
+    CGFloat hotspot_y = 0.0f;
+    if (image_props[@"hotspotX"] != nil) {
+      hotspot_x = [(NSNumber *)image_props[@"hotspotX"] floatValue];
+    }
+    if (image_props[@"hotspotY"] != nil) {
+      hotspot_y = [(NSNumber *)image_props[@"hotspotY"] floatValue];
+    }
+
+    NSImage *image = [[NSImage alloc] initWithData:image_data];
+
     NSCursor *cursor = nil;
-    // TODO: allow setting the hotspot, read it from file when loading .cur.
     if (image != nil) {
-      cursor = [[NSCursor alloc] initWithImage:image hotSpot:NSMakePoint(0, 0)];
+      // Apple recognizes that hotspots are usually specified from a .cur
+      // file, whose origin is in the top-left, so there's no need to flip
+      // it like most other Cocoa coordinates.
+      cursor = [[NSCursor alloc] initWithImage:image
+                                 hotSpot:NSMakePoint(hotspot_x, hotspot_y)];
     }
     if (cursor != nil) {
+      if (_cursor != nil) {
+        [_cursor release];
+      }
       _cursor = cursor;
     } else {
       _properties.clear_cursor_filename();
@@ -582,6 +605,8 @@ open_window() {
     // This will ensure that NSView's resetCursorRects gets called, which sets
     // the appropriate cursor rects.
     [[_view window] invalidateCursorRectsForView:_view];
+
+    [image release];
   }
 
   // Set the properties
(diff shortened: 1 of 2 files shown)


panda/src/cocoadisplay/cocoaGraphicsWindow.mm
```diff
@@ -568,20 +568,45 @@
   }
 
   if (_properties.has_cursor_filename()) {
-    NSImage *image = load_image(_properties.get_cursor_filename());
+    NSData *image_data = load_image_data(_properties.get_cursor_filename());
+
+    // Read the metadata from the image, which should contain hotspotX and
+    // hotspotY properties.
+    CGImageSourceRef cg_image = CGImageSourceCreateWithData((CFDataRef)image_data, nullptr);
+    NSDictionary *image_props = (NSDictionary *)CGImageSourceCopyPropertiesAtIndex(cg_image, 0, nil);
```
rdb: This is a copy operation, yet the properties object is not being released.

To be on the safe side, perhaps we can also check that `image_props` isn't null before accessing it.  I've been bitten before by this kind of thing.

panda/src/cocoadisplay/cocoaGraphicsWindow.mm
```diff
@@ -568,20 +568,45 @@
...
+      }
       _cursor = cursor;
     } else {
       _properties.clear_cursor_filename();
     }
     // This will ensure that NSView's resetCursorRects gets called, which sets
     // the appropriate cursor rects.
     [[_view window] invalidateCursorRectsForView:_view];
+
+    [image release];
```
rdb: Should this be inside a nil check?

panda/src/cocoadisplay/cocoaGraphicsWindow.mm
```diff
@@ -1031,10 +1056,31 @@
       properties.set_cursor_filename(cursor_filename);
       properties.clear_cursor_filename();
     } else {
-      NSImage *image = load_image(cursor_filename);
+      NSData *image_data = load_image_data(cursor_filename);
+
+      // Read the metadata from the image, which should contain hotspotX and
+      // hotspotY properties.
+      CGImageSourceRef cg_image = CGImageSourceCreateWithData((CFDataRef)image_data, nullptr);
+      NSDictionary *image_props = (NSDictionary *)CGImageSourceCopyPropertiesAtIndex(cg_image, 0, nil);
```
rdb: This code is duplicated, perhaps we could create a `load_cursor` method that returned an NSCursor pointer?  Or, failing that, a load_image_data that took the CGFloat hotspot_x/y as reference arguments?

panda/src/cocoadisplay/cocoaGraphicsWindow.mm
```diff
@@ -1287,18 +1335,24 @@
...
     return nil;
   }
 
-  image = [[NSImage alloc] initWithData:data];
-  [data release];
+  _images[resolved] = data;
+  return data;
+}
+
+NSImage *CocoaGraphicsWindow::
```
rdb: This method is missing a comment block.

panda/src/cocoadisplay/cocoaGraphicsWindow.mm
```diff
@@ -568,20 +568,45 @@
   }
 
   if (_properties.has_cursor_filename()) {
-    NSImage *image = load_image(_properties.get_cursor_filename());
+    NSData *image_data = load_image_data(_properties.get_cursor_filename());
+
+    // Read the metadata from the image, which should contain hotspotX and
+    // hotspotY properties.
+    CGImageSourceRef cg_image = CGImageSourceCreateWithData((CFDataRef)image_data, nullptr);
```
rdb: Oh, and `load_image_data` may return null, but if it does, this call will probably crash.

el-dee:
It seems this does not compile with macOS SDK 10.6 : 
<pre>
In file included from panda/src/cocoadisplay/p3cocoadisplay_composite1.mm:5:
panda/src/cocoadisplay/cocoaGraphicsWindow.mm:1424:7: error: expected method to read dictionary element not found on object of type 'NSDictionary *'
  if (image_props[@"hotspotX"] != nil) {
      ^
panda/src/cocoadisplay/cocoaGraphicsWindow.mm:1425:30: error: expected method to read dictionary element not found on object of type 'NSDictionary *'
    hotspot_x = [(NSNumber *)image_props[@"hotspotX"] floatValue];
                             ^
panda/src/cocoadisplay/cocoaGraphicsWindow.mm:1427:7: error: expected method to read dictionary element not found on object of type 'NSDictionary *'
  if (image_props[@"hotspotY"] != nil) {
      ^
panda/src/cocoadisplay/cocoaGraphicsWindow.mm:1428:30: error: expected method to read dictionary element not found on object of type 'NSDictionary *'
    hotspot_y = [(NSNumber *)image_props[@"hotspotY"] floatValue];
</pre>

rdb:
Yeah, I just noticed the buildbot failure.  Would either of [valueForKey](https://developer.apple.com/documentation/foundation/nsdictionary/1410210-valueforkey?language=objc) or [objectForKey](https://developer.apple.com/documentation/foundation/nsdictionary/1414347-objectforkey?language=objc) work instead?  We can leave the syntactic-sugar-version on the master branch.

treamology:
`objectForKey` would be what we would want to use, `valueForKey` is meant for key-value coding on any NSObject.

rdb:
Should be fixed by e011f88f52f099658b65d1368d84f6217145ba0f.  That commit also fixes an issue with `cursor` being uninitialized if `image` is `nil` (which I missed during review).

Issue #854, opened 2020-01-22
setuptools: Can't build binaries with pip >= 20.0


el-dee:
In pip version 20.0 or above, several private functions have been removed, sadly one of them is used by dist/commands.py

Here is the log :
<pre>
Traceback (most recent call last):
running build_apps
  File "setup.py", line 98, in <module>
Building platforms: manylinux1_x86_64
    setup(**config)
  File "/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/site-packages/setuptools/__init__.py", line 145, in setup
Gathering wheels for platform: manylinux1_x86_64
    return distutils.core.setup(**attrs)
  File "/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/site-packages/direct/dist/commands.py", line 390, in run
    self.build_runtimes(platform, True)
  File "/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/site-packages/direct/dist/commands.py", line 550, in build_runtimes
    wheelpaths = self.download_wheels(platform)
  File "/opt/hostedtoolcache/Python/3.7.6/x64/lib/python3.7/site-packages/direct/dist/commands.py", line 411, in download_wheels
    abi_tag = pep425tags.get_abi_tag()
AttributeError: module 'pip._internal.pep425tags' has no attribute 'get_abi_tag'
</pre>

Here is the commit in pypa/pip : https://github.com/pypa/pip/commit/896317d13d9bccf75b9cb35a37a83d93d3d2e55b#diff-542f0dc2284dcb0cb6a0382dfeeb8ed2

And here is the suggested replacement :

https://github.com/pypa/pip/blob/0b397b42c5159942c73a882ca5f33364502e1c3a/src/pip/_vendor/packaging/tags.py#L161

PR #856, opened 2020-01-23
build_apps: Fix running with Pip 20

Moguri:
This is a pretty quick and dirty fix. We may want to explore pip-independent ways to get tags so we can clean all this up. It's quite unfortunate that this showed up right after 1.10.6...

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -401,7 +401,7 @@ class build_apps(setuptools.Command):
 
         whldir = os.path.join(self.build_base, '__whl_cache__')
 
-        pip_version = [int(i) for i in pip.__version__.split('.')][0]
+        pip_version = int(pip.__version__.split('.')[0])
         if pip_version < 9:
             raise RuntimeError("pip 9.0 or greater is required, but found {}".format(pip.__version__))
 
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -405,20 +405,9 @@ class build_apps(setuptools.Command):
         if pip_version < 9:
             raise RuntimeError("pip 9.0 or greater is required, but found {}".format(pip.__version__))
 
-        #TODO find a better way to get abi tag than from internal/private pip APIs
-        if hasattr(pip, 'pep425tags'):
-            pep425tags = pip.pep425tags
-        else:
-            from pip._internal import pep425tags
-
-        if pip_version >= 20:
-            from pip._vendor.packaging.tags import cpython_tags
-            abi_tag = list(cpython_tags())[0].abi
-        else:
-            abi_tag = pep425tags.get_abi_tag()
-
-        if 'u' in abi_tag and (platform.startswith('win') or platform.startswith('macosx')):
-            abi_tag = abi_tag.replace('u', '')
+        abi_tag = 'cp%d%d' % (sys.version_info[:2])
+        if sys.version_info < (3, 8):
+            abi_tag += 'm'
 
         # For these distributions, we need to append 'u' on Linux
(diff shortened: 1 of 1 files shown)


Moguri:
This fixes #854

el-dee:
> It's quite unfortunate that this showed up right after 1.10.6...

To be fair ```pip install --upgrade pip``` installs pip 20.1 only since two or three days. I discovered the problem because the Python action of Github  workflow is doing that when installing it's dependencies, it's seems to not be the case for the Python environment of Travis.

> We may want to explore pip-independent ways to get tags so we can clean all this up

We could either rely on the installed ```packaging```module on the host, or bundle it inside the source of Panda like pip is doing.

el-dee:
> To be fair `pip install --upgrade pip` installs pip 20.1 only since two or three days. I discovered the problem because the Python action of Github workflow is doing that when installing it's dependencies, it's seems to not be the case for the Python environment of Travis.

Actually it's my own workflow that forces the upgrade of pip, not Github action :) So it may be a good idea to add the upgrade in the Travis CI (and workflow in master)

Moguri:
Just to be clear, I was lamenting the unfortunate timing between the release of 1.10.6 and Pip 20. I had no problems with your report, which was quick, descriptive, and much appreciated.

I think relying on `packaging` directly instead of the vendorized version in Pip will ultimately be the way to go. However, I think this would be the first Python dependency that Panda has, and will need to figure out how to deal with that. The wheels should be easy enough (just mark the packages as dependencies), but I am not sure about the SDK installers. Bundling/vendorizing is a possibility, but I believe one that we should avoid since it is not common practice in Python. This may also be something that's better to tackle in `master` versus the `1.10.x` branch.

el-dee:
About bundling, several distributions (like Debian) unbundle packages to use their own version of the package/tool/library (I discovered that when writing the pip wrapper), so it may be a dangerous road too (though the packaging module seems quite stable).

Another possibility would be to either rewrite cpython_tags() or include the implementation of packaging. It's under BSD license, so the code could be included if we keep the license and attribution.

direct/src/dist/commands.py
```diff
@@ -401,14 +401,22 @@ def download_wheels(self, platform):
 
         whldir = os.path.join(self.build_base, '__whl_cache__')
 
+        pip_version = [int(i) for i in pip.__version__.split('.')][0]
```
rdb: I'm not sure why you changed this, since we only are grabbing the first component.

It's legal for a version number to contain letters in the final components, so I worry that this might fail for a release candidate.

Moguri: I first kept pip_version as an array and then changed it to just grab the major version. I agree that this could cause issues with letters in version names and I'll clean it up.

Moguri:
@rdb, any other blockers on this?

Moguri:
@rdb as a heads up, I have forced pushed the changes you suggested. I know GitHub doesn't always notify on force pushes.

direct/src/dist/commands.py
```diff
@@ -399,27 +399,22 @@ def download_wheels(self, platform):
...
-
-        abi_tag = pep425tags.get_abi_tag()
+        pip_version = int(pip.__version__.split('.')[0])
+        if pip_version < 9:
+            raise RuntimeError("pip 9.0 or greater is required, but found {}".format(pip.__version__))
 
-        if 'u' in abi_tag and (platform.startswith('win') or platform.startswith('macosx')):
-            abi_tag = abi_tag.replace('u', '')
+        abi_tag = 'cp%d%d' % (sys.version_info[:2])
+        if sys.version_info < (3, 7):
```
rdb: I hate to be difficult, but this really should be 3.8, or deploy-ng will stop working on Python 3.7 :-)

I can amend it when merging though, the rest looks good, thanks!

Moguri:
@rdb I force pushed the fix.

PR #859, opened 2020-01-25
makepanda: Start a makepanda-compatible wrapper around CMake

Moguri:
This script currently resides at makepanda/makepandawrapper.py but could
be moved to makepanda/makepanda.py if/when the wrapper becomes good
enough to drop makepanda.

Some known issues:
  * Not all package options have been mapped to CMake (e.g.,
    --use-tinydisplay)
  * Not all documented flags have been implemented (e.g., --wheel)
  * Not all flags were documented (e.g., --clean)


Moguri:
It's still pretty rough, but I think this script is in a sharable state, and it should be pretty easy for anyone to extend.

makepanda/makepandawrapper.py
```diff
@@ -0,0 +1,326 @@
...
+    )
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(usage=gen_usage(), prog='makepanda', add_help=False)
+    parser.add_argument('--help', '-h', action='store_true')
+    parser.add_argument('--verbose', action='store_true')
+    parser.add_argument('--tests', action='store_true')
+    # parser.add_argument('--installer', action='store_true')
+    # parser.add_argument('--wheel', action='store_true')
```
rdb: CMake can't build wheels or packages; the idea here should be that we just run makewheel and makepackage on the CMake-generated built dir.

makepanda/makepandawrapper.py
```diff
@@ -0,0 +1,326 @@
...
+    parser.add_argument('--optimize', type=int, choices=(1, 2, 3, 4), default=3)
+    # parser.add_argument('--lzma', action='store_true')
+    parser.add_argument('--distributor', default='makepanda')
+    # parser.add_argument('--outputdir', 'built')
+    parser.add_argument('--threads', type=int)
+    # parser.add_argument('--osxtarget')
```
rdb: `--osxtarget` can map directly to `CMAKE_OSX_DEPLOYMENT_TARGET`.

makepanda/makepandawrapper.py
```diff
@@ -0,0 +1,326 @@
...
+    # parser.add_argument('--override', action='append')
+    # parser.add_argument('--static', action='store_true')
```
rdb: I think this would map to `-DBUILD_SHARED_LIBS=OFF`

makepanda/makepandawrapper.py
```diff
@@ -0,0 +1,326 @@
...
+
+    args = [
+        'cmake',
+        '../',
+    ]
+
+    if sys.platform == 'win32' and platform.architecture()[0] == '64bit':
+        args += [
+            '-A', 'x64'
+        ]
```
rdb: This should be selectable with `--arch`

rdb:
I like this.  I'd like to get this to replace makepanda in the repo once a few more things are supported.

Besides command-line options, I'd like this to also support setting THIRDPARTY_DIRECTORY from the MAKEPANDA_THIRDPARTY environment variable.

rdb:
It'd also be good to set `WANT_PYTHON_VERSION` to the version of Python that makepanda is being invoked with.

makepanda/makepandawrapper.py
CFSworks: CMake *could* be made to [create packages](https://cmake.org/cmake/help/latest/module/CPack.html); that might be worth setting up.

Wheels are an issue though. Maybe a virtual `wheel` target that just kicks makewheel would be in order.

makepanda/makepandawrapper.py
```diff
@@ -0,0 +1,372 @@
...
+
+    # Other CLI args
+    if cli_args.distributor:
+        args.append('-DPANDA_DISTRIBUTOR={}'.format(cli_args.distributor))
+
+    if cli_args.osxtarget:
+        args.append('-DCMAKE_OSX_DEPLOYMENT_TARGET={}'.format(cli_args.osxtarget))
+
+    if cli_args.static:
+        args.append('-DBUILD_SHARED_LIBS=OFF')
```
rdb: Most settings should probably have an `else` case that sets them to the default again.  Otherwise they will "stick" if you pass them and then drop them in a future invocation.

PR #885, opened 2020-03-20
pgui: Ignore PGEntry tab keypress

darktohka:
Pressing Tab on a DirectEntry or simply a PGEntry currently results in the `\t` character being appended to the end of the string. This is, I presume, unwanted behavior.

This commit forces PGEntry and all descendants to ignore Tab keypresses. As such, press events will not be raised for the Tab key anymore, neither will pressing the Tab key play the type sound. I'm not sure when the Tab key started raising events again, but I don't think it did in the past.

If there is a better way to handle this, please mention ideas by all means. Thank you!

### DIFF
--- a/panda/src/pgui/pgEntry.cxx
+++ b/panda/src/pgui/pgEntry.cxx
@@ -203,6 +203,11 @@ press(const MouseWatcherParameter &param, bool background) {
 
       ButtonHandle button = param.get_button();
 
+      if (button == KeyboardButton::tab()) {
+        // Tab. Ignore the entry.
+        return;
+      }
+
       if (button == MouseButton::one() ||
           button == MouseButton::two() ||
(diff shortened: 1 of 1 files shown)


rdb:
I think we should discuss the matter of whether pressing

rdb:
I think we should discuss the matter of whether pressing `<tab>` should even send a keystroke event.

Which OS are you trying this on?

DavidallanC:
I've noticed this as well, it seems to happen on Windows. I didn't encounter this on Linux

Moguri:
Should we close this PR and instead open an issue on `<tab>` sending a keystroke event on Windows?

rdb:
I'll accept this for 1.10.7 as a quick patch, but we do need to investigate the underlying issue.

PR #889, opened 2020-03-28
Remove python2 code from deployment tools

Chitresh13:
Remove bunch of code written in python 2 from dist directory.

### DIFF
--- a/direct/src/dist/FreezeTool.py
+++ b/direct/src/dist/FreezeTool.py
@@ -41,9 +41,6 @@ startupModules = [
 if sys.version_info >= (3, 0):
     # Modules specific to Python 3
     startupModules += ['io', 'marshal', 'importlib.machinery', 'importlib.util']
-else:
-    # Modules specific to Python 2
-    startupModules += []
 
 # These are some special init functions for some built-in Python modules that
 # deviate from the standard naming convention.  A value of None means that a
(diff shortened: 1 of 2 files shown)


direct/src/dist/commands.py
```diff
@@ -209,8 +173,7 @@ def get_data(path):
 del os
 """
 
-SITE_PY = SITE_PY3 if sys.version_info >= (3,) else SITE_PY2
-
+SITE_PY = SITE_PY3 if sys.version_info >= (3,)
```
rdb: This is not actually valid Python syntax.  You don't need the `if` here.  In fact, you don't need the alias.  I suggest just removing this line and changing the original `SITE_PY3` to `SITE_PY`.

direct/src/dist/FreezeTool.py
```diff
@@ -41,9 +41,6 @@
 if sys.version_info >= (3, 0):
```
Moguri: We don't need to branch here anymore and can just set `startupModules` directly.

direct/src/dist/commands.py
Chitresh13: Sure I will sure do it.

Chitresh13: Sure ! I will make those changes.

direct/src/dist/FreezeTool.py
Chitresh13: Ok ! I will set it

rdb:
Superseded by #892

PR #894, opened 2020-03-30
Added testing for presence of valid .ogg files

atulsaswat:
In reference to a valid ogg file referenced from the samples.

### DIFF
--- a/tests/audio/test_loading.py
+++ b/tests/audio/test_loading.py
@@ -1,9 +1,11 @@
 import pytest
+import panda3d.core
+import os
 
 def test_missing_file(audiomgr):
     sound = audiomgr.get_sound('/not/a/valid/file.ogg')
     assert str(sound).startswith('NullAudioSound')
 
-def test_found_file(audiomgr):
-    sound = audiomgr.get_sound('../../samples/music-box/music/musicbox.ogg')
-    assert not str(sound).startswith('NullAudioSound')
+def test_load_ogg_vorbis(audiomgr):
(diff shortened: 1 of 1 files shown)


tests/audio/test_loading.py
```diff
@@ -3,3 +3,7 @@
 def test_missing_file(audiomgr):
     sound = audiomgr.get_sound('/not/a/valid/file.ogg')
     assert str(sound).startswith('NullAudioSound')
+
+def test_found_file(audiomgr):
```
Moguri: Let's rename this to `test_load_ogg_vorbis()` to make adding other file types (e.g., ogg opus) more obvious.

Moguri: This path is relative to the current working directory; we should make it relative to `__file__` using something like:

```python
import panda3d.core
path = p3d.Filename.from_os_specific(os.path.join(os.path.dirname(__file__), '..', '..', 'samples', 'music-box', 'music', 'musicbox.ogg'))
```

Or, maybe a little cleaner:
```python
path = os.path.dirname(__file__) + '../../samples/music-box/music/musicbox.ogg'
```

atulsaswat: Okay, Got it.

Moguri:
@rdb Will the test still fail (as written) if the loader failed to load the audio file? For example, if the codec was not supported?

rdb:
Hi Atul, are you planning to come back to this or should this be closed?

Moguri:
Closing this due to inactivity.

@atulsaswat feel free to re-open this if you're still intending to work on this.

PR #897, opened 2020-04-02
gsg: Fix crashes introduced by 89f1a0e63c

Moguri:
By removing some returns, we could now hit code trying to access empty nodes or otherwise null data.

I do not have a simple repro case for this, but [this project](https://github.com/Kupoman/prototype-mercury) was able to get empty nodes into `GraphicsStateGuardian::fetch_specified_members()`.


Moguri:
This fix may be incomplete/incorrect. The application no longer crashes, shadows stopped working with my custom shaders (`panda3d-simplepbr`). The shadows from the auto-shader still work.

rdb:
Oops.

I think we just need to be adding the return statements back to those four if blocks.  They were removed by accident.

PR #899, opened 2020-04-02
Upated Header guard names to fit convention

FReeshabh:


rdb:
Good change, but the convention is actually not to insert _ between words unless there was a _ in the filename, see `PANDANODE_H` et al.

Moguri:
@rdb Looks like the names got cleaned up. Anything else blocking this PR?

PR #900, opened 2020-04-04
gitignore: ignore Mac OS directory cache

cmarshall108:


rdb:
Sounds good, but I'm curious, why the asterisk?  I thought that the files were always named `.DS_Store` without prefix?

cmarshall108:
Just for the sake of keeping these files out of the git repository. There have been cases for me where the filenames for these cache files get renamed under certain circumstances. You are right however in general.

Issue #440, opened 2018-11-07
Third-party SDKs slated to be dropped


CFSworks:
Many third-party SDKs and packages have been discussed being removed in the near future, so this issue is just to discuss this happening and give users the opportunity to object if they're still using one of these packages.

Note that our development time is limited and we cannot maintain everything on this list. If someone wants to step forward and maintain (i.e. update, write test cases, and field bug reports) one of these, we'll be more than happy to keep support - whether this happens now or later on down the road.

The various packages for which removal is likely are:
- [x] **Awesomium**: The README states that this isn't even built from an official Awesomium SDK, but rather an unmaintained fork somewhere on GitHub. Awesomium itself is no longer an offered product - it has been canceled in favor of something similar called "Ultralight"
- [x] **Carbon**: Removed in 58acc25ea2fb04f6a7018d12235349f7d13e5127.
- [x] **Miles Sound System**: Rad Game Tools doesn't allow their SDKs to be downloaded by anyone who doesn't hold a license. As Panda is a decentralized open source project, that would likely mean that each individual contributor working on the audio code would need a license. Since Panda already supports FMOD and OpenAL, it would likely be better to focus any audio efforts on making those backends as feature-rich as possible.
- [x] **Nvidia PhysX**: As mentioned in [this comment](https://github.com/panda3d/panda3d/issues/387#issuecomment-436218447), the version of the PhysX SDK used by Panda's PhysX code is no longer offered. To justify keeping it, we would need to update it.
- [ ] **SpeedTree**: This one's similar to Miles - it requires a license to test against. The SpeedTree folks seem like they might be more willing to talk to us about doing this, but unless there are any Panda users that really want this feature, it's not worth trying to maintain.
- [x] **libRocket**: Alas it's abandonware. It's a shame, because it was a pretty good GUI toolkit. To continue to support libRocket, we would probably want to fork it ourselves so we can continue to update it. We don't (yet!) have the manpower to maintain forks of other projects, so this one's probably better off being pulled for now.
- [ ] **FFTW**: Although I really want to keep FFT-compression functionality, we should remove FFTW as it doesn't have a permissive license. I personally want to take a hybrid approach: move to FFTS instead of FFTW (which has a much nicer license) and fall back on a hand-coded radix-2,3,5 Cooley-Tukey and DFT implementation when FFTS isn't offered. (I'll be doing this as soon as I have the headroom to write it.)

rdb:
[...]

PR #901, opened 2020-04-05
Remove SpeedTree support

Moguri:
See #440 for a discussion on removing support.

We can still discuss whether or not we actually want to remove SpeedTree support, but I figured I would rip this out real quick in an effort to move #440 along.


Moguri:
Abandoning the PR in favor of #990.

PR #902, opened 2020-04-07
corrects bug: wrong case entered in :Output::MultiplexStreamBuf:write…

JoelStienlet:
add_stdio_file() of MultiplexStream caused an assert to fail in MultiplexStreamBuf.cxx in write_string(), because it entered the OT_ostream case instead of the OT_stdio case.


JoelStienlet:
Sure, but can I write the test in C++ ? Because it looks like the symbol `add_stdio_file` from MultiplexStream is not available from the python interface (as far as I can tell).
There is a directory panda3d/tests/ with python tests only, so I don't know where to put the C++ test.

rdb:
Ah, you're right.  We don't really have a good way to do C++ testing right now, so never mind for now.

PR #910, opened 2020-04-15
pgraph: allow setting user data on PandaNode to associate custom data with nodes

lachbr:


lachbr:
Closing this and opening a new PR because the repository was deleted and I need to make some changes.

PR #916, opened 2020-04-24
dcparser: export extension classes

lachbr:
Needed to access these classes in a separate .dll

### DIFF
--- a/direct/src/dcparser/dcClass_ext.h
+++ b/direct/src/dcparser/dcClass_ext.h
@@ -27,7 +27,7 @@
  * instead of any C++ methods with the same prototype.
  */
 template<>
-class Extension<DCClass> : public ExtensionBase<DCClass> {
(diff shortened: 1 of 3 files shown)


rdb:
Not sure how this is supposed to work, these aren't in libp3direct.dll, but in panda3d/core.pyd.

lachbr:
Do you mean direct.pyd? I link with direct.lib.

rdb:
That's not a supported way to do things, and in fact, we moved away from doing that entirely in the Panda source since 1.10 because of the amount of issues we ran into doing things like that (don't remember the exact reasons right now, but I can think of a couple potential ones).

Marking these EXPORT_CLASS wouldn't be the end of the world, but I'd like to consider whether there are other ways of doing this that would work with your use cases.  Using the Python C API to access these methods, for example.

rdb:
Do you still need this change?

PR #918, opened 2020-04-25
cmake: Build deploy-stub

treamology:
This is a shot in the dark at making CMake build deploy-stub, based on what was being done in makepanda. I'm not sure if this works on Windows/Linux/FreeBSD, so testing would be appreciated.


Moguri:
@treamology Could you please update this to the latest `master` to clear out the CI errors? Have you tested deploying an application with this deploy-stub?

The Linux and Windows stuff looks fine to me, but I don't know enough about macOS to really comment on that chunk.

treamology:
Just rebased onto the latest master.

PR #921, opened 2020-04-26
[webgl-port] wasm/wasi should use .so suffix whereever possible

pmp-p:
see chrome specific preloading at https://emscripten.org/docs/porting/files/packaging_files.html#preloading-files


rdb:
OK, I guess I had changed this wrongly.

PR #924, opened 2020-04-30
add sRGB to egg

Yonnji:


rdb:
I could have sworn we had already done this.  That's a significant omission.

Actually, I wonder if I had originally thought that it would be better to have a separate setting in the .egg file to identify the color space, and have that automatically result in it choosing sRGB format when loaded into an sRGB-aware environment, but be ignored in a regular pipeline.  This would mean that you could author egg assets that would automatically work in both sRGB-correct and regular pipelines.

Other ideas are welcome, but if we don't get anything else, I'll go ahead and merge this anyway so that we have at least *some* way to load in sRGB assets from egg files.

Yonnji:
> choosing sRGB format when loaded into an sRGB-aware environment

This could be useful. And "sRGB-aware environment" could be defined (or enabled/disabled) as option in .prc file.

rdb:
Thinking a bit more about this, I would like a solution that would enable people to keep using existing .egg files without special conversion.

I would suggest adding a setting like this to `<Texture>`:
```
<Scalar> color-encoding { srgb }
```
This would indicate "linear" or "srgb", depending on the transfer function used.  The default setting would be automatically inferred from the envtype (since eg. a normal map is not assumed to be stored gamma-corrected).

Then, there will be a Config.prc setting to indicate that Panda should respect this setting.  By default, Panda would ignore the color-encoding setting, and load everything as linear.  Such a setting might be named something like `egg-respect-texture-encoding true`, `egg-force-texture-linear false` or `textures-srgb true`.

I need to think through the implications of how this would work in .bam files as well, especially with what happens when you load a .bam file that was loaded/cached using a different version of this setting.

In the meantime, other thoughts are welcome.

rdb:
Thinking even more about this… I think I am going about this wrong.  I can only think of these scenarios for a texture applied to an egg model:
1. It is an 8-bit JPEG/PNG image authored for display on a monitor.  This is always in the sRGB color space, using the sRGB transfer function, unless there is a different profile attached to the file.  If it weren't, it would look wrong in the GIMP/Photoshop.
2. It is a floating-point file, such as OpenEXR.  We don't know what color space it is in, but chances are it's linearized sRGB, not gamma-encoded.
3. It is a normal map, or heightfield, or other non-color data.  It contains no colors so cannot be plausibly said to be in a color space or have a color encoding.

All images matching 1 are, as a matter of fact, in the sRGB color encoding.  Sure, a legacy game might not do the whole gamma-dance of decoding before doing lighting and re-encoding before sending it to the monitor, but it's still sRGB-encoded one way or the other, just with less correct lighting.

This is why I'm not comfortable adding something that directly maps to the `Texture.F_srgb` format.  What that format switch does is enable hardware auto-linearization so that users can perform their lighting in linear space.  If users want to do that (which they should), it is a detail of their rendering pipeline, and says nothing about the actual asset or texture, and so that should not be in an .egg file.

Maybe what we really need is an enum to indicate whether the texture contains color data or other data, such as normals or a 
heightfield (and only if `envtype` is inadequate to determine this)?  At load time, or texture upload time, Panda should consider whether the texture needs to be decoded before read based on the specifics of the rendering pipeline.

How people configure *that* is something I'm not yet sure about.  We will have to come up with a way to do so that minimizes the chance for users to shoot themselves in the foot.  I keep thinking that there should be some global setting (or GSG-specific setting) to indicate that you would like to do framebuffer operations in linearized color space (ie. `framebuffer-srgb`, but not so confusingly named), and that Panda would automatically enable the sRGB transfer function on "color"-tagged textures.  On the other hand, this means Panda chooses the texture format instead of the user, a departure from the close-to-the-metal philosophy of Panda.

Separately, there is the issue of vertex/polygon/material colors.  We are taking floating-point colors and not gamma-encoding them before writing them to the framebuffer.  Therefore these are effectively sRGB colors but specified in the 0..1 scale.  However, by convention, floating-point colors are supposed to be linear, so I'm not quite comfortable with subjecting these to the same treatment.  On the other hand, the fact that we're not converting vertex colors and light colors is going to be the biggest source of pain for people who want to switch on gamma-correctness.

I would suggest closing this in favor of a couple of new discussion issues on each of these matters.  Thoughts and opinions are always welcome, of course.

rdb:
Based on subsequent discussion I went ahead and merged this to 1.10.7 so that we can have round-trippability with .bam files, but with the caveat that this shouldn't be used by exporters—this should not become the recommended way of getting textures loaded as srgb.  But it doesn't really hurt to have this as an option until we have a better solution, at which time we can deprecate this and F_srgb at the same time.

Thanks for this patch!

PR #925, opened 2020-05-01
shadergen: Make use of material alpha if present

### LINKED ISSUE #912: ShaderGenerator ignores material alpha
Hi,

[This forum topic](https://discourse.panda3d.org/t/material-with-transparency-working-in-pview-but-not-in-game/25918) has revealed an apparent issue with material alpha being disregarded by the ShaderGenerator, where the fixed-function pipeline does seem to do the expected thing.

It does not seem to be related to any of the other ShaderGenerator/alpha issues I reported in the past (although it could be, even if only indirectly), hence this new report.

Anyway, here is some code that should create some transparency in this matter:

```python
from panda3d.core import *


class MyApp(ShowBase):

    def __init__(self):

        ShowBase.__init__(self)

        dir_light = DirectionalLight("light")
        dir_light.set_color((1., 1., 1., 1.))
        light_np = self.render.attach_new_node(dir_light)
        light_np.set_hpr(-20., -20., 0.)
        self.render.set_light(light_np)
        self.render.set_shader_input("light", light_np)
        # activate the ShaderGenerator
        self.render.set_shader_auto()

        model = self.loader.load_model("smiley").get_child(0)
        model.set_y(4.)

        # create a copy of the smiley model to show through the transparent model
        model.copy_to(self.render)

        model.reparent_to(self.render)
        model.set_y(3.)
        # clear the render state of the model to remove its texture
        model.node().set_geom_state(0, RenderState.make_empty())
        mat = Material()
        mat.set_diffuse((0., 1., 0., .5))
#        model.set_alpha_scale(.5)
        model.set_material(mat)
        model.set_transparency(TransparencyAttrib.M_alpha)

app = MyApp()
app.run()
```

These are the results that I am getting:

* using the ShaderGenerator: a fully opaque sphere;
* using the FFP: the sphere is transparent such that the smiley behind it shows through.

Workarounds include setting an explicit alpha scale or transparent texture on the model, although I'm not convinced that either of these should effectively be a requirement.

Tested with Panda3D version 1.10.6 on Windows 10.

Moguri:
The material alpha overrides alpha values from flat color or vertex
colors. This follows what the fixed-function pipeline does.

Fixes #912

### DIFF
--- a/panda/src/pgraphnodes/shaderGenerator.cxx
+++ b/panda/src/pgraphnodes/shaderGenerator.cxx
@@ -1485,7 +1485,9 @@ synthesize_shader(const RenderState *rs, const GeomVertexAnimationSpec &anim) {
     // Combine in alpha, which bypasses lighting calculations.  Use of lerp
     // here is a workaround for a radeon driver bug.
     if (key._calc_primary_alpha) {
-      if (key._color_type == ColorAttrib::T_vertex) {
+      if (key._material_flags & Material::F_diffuse) {
+        text << "\t result.a = attr_material[1].w;\n";
+      } else if (key._color_type == ColorAttrib::T_vertex) {
         text << "\t result.a = l_color.a;\n";
       } else if (key._color_type == ColorAttrib::T_flat) {
         text << "\t result.a = attr_color.a;\n";
--- a/tests/display/test_color_buffer.py
+++ b/tests/display/test_color_buffer.py
@@ -347,3 +347,17 @@ def test_color_transparency_vertex(color_region, shader_attrib, light_attrib):
     )
     result = render_color_pixel(color_region, state, vertex_color=(1, 1, 1, 0.5))
     assert result.x == pytest.approx(0.75, 0.1)
+
+
+def test_color_transparency_no_light(color_region, shader_attrib):
+    mat = core.Material()
+    mat.diffuse = (1, 1, 1, 0.75)
+    material_attrib = core.MaterialAttrib.make(mat)
+
+    state = core.RenderState.make(
+        core.TransparencyAttrib.make(core.TransparencyAttrib.M_alpha),
+        shader_attrib,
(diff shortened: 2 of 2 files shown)


rdb:
It would be great to add a test for this to `tests/display/test_color_buffer.py`, especially if we're going to put this in a 1.10 release.

Moguri:
Updated this to add tests and follow what the fixed-function pipeline does (material alpha overrides other alpha values).

rdb:
Hmm… if we're matching FFP behaviour, that's great, but I'm wondering whether this might break someone's application if they are relying on the current shader generator.  Thoughts?

Moguri:
I cannot think of a use case (other than accidental) where a user would have a material alpha applied and be fine with the object not rendering as transparent. However, we do know at least one person has run into this issue because #912 was filed. I would be okay with rebasing this on `master` if that makes things easier. I have no strong need for this fix (I do not use the shader generator); I was just trying to clear out an issue.

@Epihaius do you have any thoughts on this?

Epihaius:
> I cannot think of a use case (other than accidental) where a user would have a material alpha applied and be fine with the object not rendering as transparent.

Indeed, it seems weird to me that anyone would consider the current behaviour of the shader generator as normal, and even more absurd if they thought it was actually a feature. Why go through the trouble of explicitly setting a material alpha to a value lower than 1.0 if the expected result is that it won't have any effect?

rdb:
I'm only wondering about the accidental case.  Maybe it's fine, but it does raise the more general question about whether renderer behaviour changes are OK for bugfix releases.

Moguri:
I have updated the branch to include a test without a light. With and without the auto-shader, the pixel renders as black (even with a non-black diffuse material color). I am not sure if I missed something in the test setup (I just removed the LightAttrib).

TEST_COLOR does have alpha.

Issue #523, opened 2019-01-10
Feature Request: .ini configuration files


wezu:
The .ini file format is almost a standard way to write configuration files. For game. On Windows.  Usually.

Players faced with a .ini file have at least a chance to recognize it as a configuration file, and probably know what to do with it (I think even W10 by default opens ini files with the Notepad). 

Panda3D on the other hand uses .prc files, and unless players used P3D before or read some instructions, they will not even know it's a configuration file.

Why not use .ini files as an alternative? A simple solution would be putting a `load_ini_file_data` function  in panda3d.core that would do:
```
def load_ini_file_data (file_name):
    #import configparser
    Config = configparser.ConfigParser()
    Config.read(file_name)
    #builtins.Config = Config #add it to buildins?
    for section in Config.sections():
        for option in Config.options(section):
            #read all options as prc_file_data in case they have p3d meaning
            load_prc_file_data("", option +' '+Config.get(section, option))
```

CFSworks:
Panda3D definitely has some heavy [NIH syndrome](https://en.wikipedia.org/wiki/Not_invented_here) with its file formats (.mf, .prc, .egg, .pz, .dc, ...), though in Panda's defense many of these were developed before the creation of the more industry-standard alternatives. Still, in the interest of modernization, we should work on supporting newer formats that beginners (or beginners' systems) can recognize immediately.

.ini's section-based organization might not fit in very well with how the PRC system is already organized - it makes it harder to change config variables from e.g. an in-engine developer console, for example. It's also not likely to help with the confusion on non-Windows platforms. But for sure, .prc is a bizarre file extension. Maybe adding .cfg to the standard search path and discouraging .prc would go a long way toward clearing up that confusion? Nonetheless, we should definitely have a broader discussion about how we want the PRC system to work (I don't like that it has some variables that don't take effect if changed after static-init time, how `#t`/`#f` is the syntax for true/false despite looking like a comment, how it's ambiguous whether an additional line is adding an entry to a list vs. modifying the previous value, etc.).

What are your thoughts?

Moguri:
For the sake of completeness, I am mentioning #245 and #269 here. Maybe we need to a config/prc label?

rdb:
I'm open to this as long as we reach concensus on how to deal with sections, and as long as we can do all this without having to maintain two config systems side-by-side.  I would be against a duct tape solution like the snippet in @wezu's issue report.

[...]

PR #926, opened 2020-05-01
config: Allow equal signs (=) between variable names and value

Moguri:
This allows for more INI-style syntax as proposed by #523. Sections are
still not supported.

### DIFF
--- a/dtool/src/prc/configPage.cxx
+++ b/dtool/src/prc/configPage.cxx
@@ -397,7 +397,7 @@ read_prc_line(const string &line) {
   EVP_VerifyUpdate((EVP_MD_CTX *)_md_ctx, line.data(), line.size());
 #endif  // HAVE_OPENSSL
 
-  // Separate the line into a variable and a value.
+  // Parse leading whitespace
   size_t p = 0;
   while (p < line.length() && isspace((unsigned char)line[p])) {
     p++;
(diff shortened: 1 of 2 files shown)


rdb:
Interesting.  I think this PR would allow `var = = value`, though?

I think the is_split_character, which is not declared externally, should be marked `static` (since it's not declared elsewhere, no point in giving it external linkage), and hoisted to the top of the .cxx file by Panda convention.

Do we want to enable this optionally, as part of a special mode for now (possibly toggled by extension), or do we want to allow this syntax for all .prc files?

Moguri:
Yes, it allows any combination of spaces and equal signs. I figured this would be the simplest approach and PRC isn't very strict as is (e.g., anything start with "t" being "true").

I'll fix up the scope of is_split_character() and move it.

Moguri:
I forced pushed a new version that only allows one equal sign.

Moguri:
@rdb Is this PR something we want? What is missing to move this along? More discussion?

rdb:
One question I raised for discussion remains unaddressed, and I'd like to carefully consider that before we make this otherwise irreversible change.

Moguri:
I am going to abandon this PR for now and we can continue discussion in #523. I started this PR in the hopes of quickly getting an issue closed, but I am otherwise not strongly invested in the feature. The code will still be around for someone to pick up.

PR #932, opened 2020-05-02
test for how to write tests to increase coverage

BMaxV:
Hello, I noticed there is now a test coverage tool integrated in CI, and I probably can write test?

Not sure. I looked at the source in the test coverage tool and if I understood it correctly, the methods are defined but never executed in tests.

This very simple test should test if the constructor and that pretty print function can be executed?

### DIFF
--- a/tests/actor/test_actor.py
+++ b/tests/actor/test_actor.py
@@ -1,0 +1,9 @@
+import pytest
+from direct.actor.Actor import Actor
+
+def test_actor_bare_init():
+    Actor()
+    
+def test_actor_pprint():
+    A=Actor()
+    A.pprint()


BMaxV:
Ok so ignoring what the super detailed report says, the check says code coverage increased by 0.05% so that's something?

rdb:
The line coverage testing (which still isn't quite working well) is meant to give a rough estimation of which parts of the Panda codebase are covered by tests, by checking which lines of code are being hit when running the test suite.  It can *not*, however, say how well the *behaviour* of these lines are being tested.  It doesn't know whether you've actually created asserts to verify the behaviour, or whether you just created a test suite that runs those methods and doesn't test their results.

Of course, testing that constructing an empty Actor doesn't crash Panda arguably tests *something* and is better than nothing, but the intent should never to be to just write the minimal test to make codecov *think* it's covered.

BMaxV:
Yes. My idea was to clear some low hanging fruit with methods that don't do anything complex and/or should "just work", like the actor constructor, which provides good defaults for all cases.

At least in the python parts that I know, I expect a lot of getter and setter functions that really should work almost trivially and should be easy to test. And it should be obvious if it's a "good" test too.

rdb:
OK, I'm quite happy to entertain these kinds of pull requests, though testing that an empty Actor constructor doesn't crash is the absolute minimum we can test about this.  This would make codecov tick off the Actor constructor code as tested but without actually testing the most common ways to use the Actor constructor, nor that the Actor constructor actually did its job correctly.

If there were tests that used the various forms of the Actor constructor and did some testing on its behaviour, I would be happy to consider that.

BMaxV:
Ok so I thought about this a bit. 

I added two tests that perform the same action as is described in the tutorial, so there is that bit of functionality. But that does not address the question whether the function is actually performed, or the code just runs without errors.

I.e. it does not test if the rendering would work. But then I don't know how to test that or if you usually test that.

And it should also be difficult to test other things for basic stuff without also testing the init?

PR #940, opened 2020-05-21
Roaming Ralph updates

Moguri:
The biggest change here is switching the Roaming Ralph sample to using CollisionHandlerFloor to handle collisions with the ground. However, CollisionHandlerFloor was not working as expected, so I have added a fix, which is in PR #957.

The other changes are small code cleanups.


rdb:
Could you split this PR up into two separate parts?  Thanks!

Moguri:
The CollisionHandlerFloor fixes have been moved to their own PR (#957)

Moguri:
I am abandoning this PR

Issue #707, opened 2019-08-16
p3d_LightSourceParameters shadowMap not universal.


lettier:
Hello,

The [documentation](https://www.panda3d.org/manual/?title=List_of_GLSL_Shader_Inputs) states that the shadow map in `p3d_LightSourceParameters` is a `sampler2DShadow`. 

However, it is a [cube map](https://github.com/panda3d/panda3d/blob/daa57733cb9b4ccdb23e28153585e8e20b5ccdb5/panda/src/display/graphicsStateGuardian.cxx#L3304) for point lights. 

When using a custom GLSL shader, that uses the shadow map for both point and non-point lights, the following [error](https://github.com/panda3d/panda3d/blob/bc563c86643df9fc228c493c81d58a71d7f5fda3/panda/src/glstuff/glShaderContext_src.cxx#L2817) is continuously reported. 

```c
:display:gsg:glgsg(error): Sampler type of GLSL shader input p3d_LightSource[...].shadowMap does not match type of texture dummy-shadow-cube.
````

If you do set the point light as a shadow caster, the shadow map is still a [cube map](https://github.com/panda3d/panda3d/blob/daa57733cb9b4ccdb23e28153585e8e20b5ccdb5/panda/src/display/graphicsStateGuardian.cxx#L3338).

Code such as this [example](https://discourse.panda3d.org/t/shadows-direction-parameters-and-inputs-in-glsl/15300/4) is not universal across all light types. Using it with a point light would give you the error. 

Other than the shadow map, `p3d_LightSourceParameters` was made to be universal. For example, the `w` component being zero for directional lights or the spotlight cutoff being [-1](https://github.com/panda3d/panda3d/blob/daa57733cb9b4ccdb23e28153585e8e20b5ccdb5/panda/src/display/graphicsStateGuardian.cxx#L1705) for non-spotlights. 

Can a special struct be made just for point lights? I understand the need for six views given that a point light has no direction. Otherwise, can a dummy 2D texture be used for point lights with the expectation that point lights produce no shadows as per the [documentation](https://www.panda3d.org/manual/?title=Lighting#Shadow_Mapping)?

As of right now, you must either:

1) Use only point lights if you want shadows (doesn't seem to work).
2) Use point, directional, and spot but without shadows.
3) Use only directional and spot if you want shadows.

:+1:

rdb:
[...]

PR #942, opened 2020-05-27
gsg: Match dummy shadow map types with the type desired by the shader

Moguri:
This is a partial fix for #707. This does *not* allow for both cube map (point light) and 2d (everything else) shadows at the same time, but it does prevent an error if point lights without shadows are being used in combination with 2d shadows.

No function signatures were touched, so this seems like something we could put into a 1.10.x release.

This was tested using simplepbr and an example scene with directional lights (one setup as a shadow caster) and a point light (not a shadow caster).

### DIFF
--- a/panda/src/display/graphicsStateGuardian.cxx
+++ b/panda/src/display/graphicsStateGuardian.cxx
@@ -1861,6 +1861,8 @@ fetch_specified_texture(Shader::ShaderTexSpec &spec, SamplerState &sampler,
   case Shader::STO_light_i_shadow_map:
     {
       const LightAttrib *target_light;
+      Texture::TextureType desired_type = (Texture::TextureType)spec._desired_type;
+      PT(Texture) tex = nullptr;
       _target_rs->get_attrib_def(target_light);
 
       // We don't count ambient lights, which would be pretty silly to handle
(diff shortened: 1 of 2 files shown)


Moguri:
Since the 1.10.7 milestone was added, I went ahead and rebased this on `release/1.10.x`.

rdb:
Thanks.  Do I understand this right that this PR basically allows a mismatch between the shadowMap sampler type and the actual light type if the light in question doesn't have shadows enabled?

Moguri:
Yes. If the light does not have a shadow enabled, the sampler type will match what the shader requested instead of an appropriate type for the light. This keeps shaders from breaking in the case where a user has shadow casting spot/directional lights and non-casting point lights, which seems -- to me -- to be a common setup. Mixing 2D casters with cube casters will still result in errors.

panda/src/display/graphicsStateGuardian.cxx
```diff
@@ -1861,30 +1861,33 @@ fetch_specified_texture(Shader::ShaderTexSpec &spec, SamplerState &sampler,
...
+        bool is_point = light.node()->is_of_type(PointLight::get_class_type());
         nassertr(!light.is_empty(), nullptr);
         Light *light_obj = light.node()->as_light();
         nassertr(light_obj != nullptr, nullptr);
 
-        PT(Texture) tex = get_shadow_map(light);
-        if (tex != nullptr) {
-          sampler = tex->get_default_sampler();
+        if ((is_point && desired_type == Texture::TT_cube_map) ||
+            (!is_point && desired_type == Texture::TT_2d_texture)) {
```
rdb, line 1879: If I understand correctly, this isn't going to give an error, but going to silently assign a dummy texture when the light doesn't use shadow casting.  Should this instead be a check to see whether it's a caster?  I'd suggest passing the desired type into get_shadow_map, but then this can't go into 1.10.

rdb:
I have pushed a more minimal implementation of the described behaviour as 40962e1eb5c7c4d5ae72f212ba351448b016f2a1.  Wheels are available.  It would be great if you could test to see if it works with your use case.

Moguri:
I can confirm that your fix works for my use case, thanks!

PR #943, opened 2020-05-27
makepanda: Add pyenv support

Moguri:
Only tested on Linux with Clang.

If this is not desired, feel free to just close the PR.

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -1972,6 +1972,10 @@ def CompileLink(dll, obj, opts):
             if GetOrigExt(x) != ".dat":
                 cmd += ' ' + x
 
+        pyenv_prefix = GetPyenvPrefix()
+        if pyenv_prefix is not None:
+            cmd += ' -L' + os.path.join(pyenv_prefix, 'lib')
+
         if (GetOrigExt(dll) == ".exe" and GetTarget() == 'windows' and "NOICON" not in opts):
             cmd += " " + GetOutputDir() + "/tmp/pandaIcon.res"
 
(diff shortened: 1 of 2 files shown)


Moguri:
Rebased the changes and fixed them to work on Python 3.8. Sorry for taking so long to fix this!

rdb:
Is there a way to detect that makepanda is actually being run using a pyenv version of Python, as opposed to a different one, and get the prefix for the one it's run with?

Just running `pyenv prefix` does not seem to be very reliable, for example:
```
❯ /home/rdb/.pyenv/versions/3.9-dev/bin/python3.9 -c 'import subprocess; subprocess.call(["pyenv", "prefix"])'
/usr
❯ pyenv global 3.9-dev
❯ python3.8 -c 'import subprocess; subprocess.call(["pyenv", "prefix"])'
/home/rdb/.pyenv/versions/3.9-dev
```

Moguri:
I would like to be able to use pyenv to switch between making Python 3.7 and 3.8 builds of 1.10. With makepanda is on its way out thanks to the CMake efforts, am not looking for a "perfect" solution for pyenv. Covering basic use cases is fine for me. However, I *do* want to make sure I do not cause regressions.

One, potentially more disruptive solution is to use `sysconfig.get_paths()` ([docs](https://docs.python.org/3/library/sysconfig.html)) to get paths for all builds (i.e., not just ones using pyenv). `platinclude` and `platstdlib` appear to point to the correct places (might need to strip a final `python3.x`) when used with pyenv.

Moguri:
@rdb Any thoughts on this? I do not have a particular need for this anymore since I no longer need to target multiple Python versions. I am tempted to just drop this and hope focus on CMake and 1.11.

Moguri:
Abandoning for now

PR #948, opened 2020-06-03
dgui: add DirectCheckBox import to DirectGui module

kamgha:
This has been missing there.

### DIFF
--- a/direct/src/gui/DirectGui.py
+++ b/direct/src/gui/DirectGui.py
@@ -24,5 +24,6 @@ from .DirectSlider import *
 from .DirectScrollBar import *
 from .DirectScrolledFrame import *
 from .DirectCheckButton import *
+from .DirectCheckBox import *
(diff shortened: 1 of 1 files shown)


rdb:
I'm curious, by the way, does DirectCheckBox do anything that DirectCheckButton doesn't?

fireclawthefox:
> I'm curious, by the way, does DirectCheckBox do anything that DirectCheckButton doesn't?

They both have a slightly different setup. The check box requires you to set an image for it's checked an unchecked state whereas the check button automatically sets up a box with a hardcoded * to represent the checked state positioned next to the buttons actual text. The check box can have a text too, but I think usually it won't.

kamgha:
I'd like to take this PR back. Removal of the module could be considered, perhaps into a sample project demonstrating DirectGUI and modification of widgets by inheritance.

PR #950, opened 2020-06-08
RecastDetour Integration

ash1215:
# Google Summer of Code 2020 
### Organisation: Python Software Foundation | Sub-Organisation: Panda3D
This is the Pull Request for the integration of Recast Detour tools into Panda3D. It is part of my **Google Summer of Code 2020** Project. My proposal for the same can be found here: [GSoC Proposal](https://blogs.python-gsoc.org/media/proposals/Panda3D_GSoC_Proposal.pdf)
Recast Detour tools are advanced pathfinding tools used in game development. Recast builds a navigation mesh, which is the geometry containing all the walkable surfaces. Detour helps in finding path between source and destination points avoiding obstacles over that navigation mesh.
For the integration, we have divided the Recast Detour tools into two lilbraries: **navmeshgen** and **navigation**.

### navmeshgen

- **NavMeshBuilder:** NavMeshBuilder class contains all the vertices and triangles. It also has the functions to build the navigation meshes using those vertices and triangles. Set the properties of the actor and environment using this class  and then build it to get a NavMesh object as output.

### navigation

- **NavMeshParams:** NavMeshParams class stores all the parameters of a navigation mesh.
- **NavMesh:** NavMesh class stores the navigation mesh. The navigation mesh can be obtained using NavMeshBuilder class or can be generated using the NavMeshParams class by the user.
- **NavMeshNode:** NavMeshNode class is a child class of PandaNode which stores the NavMesh object.
- **NavMeshQuery:** NavMeshQuery class contains the functions to query the navigation mesh stored in a NavMesh object. The queries include functions to find the paths from one point to another over the mesh.

Sample code is also provided in the `samples/navigation` directory.

panda3d

The above gif shows the successful integration of the Recast Detour tools into Panda3D world. You get the same output on running the sample code. Press `Left-Mouse-Button` to set/change destination. Additionally, press `s` on keyboard to show the navigation mesh (in blue color) and press `h` to hide the navigation mesh.
We have also prepared the python and c++ docs for the same. The Pull Request for the Panda3d-Docs is available here: [#71](https://github.com/panda3d/panda3d-docs/pull/71)

As of the end of GSoC 2020 coding period, the integrated code can be compiled using makepanda. To list down the points which require to be done in future and which are completed during the GSoC period, it looks like follows:
- [x] Implementing the code into two libraries: navmeshgen and navigation
- [x] Generating navigation mesh
- [x] BAM Serialization
- [x] Exposing the three main dtNavMeshQuery query functions: `find_path`, `find_straight_path` and `find_nearest_point`
- [x] Writing Python and C++ Documentation
- [x] Implementing navigation mesh visualisation for debugging
- [x] Writing sample code
- [x] Writing unit test files
- [x] Addressing and resolving all PR reviews
- [x] Adding MakePanda support
- [ ] Adding CMake support
- [ ] Exposing more dtNavMeshQuery queries
- [ ] Adding Area costs

The last commit before the final evaluation is : [36986d9](https://github.com/panda3d/panda3d/pull/950/commits/36986d961e2c8b3506fddd0c8a5523c81b4c078f)

Special thanks to @rdb and @Moguri for being the best mentors and helping out in every situation.
To know more about the project, you can follow the comments and the commits in this Pull Request as well as the Pull Request [#71](https://github.com/panda3d/panda3d-docs/pull/71) for Panda3D-docs.

To know about my GSoC 2020 journey, read my blogs:
- [Python GSoC Weekly Blogs](https://blogs.python-gsoc.org/en/imaj_ashwinis-blog/)
- [Panda3D Discourse Thread](https://discourse.panda3d.org/t/recast-and-detour-integration-in-panda3d/26066)

_This comment is updated as of 30th August for the final evaluations of GSoC 2020._


Moguri:
I still need to go over the new classes, but, for now, I will leave some comments on the build system changes. We already went over these on Discord, but I figured I'd repeat them here to they don't get lost.

* Recast and Detour upstream files should be in their own directories following upstream (e.g., `Recast`, `Detour`)
* I would prefer the library name (currently `p3recastdetour`)  to not contain the names of the dependencies we're using. This is an implementation detail that we don't need to leak to the user. `p3navigation` or `p3navmesh` sound better to me. We may also want to split this into two libraries so navmesh building doesn't have to be loaded at runtime if not required.
* We do not need `DebugUtils` nor `DetourTileCache` at this time, so we can skip including and building them for now.
* I see you put this into `panda3d.recastdetour` instead of `panda3d.core`. Is this the route we want to go? I think, if we want this as a plugin or two, then the answer is probably yes.

I see you've been using a modified Roaming Ralph to do some tests with, but we'll probably want to start up a new sample soon. This will make it easier to review the usage and API of these new features. Also, I'm sure users would appreciate an example.

ash1215:
@rdb I will surely make the required changes as soon as possible. Also I do understand that proper code styles need to be maintained at all levels. I will take care about the points mentioned. Since, we are just one week down, it will be easier for me to make the changes. 
@Moguri I will organised the recastdetour codes into recast and detour folders just as they were in the original repository and make the changes required in makepanda.

NathanX-S:
I'm really excited to see where this goes. Best of luck to all involved.

ash1215:
@rdb @Moguri I have updated the draft PR. the folder name is now "panda/src/navigation". @rdb I have taken care of your suggestions except for the docstring thing. Will add that as well once the idea of functions and classes gets finalized. @Moguri Only the recast and detour folders from recastdetour codebase are imported. I have made the necessary changes in makepanda as well. Please suggest if anymore changes are required.

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,164 @@
...
+  virtual void collect_settings(struct BuildSettings &settings);
+
+  float get_agent_radius() { return _agent_radius; }
+  float get_agent_height() { return _agent_height; }
+  float get_agent_climb() { return _agent_max_climb; }
+  dtNavMeshQuery *get_nav_query() { return _nav_query; }
+
+  void set_actor_height(float h) { _agent_height = h; }
+  void set_actor_radius(float r) { _agent_radius = r; }
+  void set_actor_clinb(float c) { _agent_max_climb = c; }
```
rdb, line 130: Typo, should be climb

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,164 @@
...
+  rcHeightfield *_solid;
+  rcCompactHeightfield *_chf;
+  rcContourSet *_cset;
+  rcPolyMesh *_pmesh;
+  rcConfig _cfg;
+  rcPolyMeshDetail *_dmesh;
+
+  void cleanup();
+
+public:
```
rdb, line 115: We usually put the public interface at the top so you can easily see the public interface after opening the header file.  Also, this should be `PUBLISHED:` to make these methods exposed to Python.

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,164 @@
...
+
+  void process_geom_node(GeomNode *geomnode, int &vcap, int &tcap);
+  void process_geom(CPT(Geom) geom, int &vcap, int &tcap);
+  void process_vertex_data(const GeomVertexData *vdata, int &vcap);
+  void process_primitive(const GeomPrimitive *orig_prim, const GeomVertexData *vdata, int &tcap);
+
+  std::map<LVector3, int> mp;
+  std::vector<LVector3> vc, fc;
+  bool loaded;
+  int index_temp;
```
rdb, line 73: Members should be underscored.  Also, `mp` is not a very descriptive name; you should favour readability over terseness.

panda/src/navigation/navMeshNode.h
```diff
@@ -0,0 +1,21 @@
...
+#include "navMesh.h"
+#include <string>
+
+class NavMeshNode: public PandaNode
+{
+private:
+  NavMesh _nav_mesh;
+
+public:
+  NavMeshNode(std::string &name, NavMesh nav_mesh);
```
rdb, line 17: Should be `const std::string &name` since it's not modifying it

NavMesh should be a reference-counted class, so `PT(NavMesh)` (and make NavMesh inherit from ReferenceCount, or better yet, TypedWritableReferenceCount).

panda/src/navigation/config_navigation.cxx
```diff
@@ -0,0 +1,40 @@
+
```
rdb, line 1: All your source/header files (not the Recast/Detour ones) need to have a file header with the license, author and date, like in other Panda source files.

panda/src/navigation/navMesh.h
```diff
@@ -0,0 +1,26 @@
+#ifndef NAVMESH_H
+#define NAVMESH_H
+
+#include "Recast.h"
+#include "DetourNavMesh.h"
+#include "pandaFramework.h"
```
rdb, line 6: pandaFramework.h should not be included here

panda/src/navigation/navMesh.h
```diff
@@ -0,0 +1,26 @@
...
+
+class NavMesh
+{
```
rdb, line 9: NavMesh should be reference-counted, so that it can be passed between Python and C++ applications without issues.  You can do this by inheriting from ReferenceCount and making sure you allocate it with `new` and pass it by pointer.  Use `PT()` like with other Panda classes whenever storing it to make sure that the object won't get deleted.

We actually want to inherit it from TypedReferenceCount (which adds typing information on top of ReferenceCount, for the benefit of using the class from Python), and add all the TypeHandle boilerplate that can be seen in other Panda classes (look at `panda/src/skel/typedSkel.h` for an example of this.)

Eventually, we will want to change it to inherit from TypedWritableReferenceCount, which will be necessary to add the ability to serialise NavMesh objects from/to .bam files.

These suggestions only apply to NavMesh (and NavMeshNode, but it already inherits from TypedWritableReferenceCount indirectly through PandaNode), since these are the only classes that will need to be serialised to disk.

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,164 @@
...
+
+  void set_actor_height(float h) { _agent_height = h; }
+  void set_actor_radius(float r) { _agent_radius = r; }
+  void set_actor_clinb(float c) { _agent_max_climb = c; }
+  void set_partition_type(std::string p);
+
+  void reset_common_settings();
+
+  bool build();
+  NavMesh get_navmesh();
```
rdb, line 136: Just thinking out loud, but I think that once you made NavMesh a class that is passed by pointer, you can wrap it and `build()` into a single call: `PT(NavMesh) build();` and simply return `nullptr` if it failed to build.

panda/src/navigation/navMeshBuilder.h
ash1215: Done

panda/src/navigation/navMeshBuilder.h
ash1215: Added them in the lastest commit

panda/src/navigation/navMeshBuilder.h
ash1215: Done

panda/src/navigation/navMeshNode.h
ash1215: Done in latest commit.

panda/src/navigation/config_navigation.cxx
ash1215: Done

panda/src/navigation/navMesh.h
ash1215: Done in latest commit

panda/src/navigation/navMeshBuilder.h
ash1215: Removed get_navmesh function. Now build() returns PT(NavMesh)

panda/src/navigation/navMesh.h
ash1215: Tried removing it. Gave errors.
Probably because "TypedWritableReferenceCount" requires that header.

ash1215:
@rdb Thank you for the suggestions and all the help. I have done the changes mentioned here on this thread. Please review and let me know if things are fine or not? 😅

panda/src/navigation/config_navigation.cxx
rdb: Still missing from config_navigation.*

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,30 @@
+/**
+ * PANDA 3D SOFTWARE
+ * Copyright (c) Carnegie Mellon University.  All rights reserved.
+ *
+ * All use of this software is subject to the terms of the revised BSD
+ * license.  You should have received a copy of this license along
+ * with this source code in a file named "LICENSE."
+ *
+ * @file movieAudio.h
```
rdb, line 9: Every file contains `@file movieAudio.h`

panda/src/navigation/navMeshNode.h
```diff
@@ -0,0 +1,36 @@
...
+{
+PUBLISHED:
+  NavMeshNode(const std::string &name, PT(NavMesh) nav_mesh);
+private:
+  PT(NavMesh) _nav_mesh;
+
+public:
+  
+  ~NavMeshNode();
+};
```
rdb, line 34: Class NavMeshNode needs a docstring

Since this class (indirectly) inherits from TypedObject, it needs TypeHandle boilerplate for use in Python (see the stuff at the bottom of MovieAudio and CollisionNode et al), this means adding this to the bottom of the class:

```cpp
public:
  static TypeHandle get_class_type() {
    return _type_handle;
  }
  static void init_type() {
    PandaNode::init_type();
    register_type(_type_handle, "NavMeshNode",
                  PandaNode::get_class_type());
  }
  virtual TypeHandle get_type() const {
    return get_class_type();
  }
  virtual TypeHandle force_init_type() {init_type(); return get_class_type();}

private:
  static TypeHandle _type_handle;
```

and this to the top of the .cxx file, below the includes:
```cpp
TypeHandle NavMeshNode::_type_handle;
```

And this to the config_navigation.cxx file, in the init function:
```
  NavMeshNode::init_type();
```

panda/src/navigation/navMesh.h
rdb: Well, maybe you should replace it with an include to `typedWritableReferenceCount.h` :-)

It may be that pandaFramework.h is indirectly including the header that you actually need.

panda/src/navigation/config_navigation.cxx
```diff
@@ -0,0 +1,53 @@
...
+Configure(config_navigation);
+NotifyCategoryDef(navigation, "");
+
+ConfigureFn(config_navigation) {
+  init_libnavigation();
+}
+
+ConfigVariableInt navigation_sample_config_variable
+("navigation-sample-config-variable", 25);
```
Moguri, line 35: Do not add 'config-variable' to the name, and please add a description.

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+  int partition_type;
+  float nav_mesh_bMin[3];
+  float nav_mesh_bMax[3];
+  float tile_size;
+};
+
+
+class EXPCL_NAVIGATION NavMeshBuilder {
+PUBLISHED:
+  float get_actor_radius() { return _agent_radius; }
```
Moguri, line 67: We should create properties for these using `MAKE_PROPERTY()`. `nodePath.h` has examples of this, but for `actor_radius`, you would do:
```cpp
MAKE_PROPERTY(actor_radius, get_actor_radius, set_actor_radius)
```

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+  float get_actor_height() { return _agent_height; }
+  float get_actor_climb() { return _agent_max_climb; }
+  void set_actor_height(float h) { _agent_height = h; }
```
Moguri, line 70: Please avoid single letter variable names for parameter names so that the auto-generated API docs are cleaner.

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+  void set_actor_radius(float r) { _agent_radius = r; }
+  void set_actor_climb(float c) { _agent_max_climb = c; }
+  void set_partition_type(std::string p);
+  void reset_common_settings();
+  bool from_node_path(NodePath node);
```
Moguri, line 75: These names imply that you are creating a NavMeshBuilder from these inputs. Looking at the rest of this API, I believe `add_node_path()` and `add_geom()` may be more appropriate. Does it make sense to also expose `add_vertex()` and `add_triangle()`?

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+  bool from_geom(PT(Geom) geom);
+  PT(GeomNode) draw_poly_mesh_geom();
+
+  PT(NavMesh) build();
+  NavMeshBuilder();
+
+  int get_vert_count() const { return _vert_count; }
```
Moguri, line 82: Should these be builder functions? I am not sure why the builder has verts and such; I would expect this kind of stuff on the NavMesh class.

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+  int index_temp;
+protected:
+  PT(NavMesh) _nav_mesh_obj;
+  class dtNavMesh *_nav_mesh;
+  class dtNavMeshQuery *_nav_query;
+  class dtCrowd *_crowd;
+
+  unsigned char _nav_mesh_draw_flags;
+
+  float _cell_size;
```
Moguri, line 117: Any reason to not store all of these in an instance of `BuildSettings` instead of duplicating the individual items here?

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+
+  std::map<LVector3, int> vertex_map;
+  std::vector<LVector3> vertex_vector, face_vector;
+  bool _loaded;
+  int index_temp;
+protected:
+  PT(NavMesh) _nav_mesh_obj;
+  class dtNavMesh *_nav_mesh;
+  class dtNavMeshQuery *_nav_query;
+  class dtCrowd *_crowd;
```
Moguri, line 113: Are we using DetourCrowd?

tests/navigation/test_navigation.py
```diff
@@ -0,0 +1,85 @@
...
+
+    prim.add_vertices(2,3,5)
+    prim.closePrimitive()
+    geom.add_primitive(prim)
+
+    node = core.GeomNode('gnode')
+    node.addGeom(geom)
+
+    scene = NodePath(node)
+    # Defining navmesh as object of NavMehsBuilder class.
```
Moguri, line 32: This comment doesn't make any sense to me. Maybe just remove it?

panda/src/navigation/navMeshBuilder.cxx
```diff
@@ -0,0 +1,775 @@
...
+#include <string>
+#include <iostream>
+#include "geom.h"
+#include "geomVertexFormat.h"
+#include "geomVertexWriter.h"
+#include "geomTrifans.h"
+
+
+#ifdef WIN32
+# define snprintf _snprintf
```
Moguri, line 32: This appears to be unused; can it be removed?

tests/navigation/test_navigation.py
```diff
@@ -0,0 +1,85 @@
...
+    nav = navigation.NavMeshBuilder()
```
Moguri, line 33: I think we should avoid starting a convention where we call `NavMeshBuilder` instances `nav` since that is better suited for the nav mesh itself. Maybe `nav_builder` or just `builder`?

tests/navigation/test_navigation.py
```diff
@@ -0,0 +1,85 @@
...
+    vertex.add_data3((0, 0, 0))
+    vertex.add_data3((100, 1000, 0))
+    vertex.add_data3((0, 100, 2))
+    vertex.add_data3((500, 200, 2))
+    vertex.add_data3((170, 20, 2))
+    vertex.add_data3((100, 100, 100))
+
+    prim = core.GeomTriangles(core.Geom.UH_static)
+    prim.add_vertices(0, 1, 4)
+    prim.closePrimitive()
```
Moguri, line 19: Prefer snake-case for new Python code.

panda/src/navigation/navMeshBuilder.h
ash1215: The functions like set_actor_*() are optional. Those variables have a default value while from_node_path()/from_geom() is a compulsory function and is required to be called before build(). So, I thought to have different naming for them.

ash1215: Also, we may expose add_vertex() but that I don't think would make much sense as we cannot expose add_triangle(). Add_triangle() requires indices of 3 vertices as input. The user won't be knowing what index is assigned to what vertex.

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+
+#include "navMesh.h"
+#include "geom.h"
+#include "geomTrifans.h"
+
+TypeHandle NavMesh::_type_handle;
+
+NavMesh::NavMesh() {
```
rdb, line 21: These methods still lack docstrings.

panda/src/navigation/navMesh.h
```diff
@@ -0,0 +1,56 @@
...
+
+
+#ifndef NAVMESH_H
+#define NAVMESH_H
+
+#include "Recast.h"
+#include "DetourNavMesh.h"
+#include "typedWritableReferenceCount.h"
+
+class EXPCL_NAVIGATION NavMesh: public TypedWritableReferenceCount
```
rdb, line 22: Class lacks docstring.

panda/src/navigation/navMeshBuilder.cxx
```diff
@@ -0,0 +1,775 @@
...
+        LVector3 xvx = { float(a + 1), float(b + 1), float(c + 1) };
+        face_vector.push_back(xvx);
+        add_triangle(a, b, c, tcap);
+
+      }
+    }
+    else continue;
+
+  }
+  return;
```
rdb, line 205: Return at the end of a method is unnecessary

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+  void add_triangle(int a, int b, int c, int &cap);
+
+
+  void process_geom_node(GeomNode *geomnode, int &vcap, int &tcap);
+  void process_geom(CPT(Geom) geom, int &vcap, int &tcap);
+  void process_vertex_data(const GeomVertexData *vdata, int &vcap);
+  void process_primitive(const GeomPrimitive *orig_prim, const GeomVertexData *vdata, int &tcap);
+
+  std::map<LVector3, int> vertex_map;
+  std::vector<LVector3> vertex_vector, face_vector;
```
rdb, line 106: All members need their names to start with underscores so that they can be easily distinguished.

panda/src/navigation/navMeshBuilder.h
rdb: A convenience add_triangle method exposed to Python could surely take three LPoint3 instances, which are looked up in the vertex map (and if not present, added to the vertices).

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+  ~NavMeshBuilder();
+  void set_context(rcContext *ctx) { _ctx = ctx; }
+
+  virtual void collect_settings(struct BuildSettings &settings);
+
+  
+  dtNavMeshQuery *get_nav_query() { return _nav_query; }
+
+ 
+  unsigned char getNavMeshDrawFlags() const { return _nav_mesh_draw_flags; }
```
rdb, line 163: Use snake_case

panda/src/navigation/navMeshBuilder.h
```diff
@@ -0,0 +1,181 @@
...
+  PT(NavMesh) build();
+  NavMeshBuilder();
+
+  int get_vert_count() const { return _vert_count; }
+  int get_tri_count() const { return _tri_count; }
+  int get_pmesh_vert_count() { return _pmesh->nverts; }
+  int get_pmesh_poly_count() { return _pmesh->npolys; }
+  int get_pmesh_max_poly_count() { return _pmesh->maxpolys; }
+private:
+  std::string _filename;
```
rdb, line 88: Why is there a filename on NavMeshBuilder?.

panda/src/navigation/navMeshNode.h
```diff
@@ -0,0 +1,55 @@
...
+#define NAVMESHNODE_H
+
+#include <string>
+#include "geom.h"
+#include "pandaFramework.h"
+#include "pandaSystem.h"
+#include "navMesh.h"
+#include <string>
+
+class EXPCL_NAVIGATION NavMeshNode: public PandaNode
```
rdb, line 24: Missing docstring

panda/src/navigation/navMeshBuilder.h
ash1215: yes, will add that function as well

panda/src/navigation/navMesh.h
```diff
@@ -0,0 +1,69 @@
...
+  void set_nav_mesh(dtNavMesh *m) { _nav_mesh = m; }
+  NavMesh();
+
+private:
+  dtNavMesh *_nav_mesh;
+  
+public:
+  bool init_nav_mesh();
+  dtNavMesh *get_nav_mesh() { return _nav_mesh; }
+  dtNavMeshCreateParams params;
```
Moguri, line 37: Member variables should have a leading underscore (i.e., `_params`). However, this may also be confusing with the "params" for BAM serialization. Maybe `_dt_params` or `_mesh_params` would be clearer? I would also prefer this to be private.

panda/src/navigation/navMeshNode.cxx
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Writes the contents of this object to the datagram for shipping out to a
+ * Bam file.
+ */
+void NavMeshNode::
+write_datagram(BamWriter *manager, Datagram &dg) {
+  PandaNode::write_datagram(manager, dg);
+
+  manager->write_pointer(dg, _nav_mesh);
+  // Write NULL pointer to indicate the end of the list.
```
Moguri, line 50: The Bullet stuff does this for a list of shapes, but we only have the one NavMesh, so I don't think we need a null value to indicate the end of a list.

panda/src/navigation/navMeshNode.cxx
```diff
@@ -0,0 +1,101 @@
...
+
+  return pi;
+}
+
+// /**
+//  * This function is called by the BamReader's factory when a new object of
+//  * type BulletShape is encountered in the Bam file.  It should create the
+//  * BulletShape and extract its information from the file.
+//  */
+// TypedWritable *NavMeshNode::
```
Moguri, line 73: Remove this code if we don't need it.

@rdb, we noticed that BulletBodyNode did not define a `make_from_bam()` (presumably making use of `PandaNode::make_from_bam()`). Is this correct, or is BulletBodyNode potentially broken by not defining its own `make_from_bam()`?

panda/src/navigation/navMeshNode.cxx
```diff
@@ -0,0 +1,101 @@
...
+// }
+
+/**
+ * This internal function is called by make_from_bam to read in all of the
+ * relevant data from the BamFile for the new BulletTriangleMeshShape.
+ */
+void NavMeshNode::
+fillin(DatagramIterator &scan, BamReader *manager) {
+  PandaNode::fillin(scan, manager);
+
```
Moguri, line 93: Remove extra, unnecessary whitespace.

panda/src/navigation/navMeshQuery.cxx
```diff
@@ -0,0 +1,198 @@
...
+NavMeshQuery::NavMeshQuery():
+  _nav_query(0) {
+  _nav_query = dtAllocNavMeshQuery();
+}
+
+NavMeshQuery::NavMeshQuery(PT(NavMesh) nav_mesh):
+  _nav_query(0) {
+  
+  _nav_query = dtAllocNavMeshQuery();
+  dtStatus status = _nav_query->init(nav_mesh->get_nav_mesh(), 2048);
```
Moguri, line 28: This `2048` should come from a constant (can be static to this compilation unit) or a member variable. Using a variable instead of a "magic number" makes the code more self-explanatory.

panda/src/navigation/navMeshQuery.cxx
```diff
@@ -0,0 +1,198 @@
...
+    std::cout<<"\nCannot set nav query!\n";
+  }
+
+}
+
+NavMeshQuery::~NavMeshQuery() {
+  dtFreeNavMeshQuery(_nav_query);
+}
+
+bool NavMeshQuery::set_nav_query(PT(NavMesh) nav_mesh) {
```
Moguri, line 40: These functions all need doc strings which will be used for our generated API reference.

panda/src/navmeshgen/navMeshBuilder.cxx
```diff
@@ -0,0 +1,738 @@
...
+  _cset = 0;
+  rcFreePolyMesh(_pmesh);
+  _pmesh = 0;
+  rcFreePolyMeshDetail(_dmesh);
+  _dmesh = 0;
+  
+}
+
+void NavMeshBuilder::set_partition_type(std::string p) {
+  if (p == "watershed" || p == "Watershed" || p == "WATERSHED") {
```
Moguri, line 277: `string_utils.h` has a `downcase()` to convert a string to all lowercase, which would make this check nicer and more robust:

```cpp
if (downcase(p) == "watershed") {
...
}```

File: panda/src/navigation/navMesh.h (hunk 29 above)
Comment (ash1215) [reply panda/src/navigation/navMesh.h:37]: Actually, I was calling this member and giving it values in NavMeshBuilder.cxx . So, I thought to keep it public.

File: panda/src/navigation/navMeshBuilder.h (hunk 27 above)
Comment (ash1215) [reply panda/src/navigation/navMeshBuilder.h]: Yeah, forgot to remove it. Earlier, I was taking .obj filename as input and converting it to polymesh. Removed it now.

File: panda/src/navigation/config_navigation.cxx [hunk 36]
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Initializes the library.  This must be called at least once before any of
+ * the functions or classes in this library can be used.  Normally it will be
+ * called by the static initializers and need not be called explicitly, but
+ * special cases exist.
+ */
+void
+init_libnavigation() {
+  NavMesh::init_type();
```
Moguri, line 44: I believe initializing these types should come after the init guard (`if (initialized)`). Also, I see that config_bullet.cxx is calling the `register_with_read_factory()` member functions on various types, so I am guessing you'll need to do the same.

Moguri:
A few API issues I have noticed so far:
1. NavMesh visualization should be part of NavMesh/NavMeshNode (e.g., show/hide) similar to how CollisionNodes do debug visualizations
2. The various params do not seem well exposed to Python. I think we should wrap the params in a class that can be created by Python users for things like passing into the NavMeshBuilder or even for creating NavMesh instances without NavMeshBuilder. We should also not be passing raw Detour objects between our classes (at least, not for "main" interfaces like constructors).
3. NavMeshQuery seems wonky in how stateful it is (reminds me of OpenGL, which isn't a good thing). In particular, I do not like setting the NavMesh to query against as a function of an existing NavMeshQuery instance. Instead, I would prefer to make a Query instance per NavMesh (e.g., pass a NavMesh into the constructor) or to use functions that accept the NavMesh to use for the Query.
4. Users should not have to dig deep into objects to use basic API functions. For example, I dislike the `navmesh.node().get_navmesh()` stuff just to make a basic query.

Moguri:
Please remove the unnecessary ShowBase and DirectStart imports from the unit tests so the CI can start passing. The CI environment is unable to open a Window, and it does not look like a Window is needed for these tests.

panda/src/navigation/config_navigation.cxx
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Initializes the library.  This must be called at least once before any of
+ * the functions or classes in this library can be used.  Normally it will be
+ * called by the static initializers and need not be called explicitly, but
+ * special cases exist.
+ */
+void
+init_libnavigation() {
+  
+  std::cout<<"\nCalled init_libnavigation\n";
```
Moguri, line 46: Use Notify and not stdout

panda/src/navigation/config_navigation.cxx
```diff
@@ -0,0 +1,63 @@
...
+  static bool initialized = false;
+  if (initialized) {
+    return;
+  }
+  initialized = true;
+  std::cout<<"\nCalling init_type\n";
```
Moguri, line 52: Use Notify and not stdout

panda/src/navigation/config_navigation.cxx
```diff
@@ -0,0 +1,63 @@
...
+  NavMesh::init_type();
+  NavMeshNode::init_type();
+  
+  std::cout<<"\nCalling register_with_read_factory\n";
```
Moguri, line 56: Use Notify and not stdout

panda/src/navigation/config_navigation.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "pandabase.h"
+#include "notifyCategoryProxy.h"
+#include "configVariableDouble.h"
+#include "configVariableString.h"
+#include "configVariableInt.h"
+
+NotifyCategoryDecl(navigation, EXPCL_NAVIGATION, EXPTP_NAVIGATION);
+
+extern ConfigVariableInt    navigation_sample_config_variable;
```
Moguri, line 25: Remove unused PRC variable left over from the skeleton/sample file.

panda/src/navigation/navMesh.h
```diff
@@ -0,0 +1,101 @@
...
+  float b_min[3];
+  float b_max[3];
+  unsigned short *verts;        // size: 3 * vert_count
+  unsigned short *polys;        // size: poly_count * 2 * nvp
+  unsigned short *poly_flags;   // size: poly_count
+  unsigned char *poly_areas;    // size: poly_count
+  unsigned int *detail_meshes;  // size: poly_count * 4
+  float *detail_verts;          // size: detail_vert_count * 3
+  unsigned char *detail_tris;   // size: detail_tri_count * 4
+  int RC_MESH_NULL_IDX = 65535;
```
Moguri, line 46: Should this be a constant? Also, is this something we should expose to Python?

panda/src/navigation/navMeshNode.cxx
```diff
@@ -0,0 +1,96 @@
...
+int NavMeshNode::
+complete_pointers(TypedWritable **p_list, BamReader *manager) {
+  int pi = PandaNode::complete_pointers(p_list, manager);
+
+  _nav_mesh = DCAST(NavMesh, p_list[pi++]);
+
+  return pi;
+}
+
+// /**
```
Moguri, line 68: Remove the single-line (`//`) comments, you already have this as a mutli-line (`/* ... */`) comment.

panda/src/navmeshgen/config_navmeshgen.cxx
```diff
@@ -0,0 +1,49 @@
...
+#endif
+
+Configure(config_navmeshgen);
+NotifyCategoryDef(navmeshgen, "");
+
+ConfigureFn(config_navmeshgen) {
+  init_libnavmeshgen();
+}
+
+ConfigVariableInt navmeshgen_sample_config_variable
```
Moguri, line 31: Remove unused PRC variable

panda/src/navmeshgen/config_navmeshgen.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "pandabase.h"
+#include "notifyCategoryProxy.h"
+#include "configVariableDouble.h"
+#include "configVariableString.h"
+#include "configVariableInt.h"
+
+NotifyCategoryDecl(navmeshgen, EXPCL_NAVMESHGEN, EXPTP_NAVMESHGEN);
+
+extern ConfigVariableInt    navmeshgen_sample_config_variable;
```
Moguri, line 25: Remove unused PRC variable

samples/street-navigation/main.py
```diff
@@ -0,0 +1,46 @@
...
+from panda3d.core import PointLight,DirectionalLight
+
+class MyApp(ShowBase):
+
+    def __init__(self):
+        ShowBase.__init__(self)
+
+        # Setting up light for better view.
+        plight = PointLight('plight')
+        plight.setColor((20, 0.9, 0.9, 1))
```
Moguri, line 12: Prefer snake_case for new samples

samples/street-navigation/main.py
```diff
@@ -0,0 +1,46 @@
...
+        self.scene.setPos(-8, 42, 0)
+
+        # NavMeshBuilder is a class that is responsible for building the polygon meshes and navigation meshes.
+        self.builder = navigation.NavMeshBuilder()
+        # Take Nodepath as input. Nodepath should contain the required geometry.
+        self.builder.fromNodePath(self.scene)
+
+        self.builder.build()
+
+        # Code to attach the polymesh generated to the scene graph
```
Moguri, line 37: We should be attaching a NavMeshNode to the scene graph, not a Geom representation of the NavMesh. Also, show()/hide() on the NavMeshNode (or maybe it's the NavMesh?) should use this debug drawing capability. We should remove Geom generation from the builder.

A shortcut (e.g., bind f1) to show/hide the navmesh debug mesh would be nice.

samples/street-navigation/main.py
```diff
@@ -0,0 +1,46 @@
...
+        self.node1 = self.builder.drawPolyMeshGeom()
+        self.node = self.scene.attachNewNode(self.node1)
+        self.node.setColor(0,0,1)
+        
+
```
Moguri, line 42: The sample should demonstrate querying. I think the most straight-forward example would be to have an Actor that moves to a location based on a mouse click. You can use Intervals to move the Actor along the points found from querying. It would also be nice to represent the current path (e.g., using line segments).

tests/navigation/test_nav_mesh_bam.py
```diff
@@ -0,0 +1,165 @@
...
+
+    reader = core.BamReader(buffer)
+    reader.init()
+    object = reader.read_object()
+    reader.resolve()
+    return object
+
+
+def test_nav_mesh_bam():
+
```
Moguri, line 28: No empty line after function signatures

tests/navigation/test_nav_mesh_bam.py
```diff
@@ -0,0 +1,165 @@
...
+    assert type(navmesh) == type(navmeshBam)
+    assert vertCount == navmeshBam.get_vert_count()
+    assert polyCount == navmeshBam.get_poly_count()
+    assert detailVertsCount == navmeshBam.get_detail_vert_count()
+    assert detailTriCount == navmeshBam.get_detail_tri_count()
+
+
+def test_nav_mesh_node_bam():
+
```
Moguri, line 74: No empty line after function signatures

tests/navigation/test_nav_mesh_bam.py
```diff
@@ -0,0 +1,165 @@
...
+    assert type(navmeshnode) == type(navmeshNodeBam)
+    assert vertCount == navmeshNodeBam.get_nav_mesh().get_vert_count()
+    assert polyCount == navmeshNodeBam.get_nav_mesh().get_poly_count()
+    assert detailVertsCount == navmeshNodeBam.get_nav_mesh().get_detail_vert_count()
+    assert detailTriCount == navmeshNodeBam.get_nav_mesh().get_detail_tri_count()
+
+
+def test_nav_query_bam():
+
```
Moguri, line 121: No empty line after function signatures

tests/navigation/test_nav_mesh_bam.py
```diff
@@ -0,0 +1,165 @@
...
+    navmeshBam = reconstruct(navmesh)
+    queryBam = navigation.NavMeshQuery(navmeshBam)
+    posBam = LPoint3(50,55,3);
+    queryBam.nearestPoint(posBam)
+    pathBam = query.findPath(posBam,LPoint3(100,0,3))
+
+    assert pos == posBam
+    assert len(path) == len(pathBam)
+
```
Moguri, line 165: No extra new-line at the end of the file

tests/navmeshgen/test_mesh_gen.py
```diff
@@ -0,0 +1,85 @@
+from direct.showbase.ShowBase import ShowBase
+from panda3d import navigation
+from panda3d import core
+from panda3d.core import NodePath
+
+def test_input_geom():
+
```
Moguri, line 7: No empty line after function signatures

tests/navmeshgen/test_mesh_gen.py
```diff
@@ -0,0 +1,85 @@
...
+    nav.fromNodePath(scene)
+    # Finding number of vertices.
+    vcount = nav.get_vert_count()
+    # Finding number of triangles.
+    tcount = nav.get_tri_count()
+    assert vcount == 6
+    assert tcount == 4
+
+def test_poly_mesh():
+
```
Moguri, line 45: No empty line after function signatures

tests/navigation/test_nav_mesh_bam.py
```diff
@@ -0,0 +1,165 @@
+from direct.showbase.ShowBase import ShowBase
+import direct.directbase.DirectStart
+from panda3d import navigation
+from panda3d import navmeshgen
+from panda3d import core
+from panda3d.core import NodePath
```
Moguri, line 6: Why import `NodePath` and `LPoint3` specifically when you grab everything else via `core`? Might as well keep the namespacing intact and use `core.NodePath` and `core.LPoint3`.

samples/street-navigation/main.py
```diff
@@ -0,0 +1,46 @@
...
+        # Loading the model
+        self.scene = self.loader.loadModel("models/street.obj")
+        self.scene.reparentTo(self.render)
+        self.scene.setP(90)
+        self.scene.flatten_light()
+        self.scene.setScale(0.25, 0.25, 0.25)
+        self.scene.setPos(-8, 42, 0)
+
+        # NavMeshBuilder is a class that is responsible for building the polygon meshes and navigation meshes.
+        self.builder = navigation.NavMeshBuilder()
```
Moguri, line 31: Shouldn't this be `navmeshgen.NavMeshBuilder()`?

ash1215: I have not updated the sample code as of now. I planned to write the sample code again after finding new level

panda/src/navigation/navMeshBuilder.h
ash1215: Removed

panda/src/navigation/navMeshNode.cxx
ash1215: Removed

panda/src/navmeshgen/navMeshBuilder.cxx
ash1215: done

panda/src/navigation/config_navigation.cxx
rdb: If this is not used for anything it should be removed.

panda/src/navigation/config_navigation.cxx
rdb: This should be removed as it looks like a test print.  Same with the rest of the file.

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,388 @@
...
+
+
+#include "navMesh.h"
+#include "geom.h"
+#include "geomTrifans.h"
+#include <iostream>
+
+TypeHandle NavMesh::_type_handle;
+
+NavMesh::NavMesh():
```
rdb, line 22: Don't forget about the docstrings for the methods!

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,388 @@
...
+  _nav_mesh = 0;
+}
+
+NavMesh::NavMesh(dtNavMesh *nav_mesh) {
+  _nav_mesh = nav_mesh;
+}
+
+NavMesh::NavMesh(NavMeshParams mesh_params):
+  _nav_mesh(0) {
+  memset(&(_params), 0, sizeof(_params));
```
rdb, line 36: There's probably a more idiomatic way to do this in C++.  I guess `_params = {};` should do it.

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,388 @@
...
+  _params.ch = mesh_params.ch;
+  _params.buildBvTree = mesh_params.build_bv_tree;
+  
+  RC_MESH_NULL_IDX = mesh_params.RC_MESH_NULL_IDX;
+  
+  init_nav_mesh();
+}
+
+bool NavMesh::init_nav_mesh() {
+	unsigned char *nav_data = 0;
```
rdb, line 73: Tab character

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,388 @@
...
+  int nav_data_size = 0;
+
+  if (!dtCreateNavMeshData(&_params, &nav_data, &nav_data_size)) {
+    std::cout<<"\nCould not build Detour navmesh.\n";
```
rdb, line 77: You should use a notify category, like:
```cpp
    navigation_cat.error()
      << "Could not build Detour navmesh.\n";
```

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,388 @@
...
+
+    //convert to world space
+    const float x = orig[0] + v[0] * cs;
+    const float y = orig[1] + v[1] * ch;
+    const float z = orig[2] + v[2] * cs;
+
+    vertex.add_data3(x, -z, y); //if origingally model is z-up
+    //vertex.add_data3(x, y, z); //if originally model is y-up
+    colour.add_data4((float)rand() / RAND_MAX, (float)rand() / RAND_MAX, (float)rand() / RAND_MAX, 1);
+    //std::cout << "index: " << i / 3 << "\t" << x << "\t" << y << "\t" << z << "\n";
```
rdb, line 134: If there is time you should take into account that the `coordinate-system` Config.prc variable may not be set to Z-up necessarily.  It may be necessary to, before the loop, store something like:
```python
LMatrix4 mat = LMatrix4::convert_mat(CS_yup_right, CS_default);

And then during the loop use `mat.convert_point` to convert your points according to the given transformation.

Either that, or you can pass the matrix to a call to `geom.transform_vertices` after the fact.
```

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,388 @@
...
+  std::cout << "Number of Polygons: " << _params.polyCount << std::endl;
+
+
+  return node;
+}
+
+
+
+
+
```
rdb, line 184: Please remove unnecessary lines of spacing.  One empty line between methods is enough.

panda/src/navigation/navMesh.h
rdb: Looks like you added the typedWirtableReferenceCount.h include, so now it should be OK to remove pandaFramework.h.

panda/src/navigation/navMesh.h
rdb: Agreed, is this necessary?

panda/src/navigation/navMesh.cxx
```diff
@@ -0,0 +1,388 @@
...
+    detail_verts[i] = scan.get_float32();
+  }
+
+  unsigned char *detail_tris = new unsigned char[detail_tri_count * 4];
+  for(int i=0 ; i < detail_tri_count * 4 ;i++) {
+    detail_tris[i] = scan.get_uint8();
+  }
+
+  //RC_MESH_NULL_IDX
+  RC_MESH_NULL_IDX = scan.get_int32();
```
rdb, line 352: This is a constant, we don't need to store it anywhere.

panda/src/navigation/navMesh.h
```diff
@@ -0,0 +1,101 @@
...
+  PT(GeomNode) draw_nav_mesh_geom();
+  int get_vert_count() { return _params.vertCount; }
+  int get_poly_count() { return _params.polyCount; }
+  int get_detail_vert_count() { return _params.detailVertsCount; }
+  int get_detail_tri_count() { return _params.detailTriCount; }
+
+private:
+  dtNavMesh *_nav_mesh;
+  dtNavMeshCreateParams _params;
+  int RC_MESH_NULL_IDX;
```
rdb, line 65: Again, since this is a constant, this should not be stored (and if it really should be stored, it should conform to the naming conventions for members).

panda/src/navigation/navMeshQuery.cxx
```diff
@@ -0,0 +1,229 @@
...
+  if (dtStatusFailed(status)) {
+    std::cout<<"\nCannot set nav query!\n";
+    return false;
+  }
+
+  return true;
+}
+
+bool NavMeshQuery::set_nav_query(NodePath nav_mesh_node_path) {
+  NavMeshNode *node = dynamic_cast<NavMeshNode*>(nav_mesh_node_path.node());
```
rdb, line 67: Use Panda's own casting system, which does not rely on RTTI.  To use it: `DCAST(NavMeshNode, nav_mesh_node_path.node())`

samples/navigation/main.py
```diff
@@ -0,0 +1,211 @@
...
+        #self.scene.attachNewNode(cm.generate()).lookAt(0, 0, -1)
+        taskMgr.add(self.__getMousePos, "_YourClass__getMousePos")
+        #mouseclick end 
+
+
+        #Here start commenting
+        #NavMeshBuilder is a class that is responsible for building the polygon meshes and navigation meshes.
+        self.builder = navmeshgen.NavMeshBuilder()
+        self.structBuilder = navmeshgen.BuildSettings()
+        self.structBuilder.cell_size = 1;
```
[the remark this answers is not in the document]
ash1215: In the latest commit, I removed BuildSettings class and added separate functions to set and get settings which were defined in BuildSettings.

panda/src/navmeshgen/navMeshBuilder.h
```diff
@@ -0,0 +1,173 @@
...
+  float agent_max_climb;
+  float agent_max_slope;
+  float region_min_size;
+  float region_merge_size;
+  float edge_max_len;
+  float edge_max_error;
+  float verts_per_poly;
+  float detail_sample_dist;
+  float detail_sample_max_error;
+  int partition_type;
```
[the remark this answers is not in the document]
ash1215: Done. Please review if it is done correctly.

panda/src/navmeshgen/navMeshBuilder.cxx
```diff
@@ -0,0 +1,738 @@
...
+  rcFreeContourSet(_cset);
+  _cset = 0;
+  rcFreePolyMesh(_pmesh);
+  _pmesh = 0;
+  rcFreePolyMeshDetail(_dmesh);
+  _dmesh = 0;
+  
+}
+
+void NavMeshBuilder::set_partition_type(std::string partition) {
```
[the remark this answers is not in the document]
ash1215: Done. Please review if it is correctly done.

panda/src/navigation/navMeshBuilder.h
ash1215: I have added add_polygon method exposed to Python which takes in 3 LPoint3 instances.

panda/src/navigation/navMeshBuilder.h
ash1215: i have removed buildsettings class

panda/src/navigation/config_navigation.cxx
ash1215: Done

panda/src/navigation/navMesh.cxx
ash1215: Done for all such output instructions

panda/src/navmeshgen/config_navmeshgen.cxx
ash1215: Done

panda/src/navmeshgen/config_navmeshgen.h
ash1215: Done

samples/street-navigation/main.py
ash1215: Done

panda/src/navigation/config_navigation.h
ash1215: done

panda/src/navigation/config_navigation.cxx
ash1215: Removed it

panda/src/navigation/config_navigation.cxx
ash1215: Removed it

panda/src/navigation/navMeshNode.h
ash1215: Should docstrings be added in header files as well? I have added them in cxx files only for all the functions

panda/src/navigation/navMesh.h
ash1215: Should docstrings be added in header files as well? I have added them in cxx files only for all the functions @rdb

panda/src/navigation/navMeshBuilder.h
ash1215: Done

panda/src/navigation/config_navigation.cxx
ash1215: Removed

panda/src/navigation/navMesh.h
ash1215: It still gives error

panda/src/navigation/navMesh.cxx
ash1215: Done

panda/src/navigation/navMesh.cxx
ash1215: @rdb the vertices with this value belong to the edges which are solid borders. So if someone does not use navmeshgen to build the navmesh, then he/she might have a different value for such vertices.

panda/src/navigation/navMesh.h
ash1215: @rdb the vertices with this value belong to the edges which are solid borders. So if someone does not use navmeshgen to build the navmesh, then he/she might have a different value for such vertices.

panda/src/navmeshgen/navMeshBuilder.cxx
```diff
@@ -0,0 +1,738 @@
...
+
+  // Start the build process. 
+  _ctx->startTimer(RC_TIMER_TOTAL);
+
+  _ctx->log(RC_LOG_PROGRESS, "Building navigation:");
+  std::cout << "\nBuilding navigation:\n";
+  _ctx->log(RC_LOG_PROGRESS, " - %d x %d cells", _cfg.width, _cfg.height);
+  std::cout << "\n - " << _cfg.width << " x " << _cfg.height << " cells\n";
+  _ctx->log(RC_LOG_PROGRESS, " - %.1fK verts, %.1fK tris", nverts / 1000.0f, ntris / 1000.0f);
+  std::cout << "\n - " << nverts / 1000.0f << "K vetrs, " << ntris / 1000.0f << "K tris\n";
```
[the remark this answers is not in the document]
ash1215: Done

panda/src/navigation/navMesh.cxx
ash1215: Done

panda/src/navigation/navMesh.h
ash1215: the vertices with this value belong to the edges which are solid borders. So if someone does not use navmeshgen to build the navmesh, then he/she might have a different value for such vertices.

panda/src/navigation/navMesh.h
ash1215: Done

panda/src/navigation/navMesh.cxx
ash1215: Renamed it

panda/src/navigation/navMesh.h
ash1215: Renamed

samples/navigation/main.py
```diff
@@ -0,0 +1,211 @@
...
+        # Take Nodepath as input. Nodepath should contain the required geometry.
+        self.builder.fromNodePath(self.scene)
+
+        #self.builder.setActorRadius(5)
+        self.builder.set_actor_height(10)
+        self.navmesh = self.builder.build()
+
+        # Code to attach the polymesh generated to the scene graph
+        self.node1 = self.navmesh.drawNavMeshGeom()
```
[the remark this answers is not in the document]
ash1215: Done. Thanks :smiley:

samples/street-navigation/main.py
ash1215: Bound 's' for show() and 'h' for hide()

makepanda/makepanda.py
```diff
@@ -95,6 +95,8 @@
   "PANDAPARTICLESYSTEM",                               # Built in particle system
   "CONTRIB",                                           # Experimental
   "SSE2", "NEON",                                      # Compiler features
+  "NAVIGATION",				       # Detour Toolset
+  "NAVMESHGEN",                    # Recast Toolset
```
rdb, line 99: These look like tabs, rather than spaces!

rdb (changes requested):
This is good work, thank you!

I will be finding some more time to spend with this in the coming weeks, so that I can test out some more edge cases and see if any more API changes are needed (which I can make too if necessary), before merging this into master.  The plan is to target this feature for 1.11, which is probably not yet going to be released this year.

I've marked a few things that jumped out as still needing to be changed.  If you have time to address them, that would be great; otherwise, no problem, and someone else can pick those up.

Moguri:
This work is being continued in #1268; closing this PR.

PR #952, opened 2020-06-08
Add Distributed Network samples

fireclawthefox:
1: Simple client-server connection
2: Client-server connection with timeManager
3: Create distributed objects, sending and receiving messages
4: Distributed model file
5: Simple text chat
6: Simple smooth moving actor


Moguri:
Taking a skim over the samples, it looks like they are well commented, and that they follow the conventions set by the other samples. I cannot comment on the Distributed Networking stuff itself (e.g., if best practices are being observed).

@rdb This is a pretty large PR that could take quite some time to review. Since this is "just samples" and @fireclawthefox has shown a good understanding of Direct, maybe we just trust this PR is good and get it merged? We can always fix up the samples later as needed.

PR #954, opened 2020-06-15
gobj: Implement PythonTexturePoolFilter

darktohka:
This commit implements a special type of TexturePoolFilter: one that can be used to implement a texture pool filter directly in Python.

Rationale behind this change: I'd like to implement a texture pool filter that I'd like others to be able to use, but one that would most likely not be accepted upstream into Panda3D. As such, my goal was to create a general texture pool filter that could be programmed using Python, without having to rebuild Panda3D's C++ portion.

To use PythonTexturePoolFilter, create a Python class inheriting PythonTexturePoolFilter, and implement either the `pre_load` or `post_load` function, or both:

```py
pre_load(self, orig_filename, orig_alpha_filename,
         primary_file_num_channels, alpha_file_channel,
         read_mipmaps, options)
post_load(self, texture)
```

Returning a Texture inside `pre_load` will make Panda3D use that texture instead of looking for the on-disk variant.

To register the filter in the global texture pool:

```py
tex_filter = MyTexFilter()
tex_filter.register_filter(tex_filter)
```

To register the filter in another texture pool:
```py
tex_filter = MyTexFilter()
tex_pool = TexturePool.get_global_ptr()
tex_filter.register_filter(tex_filter, tex_pool)
```

Caveats:
* `pre_load` in TexturePool holds the `_lock` mutex. As such, loading another texture inside a custom texture pool filter and running the exact same filter on another texture causes a deadlock.
* A new LoaderOptions flag was added: `LF_no_filters`, which prevents the texture pool from attempting to use filters when enabled for the load request. This way, it does not attempt to hold `_lock` again.
* However, `_lock` is still held by `pre_load` at this time. To fix this, I've created a lock specific to filters: `_filter_lock`. This lock is only used while a texture is processing filters. (`_filter_lock` is necessary in case `_filter_registry` is modified during runtime).

I've created an example texture pool filter, one that will prioritize loading ".jpg" and ".rgb" textures over ".png" textures. If you're trying to load a PNG texture, this example texture pool filter will first check if there's a loadable JPG+RGB alpha combo texture, and if not, it will try to load the JPG variant of the same texture (for legacy reasons).

JPG+RGB > JPG > PNG

Please let me know your thoughts, and let me know if I've missed something!

[texturepoolfilter.zip](https://github.com/panda3d/panda3d/files/4777477/texturepoolfilter.zip)

### DIFF
--- a/panda/src/gobj/pythonTexturePoolFilter.cxx
+++ b/panda/src/gobj/pythonTexturePoolFilter.cxx
@@ -107,7 +107,7 @@ pre_load(const Filename &orig_filename, const Filename &orig_alpha_filename,
     DTool_CreatePyInstance((void *)&options, Dtool_LoaderOptions, false, true)
   );
 
-  PT(Texture) tex;
+  PT(Texture) tex = nullptr;
   PyObject *result = PythonThread::call_python_func(_pre_load_func, args);
 
   Py_DECREF(args);
@@ -154,7 +154,7 @@ post_load(Texture *tex) {
   // Wrap the arguments.
   PyObject *args = PyTuple_Pack(1, DTool_CreatePyInstance(tex, Dtool_Texture, true, false));
   PyObject *result = PythonThread::call_python_func(_post_load_func, args);
-  Texture *tex_res;
+  Texture *tex_res = nullptr;
 
   Py_DECREF(args);
 
--- a/panda/src/gobj/pythonTexturePoolFilter.cxx
+++ b/panda/src/gobj/pythonTexturePoolFilter.cxx
@@ -116,14 +116,17 @@ pre_load(const Filename &orig_filename, const Filename &orig_alpha_filename,
 
   Py_DECREF(args);
 
-  if (result != nullptr) {
+  if (result == Py_None) {
+    // We've got None as our result, so we shouldn't return any Texture
+    Py_DECREF(result);
+  } else if (result != nullptr) {
     if (DtoolInstance_Check(result)) {
       tex = (Texture *)DtoolInstance_UPCAST(result, Dtool_Texture);
     }
 
     if (tex == nullptr) {
       gobj_cat.error()
-        << "Preloading " << orig_filename.get_basename() << " failed as"
+        << "Preloading " << orig_filename.get_basename() << " failed as "
         << "preloaded texture is not of Texture type.\n";
     }
 
@@ -170,7 +173,10 @@ post_load(Texture *tex) {
 
   Py_DECREF(args);
 
-  if (result != nullptr) {
+  if (result == Py_None) {
+    // We've got None as our result, so we shouldn't return any Texture
+    Py_DECREF(result);
+  } else if (result != nullptr) {
     PT(Texture) result_tex;
 
     // Check if the call returned a texture pointer.
--- a/panda/src/gobj/pythonTexturePoolFilter.cxx
+++ b/panda/src/gobj/pythonTexturePoolFilter.cxx
@@ -72,7 +72,7 @@ init(PyObject *tex_filter) {
   if (_pre_load_func == nullptr && _post_load_func == nullptr) {
     PyErr_Format(PyExc_TypeError,
                  "texture pool filter plug-in %R does not define pre_load or post_load function",
-				 tex_filter);
+                 tex_filter);
     return false;
   }
 
--- a/panda/src/gobj/pythonTexturePoolFilter.cxx
+++ b/panda/src/gobj/pythonTexturePoolFilter.cxx
@@ -1,0 +1,201 @@
+/**
+ * PANDA 3D SOFTWARE
+ * Copyright (c) Carnegie Mellon University.  All rights reserved.
+ *
+ * All use of this software is subject to the terms of the revised BSD
+ * license.  You should have received a copy of this license along
+ * with this source code in a file named "LICENSE."
+ *
+ * @file pythonTexturePoolFilter.cxx
+ * @author Derzsi Daniel
+ * @date 2020-06-14
+ */
+
+#include "pythonTexturePoolFilter.h"
+
+#ifdef HAVE_PYTHON
+
+#include "pythonThread.h"
+#include "py_panda.h"
+
+extern struct Dtool_PyTypedObject Dtool_Filename;
+extern struct Dtool_PyTypedObject Dtool_LoaderOptions;
+extern struct Dtool_PyTypedObject Dtool_Texture;
+extern struct Dtool_PyTypedObject Dtool_PythonTexturePoolFilter;
+
+TypeHandle PythonTexturePoolFilter::_type_handle;
+
+/**
+ * Constructor.
+ */
+PythonTexturePoolFilter::
+PythonTexturePoolFilter() {
+  init_type();
+}
+
+/**
+ * Destructor.
+ */
+PythonTexturePoolFilter::
+~PythonTexturePoolFilter() {
+  if (_pre_load_func != nullptr || _post_load_func != nullptr) {
+#if defined(HAVE_THREADS) && !defined(SIMPLE_THREADS)
+    PyGILState_STATE gstate;
+    gstate = PyGILState_Ensure();
+#endif
+
+    Py_CLEAR(_pre_load_func);
+    Py_CLEAR(_post_load_func);
+
+#if defined(HAVE_THREADS) && !defined(SIMPLE_THREADS)
+    PyGILState_Release(gstate);
+#endif
+  }
+}
+
+/**
+ * Initializes the fields from the given Python texture pool object,
+ * then registers the filter at a texture pool.
+ */
+bool PythonTexturePoolFilter::
+register_filter(PyObject *tex_filter, TexturePool *tex_pool) {
+  nassertr(tex_filter != nullptr, false);
+  nassertr(_pre_load_func == nullptr, false);
+  nassertr(_post_load_func == nullptr, false);
+
+  _pre_load_func = PyObject_GetAttrString(tex_filter, "pre_load");
+  _post_load_func = PyObject_GetAttrString(tex_filter, "post_load");
+  PyErr_Clear();
+
+  if (_pre_load_func == nullptr && _post_load_func == nullptr) {
+    PyErr_Format(PyExc_TypeError,
+                 "texture pool filter plug-in %R does not define pre_load or post_load function",
+                 tex_filter);
+    return false;
+  }
+
+  if (tex_pool != nullptr) {
+    tex_pool->register_filter(this);
+  }
+
+  return true;
+}
+
+/**
+ * Initializes the fields from the given Python texture pool object,
+ * then registers the filter at the global texture pool.
+ */
+bool PythonTexturePoolFilter::
+register_filter(PyObject *tex_filter) {
+  return register_filter(tex_filter, TexturePool::get_global_ptr());
+}
+
+/**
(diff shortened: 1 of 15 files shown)


rdb:
Based on a cursory scan, this looks great!  Thank you!

I'm only not entirely certain about the interface to register a filter.  Personally, I think this belongs on TexturePool, especially so that the interface for registering Python filters is similar to that of C++ filters.  It may be necessary to create an extension method for TexturePool on this; see loaderFileTypeRegistry.h and loaderFileTypeRegistry_ext.h.  Thoughts?

darktohka:
Would you recommend moving `_filter_registry` (`pvector<TexturePoolFilter *>`) out into a TexturePoolFilterRegistry class?

Would this be a `get_global_ptr` singleton class (shared between all TexturePool instances) or would a seperate instance be created for each `TexturePool`?

Right now, TexturePool only supports registering a filter (this is most likely because it was only used internally, for the TXA texture filter).

With a texture pool filter registry, we could also add functionality for unregistering filters.

My biggest concern is this method: `register_filter(PyObject *tex_filter, TexturePool *tex_pool)`, which requires us to provide a PyObject of the current instance of the Python class that implements the filter. Naturally, the dependence on the TexturePool will be removed. I'm not sure if we can run this automatically during the `__init__` phase of the class, though (I'm not that well versed with C++ Interrogate, but if you've got any pointers on this, I'd be really content).

`Loader.py` seems to work around this by automatically registering every class present in the `panda3d.loaders` package as a loader entry point. but I'd like to keep texture pool filters out of deploy_stub's functionality for now, if possible.

rdb:
I'm not sure that that a separate TexturePoolFilterRegistry is required; off the top of my head I don't see a reason why register_filter can't just live on TexturePool.

I am suggesting taking that `register_filter` method and putting it on TexturePool directly.  We can do that via a special EXTENSION-marked method, like:
```cpp
  EXTENSION(void register_filter(PyObject *tex_filter));
```
Interrogate will then call `Extension<TexturePool>::register_filter()` when calling from Python, which can be implemented in `texturePool_ext.h`, see the `loaderFileTypeRegistry_ext.h` for an example of this.

The Loader.py code is related to supporting pkg_resources entry points (so that `pip install panda3d-gltf` can automatically register the .gltf loader plug-in).  We can do that here but it is not necessary.

darktohka:
I see what you mean! Thanks for the clarification.

Right now, I've got final exams until Wednesday, but I'll get to work on the register_filter extension after those. I'll also work on an unregister_filter implementation, so that tests could also be written without having an effect on the execution of other tests.

darktohka:
I've implemented a different approach for registering filters.

First of all, `PythonTexturePoolFilter` has been unpublished. Texture pool filters implemented in Python do not have to inherit from `PythonTexturePoolFilter` anymore, they can simply inherit from `object` (as a new-style class).

TexturePool's get_global_ptr has been published in order to allow access to the new `register_filter` and `unregister_filter` methods.

Any registered texture pool filter can now be unregistered using `unregister_filter`. If the user attempts to unregister a filter that has not been registered yet, an `gobj` warning will be printed to the user.

Any attempt to register a filter that has already been registered will result in a similar `gobj` warning.

New methods `get_num_filters`, `get_filters` and `get_filter` have been added in order to allow more access to the texture pool filters.

PythonTexturePoolFilters are identified by the base class's Python hash. I tried to identify them by `pre_load` and `post_load`'s addresses at first, but this is not possible, as `PyObject_GetAttrString` will always return different addresses for the same object.

Here's the same texture pool filter example from before, edited to work with this new approach:
[texfilter_example_new.zip](https://github.com/panda3d/panda3d/files/4826395/texfilter_example_new.zip)

darktohka:
Tests have now been added for the texture pool filter implementation!

I've addded two new methods. `is_filter_registered` can be used to check if a filter has already been registered. On the other hand, `clear_filters` can be used to quickly unregister all filters.

Static implementations have been added for most of the new filter manipulation operations.

`get_filter` now returns nullptr (or None on Python) if an out of bounds index is given.

This PR is now at a mature point, I believe. Reviews are welcome! :)

panda/src/gobj/pythonTexturePoolFilter.cxx
```diff
@@ -0,0 +1,183 @@
...
+
+  // Wrap the arguments.
+  PyObject *args = PyTuple_Pack(6,
+    DTool_CreatePyInstance((void *)&orig_filename, Dtool_Filename, false, true),
+    DTool_CreatePyInstance((void *)&orig_alpha_filename, Dtool_Filename, false, true),
+    Dtool_WrapValue(primary_file_num_channels),
+    Dtool_WrapValue(alpha_file_channel),
+    Dtool_WrapValue(read_mipmaps),
+    DTool_CreatePyInstance((void *)&options, Dtool_LoaderOptions, false, true)
+  );
```
rdb, line 108: PyTuple_Pack increments the reference count on the arguments, and so does Dtool_WrapValue, so the read_mipmaps, alpha_file_channel, primary_file_num_channels values are leaking.

I suggest two possible routes to solve this:
1. (More convenient) use Py_BuildValue to create the tuple, which obviates the need for Dtool_WrapValue since you can use the format codes directly
2. (More efficient) first create the tuple using PyTuple_New(6) and then use PyTuple_SET_ITEM to set the items, which doesn't increment the reference count.  Note that DTool_CreatePyInstance does not (I believe) increment the reference count, so you do need to then manually do that.

panda/src/gobj/pythonTexturePoolFilter.cxx
```diff
@@ -0,0 +1,183 @@
...
+
+  PT(Texture) tex = nullptr;
```
rdb, line 110: Nitpick: PTs are null-initialized out of the box, so no need for an assignment here.

panda/src/gobj/pythonTexturePoolFilter.cxx
```diff
@@ -0,0 +1,183 @@
...
+  PyObject *result = PythonThread::call_python_func(_pre_load_func, args);
+
+  Py_DECREF(args);
+
+  if (result != nullptr) {
+    if (DtoolInstance_Check(result)) {
+      tex = (Texture *)DtoolInstance_UPCAST(result, Dtool_Texture);
+    }
+    Py_DECREF(result);
```
rdb, line 119: Could you clarify the behaviour here?  Anything returned that is not a Panda type is ignored?  Should we error for this?

panda/src/gobj/texturePool.cxx
```diff
@@ -66,13 +68,87 @@ register_texture_type(MakeTextureFunc *func, const string &extensions) {
  * Records a TexturePoolFilter object that may operate on texture images as
  * they are loaded from disk.
  */
+bool TexturePool::
+ns_register_filter(TexturePoolFilter *tex_filter) {
+  MutexHolder holder(_filter_lock);
+
+  // Make sure we haven't already registered this filter.
+  if (find(_filter_registry.begin(), _filter_registry.end(), tex_filter) != _filter_registry.end()) {
```
rdb, line 76: I'm sure there's a `using std::find;` somewhere that makes this work, I just can't find it… is a thirdparty library pulling this in?  Maybe use `std::find` to be safe.

panda/src/gobj/texturePool.h
```diff
@@ -80,18 +80,30 @@ class EXPCL_PANDA_GOBJ TexturePool {
...
   INLINE static PT(Texture) make_texture(const std::string &extension);
 
+  INLINE static bool register_filter(TexturePoolFilter *tex_filter);
+  INLINE static bool unregister_filter(TexturePoolFilter *tex_filter);
+  INLINE static void clear_filters();
+
+  INLINE static bool is_filter_registered(TexturePoolFilter *tex_filter);
+
+  INLINE static size_t get_num_filters();
+  INLINE static TexturePoolFilter *get_filter(size_t i);
```
rdb, line 90: Let's use a MAKE_SEQ_PROPERTY interface for this (and leave the get_num_filter/get_filter unpublished, so we have only the property interface)

panda/src/gobj/texturePool_ext.cxx
```diff
@@ -0,0 +1,110 @@
...
+        return filter;
+      }
+    }
+  }
+
+  // No filter found.
+  return nullptr;
+}
+
+#endif
```
rdb, line 110: Please configure your editor to add newlines at EOF.

panda/src/gobj/texturePool_ext.cxx
```diff
@@ -0,0 +1,110 @@
...
+  long filter_hash = PyObject_Hash(tex_filter);
+
+  for (size_t i = 0; i < num_filters; ++i) {
+    TexturePoolFilter *filter = _this->get_filter(i);
+
+    if (filter != nullptr && filter->is_of_type(PythonTexturePoolFilter::get_class_type())) {
+      PythonTexturePoolFilter *py_filter = (PythonTexturePoolFilter *)filter;
+
+      if (py_filter->_filter_hash == filter_hash) {
+        return filter;
```
rdb, line 101: Is this meant to compare against hash value?  Doesn't this risk hash collisions?

panda/src/gobj/texturePool_ext.h
```diff
@@ -0,0 +1,45 @@
...
+ * license.  You should have received a copy of this license along
+ * with this source code in a file named "LICENSE."
+ *
+ * @file texturePool_ext.h
+ * @author Derzsi Daniel
+ * @date 2020-06-24
+ */
+
+#ifndef TEXTUREPOOLEXT_H
+#define TEXTUREPOOLEXT_H
```
rdb, line 15: `TEXTUREPOOL_EXT_H`

panda/src/putil/loaderOptions.h
```diff
@@ -34,6 +34,7 @@ class EXPCL_PANDA_PUTIL LoaderOptions {
     LF_no_cache          = 0x0030,  // no_disk + no_ram
     LF_cache_only        = 0x0040,  // fail if not in cache
     LF_allow_instance    = 0x0080,  // returned pointer might be shared
+    LF_no_filters        = 0x0100,  // disallow using texture pool filters
```
rdb, line 37: Should this not be under TextureFlags?  Not sure of the implications one way or the other.

darktohka:
I've made the changes requested.

The no filter texture load flag is now under TextureFlags, rather than LoaderFlags. TF_no_filters is required in case the user wants to load a texture without running any texture filters on it. This is useful, for example, if the user wants to load another unrelated texture in pre_load or post_load (as running the texture filter on it would result in the texture being loaded ad infinitum - a stack overflow). This stack overflow is actually prevented by the _filter_lock, which prevents running any concurrent filters as the filter lock is currently held.

TEXTUREPOOLEXT_H has been renamed to TEXTUREPOOL_EXT_H.

I decided to save the Python texture pool filter object as _entry_point, replacing the _filter_hash with a PyObject* pointer. I increment the reference count of it when I set it as the _entry_point, and only decrement the reference count when the PythonTexturePoolFilter is destructed. This way, we don't have to use hashes. The reason why my pointers were always different is because PyObject_GetAttrString always returns a different pointer, and I was comparing not the object pointer itself, but the attribute pointers (which are always different).

Here is the new revision of the example texture pool filter:

[texfilter_example_v3.zip](https://github.com/panda3d/panda3d/files/4956971/texfilter_example_v3.zip)

rdb:
Merged.  This is a great addition, thank you so much for working on it!

PR #955, opened 2020-06-15
Fixed regression from removing choice function from PythonUtils

fireclawthefox:
Replaced removed method with normal python style
Add sanity check for _crashOnProactiveLeakDetect, don't crash by default

### DIFF
--- a/direct/src/showbase/DirectObject.py
+++ b/direct/src/showbase/DirectObject.py
@@ -94,12 +94,14 @@ class DirectObject:
         if hasattr(self, '_taskList'):
             tasks = [task.name for task in self._taskList.values()]
         if len(events) or len(tasks):
-            estr = choice(len(events), 'listening to events: %s' % events, '')
-            andStr = choice(len(events) and len(tasks), ' and ', '')
-            tstr = choice(len(tasks), '%srunning tasks: %s' % (andStr, tasks), '')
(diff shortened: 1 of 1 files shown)


direct/src/showbase/DirectObject.py
```diff
@@ -94,12 +94,14 @@ def detectLeaks(self):
...
-            tstr = choice(len(tasks), '%srunning tasks: %s' % (andStr, tasks), '')
+            estr = ('listening to events: %s' % events if len(events) else '')
+            andStr = (' and ' if len(events) and len(tasks) else '')
+            tstr = ('%srunning tasks: %s' % (andStr, tasks) if len(tasks) else '')
             notify = directNotify.newCategory('LeakDetect')
-            func = choice(getRepository()._crashOnProactiveLeakDetect,
-                          self.notify.error, self.notify.warning)
+            crash = False
+            if hasattr(getRepository(), '_crashOnProactiveLeakDetect'):
+                crash = getRepository()._crashOnProactiveLeakDetect
```
rdb, line 103: You can just replace these three lines with `crash = getattr(getRepository(), '_crashOnProactiveLeakDetect', False)`

fireclawthefox: Thanks for the hint. Changed it accordingly.

PR #965, opened 2020-06-25
Add initial issue templates

Kupoman:
## Issue description
This repository currently lacks issue templates as recommended by the [Community Profile](https://github.com/panda3d/panda3d/community).

## Solution description
I put together three simple issue templates for bugs, enhancement requests, and discussion issues. The discussion template will automatically apply the "discussion" label. No other labels are applied automatically. If these templates are not sufficient for a user, they can still select the option to use a blank template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,36 @@
+
+---
+name: Bug report
+about: Report a defect
+title: ''
+labels: 
+assignees: ''
+
+---
+
+## Environment
+ * OS:
+ * Panda3D version:
+ * Python version (if using Python):
+ * Compiler (if using C++):
+
+## Summary
+A clear and concise description of what the bug is.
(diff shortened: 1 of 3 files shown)


darktohka:
`enhacnement.md` should be renamed to `enhancement.md` 

Otherwise, good work! :)

Kupoman:
Oops, good catch. Fixed.

Moguri:
The bug template does not address C++ users. I think it's fine to assume Python for an example stub, but we may want to do something a little different for environment. Maybe something like:
* Python version (if using Python):
* Compiler (if using C++):

Kupoman:
@Moguri, good point about C++ users. I have updated the bug templates as you suggested.

rdb:
Pushed with some changes.  I removed the code block.  We don't want people to start writing their reproduction case inside the GitHub issue form (and certainly don't want to encourage adding unnecessary abstractions such as subclassing ShowBase if that's not necessary for reproduction).

Thanks for your initiative!

PR #967, opened 2020-06-30
cocoadisplay: Fix handling of shift modifier

### LINKED ISSUE #959: Shift + number keys sends wrong events on macOS
When pressing shift+2, instead of sending the "shift-2" event like on other systems, the "shift-@" event is sent on macOS.

This is because the code relies on `[NSEvent charactersIgnoringModifiers]`, which does not ignore the shift modifier specifically.

treamology:
This replaces the use of [NSEvent charactersIgnoringModifiers] with a Carbon function. In macOS 10.15+ there is a way in Foundation of achieving the same thing, so we'll be able to switch to that once Carbon is fully removed from macOS.

Fixes #959

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -1720,20 +1720,19 @@ handle_key_event(NSEvent *event) {
     return;
   }
 
+  TISInputSourceRef input_source = TISCopyCurrentKeyboardLayoutInputSource();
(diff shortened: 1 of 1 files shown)


rdb:
Thanks for working on this!

This fixes shift modifiers, but it seems to break certain alt + letter combinations, giving messages like:
```
Unhandled keypress, character 252, keyCode 3, type 10, flags 524576
```

treamology:
Looks like I needed to reset the dead key state even though I had specified `kUCKeyTranslateNoDeadKeysMask`. Should be fixed now.

PR #973, opened 2020-07-15
joystick: Fix erroneous dpad_*-up events when emulating a dpad on Linux

Moguri:
## Issue Description
When emulating a dpad using the hat axes, an change in an axis will trigger updates for both associated dpad "buttons" (i.e., left/right or up/down). The depressed button is updated as "down" and the other button is updated to "up." However, since these buttons start in an unknown state and a state change triggers an event to fire, the first time an emulated dpad button is updated to "down" the opposite button will throw an "up" event.

## Solution Description
Start emulated dpad buttons in an S_up state instead of S_unknown. This way, the first S_up event will go from S_up to S_up instead of S_unknown to S_up and not trigger an event to fire.

This PR applies this fix for the Linux Joystick API and for the Linux evdev API, but only the evdev changes were tested. I do not know if other platforms have similar issues.

### DIFF
--- a/panda/src/device/evdevInputDevice.cxx
+++ b/panda/src/device/evdevInputDevice.cxx
@@ -595,6 +595,8 @@ init_device() {
               _buttons.push_back(ButtonState(GamepadButton::hat_left()));
               _buttons.push_back(ButtonState(GamepadButton::hat_right()));
             }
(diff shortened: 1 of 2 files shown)


Moguri:
PR description updated to more clearly describe the issue and the proposed solution.

PR #975, opened 2020-07-22
direct: Fix TexMemWatcher crash when graphics memory reaches 1 GB

darktohka:
TexMemWatcher crashes frequently when the graphics memory approaches or reaches 1 GB.

The first crash happens when two texture match dimensions are the same. Python tries to compare the texture placements, but this is impossible as they are not comparable, therefore the following traceback occurs:
```
  File "C:\Panda3D-1.11.0\direct\showutil\TexMemWatcher.py", line 787, in placeTexture
    tp = self.findHole(tr.area, tr.w, tr.h)
  File "C:\Panda3D-1.11.0\direct\showutil\TexMemWatcher.py", line 886, in findHole
    return max(matches)[1]
TypeError: '>' not supported between instances of 'TexPlacement' and 'TexPlacement'
```

The second crash happens because of Texture's `up_to_power_2`. Unfortunately, `up_to_power_2` returns a negative number as soon as it reaches 1073741824, because the next power of two, 2147483648, does not fit into an integer.
```
  File "C:\Panda3D-1.11.0\direct\showutil\TexMemWatcher.py", line 335, in __doSetLimit
    self.win.getGsg().getPreparedObjects().setGraphicsMemoryLimit(self.limit)
OverflowError: can't convert negative value to size_t
```

I've changed `up_to_power_2` and `down_to_power_2` to operate on `size_t` instead. All usages of these have been downcasted to int where necessary for DirectX APIs or texture sizes, which are of the `int` type.

This means that `up_to_power_2` and `down_to_power_2` now support all numbers from 0 to 9223372036854775807 (2**63 - 1) (as bit-counting methods only exist up to uint64_t).

The TexMemWatcher now works properly when 1 GB of texture memory is reached.

### DIFF
--- a/direct/src/showutil/TexMemWatcher.py
+++ b/direct/src/showutil/TexMemWatcher.py
@@ -883,7 +883,7 @@ class TexMemWatcher(DirectObject):
             matches.append((match, tp))
 
         if matches:
-            return max(matches)[1]
+            return max(matches, key=lambda match: match[0])[1]
         return None
 
     def findHolePieces(self, area):
@@ -937,7 +937,7 @@ class TexMemWatcher(DirectObject):
     def findLargestHole(self):
         holes = self.findAvailableHoles(0)
         if holes:
-            return max(holes)[1]
+            return max(holes, key=lambda hole: hole[0])[1]
         return None
 
     def findAvailableHoles(self, area, w = None, h = None):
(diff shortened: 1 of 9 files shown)


rdb:
I'm going to push a slightly different fix that does not rely on changing up_to_power_2, so that I can put this in 1.10.7 without ABI changes.

I think `Texture::up_to_power_2` should be meant for texture dimensions (which always comfortably fit in an int) and it's weird to have code use a Texture method to align memory.

PR #977, opened 2020-07-28
OnScreenDebug: Parent to a2dTopLeft as opposed to aspect2d

drewc5131:
The OSD hasn't been touched in well over a decade, so it could definitely be adjusted to better suit the current state of the engine.

Currently, the OSD is just parented to aspect2d with a position of -1.0, 0.9, whereas parenting it directly to a2dTopLeft would make much more sense and would look better on all screens.

### DIFF
--- a/direct/src/showbase/OnScreenDebug.py
+++ b/direct/src/showbase/OnScreenDebug.py
@@ -37,10 +37,10 @@ class OnScreenDebug:
         if not font.isValid():
             print("failed to load OnScreenDebug font %s" % fontPath)
             font = TextNode.getDefaultFont()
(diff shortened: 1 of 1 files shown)


rdb:
What is OnScreenDebug and how do I enable it in order to test this change?

drewc5131:
on-screen-debug-enabled 1
its a bulitin so from anywhere you can call `onScreenDebug.add("Key Here", "Value Here")`

An example use case is in GravityWalker 
https://github.com/panda3d/panda3d/blob/master/direct/src/controls/GravityWalker.py#L384
This is called every frame. Adding something that already exists will simply change the value and wont just add multiple. Its something that can be very useful for debugging and i'm surprised its not more known

rdb (approved):
OK, why not.

Issue #978, opened 2020-08-01
AppImage support


cflavio:
Add an *AppImage* installer to the current deployment tools.

AppImage offers several advantages as a deployment solution. Specifically, the user doesn't need to install the Panda3D application. She can download the file, make it executable, and run it. It may be a good solution for games, where the installation may be a barrier for trying them.

PR #979, opened 2020-08-01
appimage support

cflavio:
use it by adding 'appimage' to the installers in setup.py
e.g.
… 'bdist_apps': {… 'installers': … 'manylinux1_x86_64': [… 'appimage']} …

This is a proposal for the issue #978.

The solution just adds the AppImage creation to the current deployment options.

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -18,6 +18,8 @@ import imp
 import string
 import time
 import tempfile
+import urllib.request
+import platform as py_platform
 
 import setuptools
 import distutils.log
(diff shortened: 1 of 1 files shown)


Moguri:
This doesn't look like it will work if run from a non-Linux platform. This would be the first installer that requires a specific platform to build and would break the "build for all platforms from any platform" paradigm that these deployment tools strive for.

el-dee:
> This doesn't look like it will work if run from a non-Linux platform. This would be the first installer that requires a specific platform to build and would break the "build for all platforms from any platform" paradigm that these deployment tools strive for.

The NSIS installer does actually require that you run it on Windows platform, so it's not a first :

https://github.com/panda3d/panda3d/blob/abe761c30ec2c4d5d2c17616bbd725e3452a9ac9/direct/src/dist/commands.py#L1426

(Though I think you can install makensis on Linux (and perhaps on macOS using brew), but I never tested it)

For me, I don't think it's a problem to require to use a given platform to build a specific installer for that platform. If one want to stay generic, one could still use the .zip or .tgz targets.

Moguri:
That check is for the target platform, not the host platform. As you said, you can run makensis on Linux (I haven't really looked into macOS). 

I should have mentioned this in my original comment, but I am not opposed to the idea of having some options limited by the host OS. However, we should only go this route after exploring other options and only if we intentionally decide to break -- or otherwise make an exception  to -- the current cross-platform goals of the deployment system.

el-dee:
> That check is for the target platform, not the host platform. As you said, you can run makensis on Linux (I haven't really looked into macOS).

Ah you're right, I forgot my own rule "never comment when it's too late" :)

> I should have mentioned this in my original comment, but I am not opposed to the idea of having some options limited by the host OS. However, we should only go this route after exploring other options and only if we intentionally decide to break -- or otherwise make an exception to -- the current cross-platform goals of the deployment system.

Fair enough (I'm actually biased as my own code relies on native code for the target platform, so I'm already paying the price of running the deployment system on each target platform)

cflavio:
I am sorry, obviously I didn't want to break any Panda3D's paradigm. I apologize for this: I accidentally missed this point. Actually, the cross-platform goal is very important, so I agree that this installer would be an issue.

Moguri:
@cflavio Are you aware of any cross-platform ways to build an AppImage?

@rdb Any thoughts on installers that can only be run on certain platforms?

If we cannot find a cross-platform way to build AppImages, then I am inclined to reject this PR, but I could be persuaded otherwise. AppImage support *would* be very nice to have.

cflavio:
> @cflavio Are you aware of any cross-platform ways to build an AppImage?

@Moguri I have read that some results may be obtained by using WSL on Windows, anyway this would not be a many-to-many approach like the Panda3D's one (this would exclude non-WSL buildings).

> If we cannot find a cross-platform way to build AppImages, then I am inclined to reject this PR, but I could be persuaded otherwise. AppImage support _would_ be very nice to have.

I agree with this, perhaps this is not the right place for it since it would disrupt the current Panda3D paradigm.

rdb:
I appreciate the work that has gone into this but I think this would be best published as an external tool, as it stands, due to the fact that it is not platform-independent.

PR #988, opened 2020-08-11
Fix invalid constant used to detect infinity and NaN when code is compiled with finite-math-only

### LINKED ISSUE #987: Setting far plane to inf does not work with STDFLOAT_DOUBLE=1 builds
## Description

When using Panda3D built with STDFLOAT_DOUBLE=1 setting the far plane of the camera to infinity makes all the objects in the scene to be culled out.

## Steps to Reproduce

* Install Panda3D build with STDFLOAT_DOUBLE=1
* Add to any samples the following line before base.run()

      base.camLens.setNearFar(0.1, math.inf)

* Run the sample, nothing except 2D stuff is rendered.

If one prints the projection matrix, the result is a null matrix

<pre>
0 0 0 0
0 0 0 0
0 0 0 0
0 0 0 0
</pre>

## Environment
 * Operating system: Linux
 * System architecture: x86_64
 * Panda3D version: 1.10.6
 * Installation method: built from source
 * Python version (if using Python): 3.7 and 3.8
 * Compiler (if using C++): N/A

el-dee:
## Issue description

This PR fixes #987. The  constant used to detect infinite (and also NaN) for double precision value when using code compiled with finite-math-only (implied by optimize=4) is wrong.

Current code is using 0x7ff0000000000000ull as the representation of infinity after a unit left shift. However, the actual representation of infinity is 0x7FF0000000000000, so after a unit left shift it is 0xffe0000000000000ull

Note that the same problem is present in cnan(), a NaN is 0x7FFxxxxxxxxxxxxxx

## Solution description

Change the constant in cinf(double v) and cnan(double v) in cmath.I

## Checklist

It seems that there are no unit tests covering dtoolbase (unless I'm mistaken), this could have been detected there...

### DIFF
--- a/dtool/src/dtoolbase/cmath.I
+++ b/dtool/src/dtoolbase/cmath.I
@@ -351,7 +351,7 @@ cnan(double v) {
 #if __FINITE_MATH_ONLY__
   // GCC's isnan breaks when using -ffast-math.
   union { double d; uint64_t x; } u = { v };
-  return ((u.x << 1) > 0xff70000000000000ull);
+  return ((u.x << 1) > 0xffe0000000000000ull);
 #elif !defined(_WIN32)
   return std::isnan(v);
 #else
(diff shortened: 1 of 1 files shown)


rdb:
Thanks!

I added in some unit tests in 9ac1a4f7cb3697e74e561d6d7f0d02131b55fd33.

PR #989, opened 2020-08-15
Create PyLint configuration

Kupoman:
## Issue description
I have noticed that we have several pull requests related to Direct modules. In order to help move these along quicker I thought it would be useful to have a linter configured for Direct code.

## Solution description
I have created a PyLint configuration that I mostly left at the defaults except for the following changes:
* panda3d is added to extension-pkg-whitelist
* invalid-name, missing-*-docstring, and refactor checks are disabled
* taskMgr,loader,base,directNotify are added to additional-builtins

To test out this configuration, I modified the code in direct/src/filter until it passed a run of PyLint. I have not yet modified the CI/CD pipeline to run the linter.

## Checklist

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,521 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=panda3d
+
+# Specify a score threshold to be exceeded before program exits with error.
+fail-under=10
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+limit-inference-results=100
+
+# List of plugins (as comma separated values of python module names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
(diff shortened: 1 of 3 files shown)


direct/src/filter/FilterManager.py
```diff
@@ -307,29 +329,47 @@ def createBuffer(self, name, xsize, ysize, texgroup, depthbits=1, fbprops=None):
...
+                GraphicsOutput.RTPAuxRgba1
+            )
         buffer.setSort(self.nextsort)
         buffer.disableClears()
         self.nextsort += 1
         return buffer
 
+    # Disable unused variable warning to avoid API change
+    # pylint: disable=unused-argument
     def windowEvent(self, win):
```
Moguri, line 373: You can change `win` to `_win` to mark the variable as intentionally unused without breaking the API. The method will still take two arguments, and since no keyword arguments are used, renaming the argument should not be a problem either.

Kupoman: Thanks, forgot about that trick. Fixed.

rdb:
Does direct pass these pylint rules as-is?

Kupoman:
Just src/direct/filter. I should have added in the description that I started with filter as a test. If there is enough buy in on this linter and configuration, I will keep going through the rest of direct. I was thinking about doing a PR per direct module to keep them easier to review.

.pylintrc
```diff
@@ -0,0 +1,521 @@
...
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
```
Moguri, line 25: Should we set this to 0?

Moguri (commented):
Overall looks good to me, and I would be okay with using this configuration to update the rest of direct.

If direct has a habit of not defining things in `__init__()` I wonder what kind of weird corners we may get into if we start doing this. Could someone have relied on these attributes being undefined?

Moguri:
@rdb Any thoughts on this? If we're fine with the settings, then I think we can go ahead and merge this now and wait to enable linting as part of CI until all of direct has been fixed up.

@Kupoman there is still an unresolved conversation open on the number of PyLint jobs.

rdb:
Thanks for this!  I have been playing with this and so far I've already found a bunch of obscured bugs in the direct tree that pylint uncovered, so I'm happy with continuing with this.  Merged.

I also did a cleanup pass over direct, fixing the majority but not all of the warnings.  People should feel free to help out and make additional clean-up PRs.  I'm open to adding additional disables if that would help.

I did add in a bunch more disables that I don't think should be high priority right now, but once we get the codebase cleaned up a bit more we can consider making it stricter again.

I didn't check in some of the line length changes in CommonFilters because I don't think they help readability as much as, say, switching CommonFilters to use f-strings instead of percent-interpolation.

PR #990, opened 2020-08-16
Move speedtree code to contrib

Moguri:
## Issue description
SpeedTree integration with Panda3D is unmaintained, and there is little desire to support it with the CMake migration (see issue #440)

## Solution description
Instead of removing the code (as PR #901 does) we could instead move the logic to contrib. This gives users access to the old code, but signifies that the code is no longer supported. Also, it means we do not have to add CMake support (users have access to the code "as-is"). This PR *does* update makepanda, but, as I do not have a SpeedTree license, compiling was not tested (outside of checking that compiling without SpeedTree still works).

## Checklist


rdb:
Should this not be `contrib/src/speedtree`, to match the style of the rest of the tree?

Moguri:
:facepalm: 

I thought I had checked `contrib` and was surprised that it *didn't* have a `src` directory. Turns out I am just blind.

Fixed by a force push.

Moguri:
Forced pushed the symbol changes.

Issue #745, opened 2019-09-25
Feature request: Compressing embedded texture data in .bam


rdb:
At the moment, embedded texture data in .bam files is stored uncompressed.  This makes these .bam files unnecessarily massive.  It would be useful if Panda could use a lossless compression algorithm, such as PNG or ZLib, to store embedded textures.

ChinYing-Li:
Does it make sense to use the existing class handle of zlib compression in this issue?

rdb:
I'm not entirely sure what you mean by "the existing class handle of zlib compression".  We do already pull in the zlib library for other purposes in our C++ codebase, so we can use zlib functions if that's what you mean.

ChinYing-Li:
@rdb  Thanks for confirmation\. I would like to work on this.

Moguri:
I suspect our main concern for compression algorithms/libraries for BAM is decompression speed. There are better options for decompression speed than zlib (e.g., lz4 looks nice for decompression). However, zlib doesn't seem terrible for this, and it has the benefit that we already depend on zlib.

If/when someone gets around to a PR, benchmarks would be much appreciated so we can make sure we're not hurting BAM read performance with decompression.

rdb:
I agree, it would be nice to see some comparisons of both size and decompression speed of the image data when (1) uncompressed, (2) compressed with zlib, and (3) compressed with a lossless image format like PNG.

If zlib proves to provide adequate compression speed without too much performance loss, we can make the zlib compression strength adjustable with a Config.prc value.

@ChinYing-Li You are welcome to work on this.  Let me know if you need any assistance (or feel free to ask any questions on our Discord server in the #development channel).

Clockwork-Muse:
... okay, I'll bite.  Why a _lossless_ format?  If a `.bam` is being created, the asset is being baked for distribution, and in most cases an appropriate GPU-preferred (lossy) compression format can be chosen.  This is essentially what I'm doing now: packing a `.txo.pz` +`.bam` into a `.mf` for single-file distribution.  At the very least, having the option to use the existing (already compressed) texture stream would cover most use cases.

Moguri:
I can't speak for why @rdb mentioned a lossless format, but I can talk about why I prefer one. Panda should always preserve user data (especially in automatic operations). If a user is okay with losing information, they should explicitly say so. This could mean using an already lossy format (e.g., DDS), setting a flag on the texture, or using some form of global flag.

Clockwork-Muse:
[...]

PR #992, opened 2020-08-18
Allow texture compression in .bam (WIP)

ChinYing-Li:
## Issue description

This PR is aimed to resolve issue #745 .

## Solution description

This is still a work in progress and a bug probably exists, as turning `bam-tex-compression-level` to `on` doesn't seems to make the .bam file smaller.
Also, as suggested by @Moguri , I checked out the comparison between zlib and other algorithms. Since the decompression speed is of concern, I think lz4 is better suited for the issue.

## Checklist

### DIFF
--- a/panda/src/gobj/texture.cxx
+++ b/panda/src/gobj/texture.cxx
@@ -10059,10 +10059,6 @@ do_write_datagram_rawdata(CData *cdata, BamWriter *manager, Datagram &me) {
   me.add_uint8(cdata->_component_width);
   me.add_uint8(cdata->_ram_image_compression);
 
-  if(manager->get_file_minor_ver() >= 46) {
-    me.add_uint8(manager->get_tex_compression_format());
-  }
-
   if (cdata->_ram_images.empty() && cdata->_has_clear_color &&
       manager->get_file_minor_ver() < 45) {
     // For older .bam versions that don't support clear colors, make up a RAM
--- a/panda/src/gobj/texture.cxx
+++ b/panda/src/gobj/texture.cxx
@@ -9848,7 +9848,7 @@ write_datagram(BamWriter *manager, Datagram &me) {
 
   // If we are also including the texture's image data, then stuff it in here.
   if (has_rawdata) {
-    (cdata, manager, me);
+    do_write_datagram_rawdata(cdata, manager, me);
   }
 }
 
@@ -10060,7 +10060,7 @@ do_write_datagram_rawdata(CData *cdata, BamWriter *manager, Datagram &me) {
   me.add_uint8(cdata->_ram_image_compression);
 
   if(manager->get_file_minor_ver() >= 46) {
-    me.add_uint8(manager->get_tex_compress_level());
+    me.add_uint8(manager->get_tex_compression_format());
   }
 
   if (cdata->_ram_images.empty() && cdata->_has_clear_color &&
@@ -10086,7 +10086,7 @@ do_write_datagram_rawdata(CData *cdata, BamWriter *manager, Datagram &me) {
       me.add_uint32(cdata->_ram_images[n]._page_size);
 
       // Only write the size of the image data written to bam.
-      if(manager->get_tex_compress_level() == BamEnums::BTC_on) {
+      if(manager->get_tex_compression_format() == BamEnums::BTC_zlib) {
         string raw_image = cdata->_ram_images[n]._image.get_data();
         string compressed_image = compress_string(raw_image, 6);
         me.add_uint32(compressed_image.length());
@@ -10379,7 +10379,7 @@ do_fillin_rawdata(CData *cdata, DatagramIterator &scan, BamReader *manager) {
     PTA_uchar image_data = PTA_uchar::empty_array(read_size, get_class_type());
     scan.extract_bytes(image_data.p(), read_size);
 
-    if(manager->get_texture_compression_level() != BamEnums::BTC_off) {
+    if(manager->get_tex_compression_format() == BamEnums::BTC_zlib) {
       string decompressed = decompress_string(image_data.get_data());
       image_data.resize(decompressed.length());
       image_data.set_data(decompressed);
--- a/panda/src/gobj/texture.cxx
+++ b/panda/src/gobj/texture.cxx
@@ -15,6 +15,7 @@
 
 #include "pandabase.h"
 #include "texture.h"
+#include "compress_string.h"
 #include "config_gobj.h"
 #include "config_putil.h"
 #include "texturePool.h"
@@ -47,8 +48,10 @@
 #include <squish.h>
 #endif  // HAVE_SQUISH
 
+#include <algorithm>
 #include <stddef.h>
 
+using std::copy;
 using std::endl;
 using std::istream;
 using std::max;
@@ -9845,7 +9848,7 @@ write_datagram(BamWriter *manager, Datagram &me) {
 
   // If we are also including the texture's image data, then stuff it in here.
   if (has_rawdata) {
-    do_write_datagram_rawdata(cdata, manager, me);
+    (cdata, manager, me);
   }
 }
 
@@ -10056,6 +10059,10 @@ do_write_datagram_rawdata(CData *cdata, BamWriter *manager, Datagram &me) {
   me.add_uint8(cdata->_component_width);
   me.add_uint8(cdata->_ram_image_compression);
 
+  if(manager->get_file_minor_ver() >= 46) {
+    me.add_uint8(manager->get_tex_compress_level());
+  }
+
   if (cdata->_ram_images.empty() && cdata->_has_clear_color &&
       manager->get_file_minor_ver() < 45) {
     // For older .bam versions that don't support clear colors, make up a RAM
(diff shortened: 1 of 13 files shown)


panda/src/gobj/texture.cxx
```diff
@@ -9845,7 +9848,7 @@ write_datagram(BamWriter *manager, Datagram &me) {
 
   // If we are also including the texture's image data, then stuff it in here.
   if (has_rawdata) {
-    do_write_datagram_rawdata(cdata, manager, me);
+    (cdata, manager, me);
```
Moguri, line 9851: Was this change intentional? What is the goal with this change?

panda/src/gobj/texture.cxx
```diff
@@ -10077,8 +10084,17 @@ do_write_datagram_rawdata(CData *cdata, BamWriter *manager, Datagram &me) {
...
+      // Only write the size of the image data written to bam.
+      if(manager->get_tex_compress_level() == BamEnums::BTC_on) {
+        string raw_image = cdata->_ram_images[n]._image.get_data();
+        string compressed_image = compress_string(raw_image, 6);
+        me.add_uint32(compressed_image.length());
+        me.append_data(compressed_image.data(), compressed_image.length());
+      } else {
+        me.add_uint32(cdata->_ram_images[n]._image.size());
+        me.append_data(cdata->_ram_images[n]._image, cdata->_ram_images[n]._image.size());
+      }
```
Moguri, line 10097: We should throw a meaningful error if the compression format/level requested is not supported.

panda/src/gobj/texture.cxx
```diff
@@ -10349,20 +10365,27 @@ do_fillin_rawdata(CData *cdata, DatagramIterator &scan, BamReader *manager) {
...
         << "RAM image " << n << " extends past end of datagram, is texture corrupt?\n";
       return;
     }
 
-    PTA_uchar image = PTA_uchar::empty_array(u_size, get_class_type());
-    scan.extract_bytes(image.p(), u_size);
+    PTA_uchar image_data = PTA_uchar::empty_array(read_size, get_class_type());
+    scan.extract_bytes(image_data.p(), read_size);
+
+    if(manager->get_texture_compression_level() != BamEnums::BTC_off) {
```
Moguri, line 10382: We should explicitly check for BTC_on here to make it easier to add new formats in the future. Also, we should throw a meaningful error if the compression format/level requested is not supported.

panda/src/gobj/texture.h
```diff
@@ -749,6 +749,7 @@ class EXPCL_PANDA_GOBJ Texture : public TypedWritableReferenceCount, public Nama
   bool do_has_compression(const CData *cdata) const;
   virtual bool do_has_ram_image(const CData *cdata) const;
   virtual bool do_has_uncompressed_ram_image(const CData *cdata) const;
+
```
Moguri, line 752: Since this is the only change to this file and it's just a whitespace change, I would prefer to see this change omitted from the PR to keep it smaller.

panda/src/putil/bamEnums.h
```diff
@@ -69,6 +69,12 @@ class EXPCL_PANDA_PUTIL BamEnums {
     BTM_basename,
     BTM_rawdata
   };
+
+  // Level of compression for textures in the .bam
+  enum BamTexCompressionLv{
+    BTC_off,
+    BTC_on,
```
Moguri, line 76: I would prefer BTC_zlib over BTC_on to make it easier to add other formats in the future (if desired).

panda/src/putil/bamEnums.h
```diff
@@ -69,6 +69,12 @@ class EXPCL_PANDA_PUTIL BamEnums {
     BTM_basename,
     BTM_rawdata
   };
+
+  // Level of compression for textures in the .bam
```
Moguri, line 73: Is this truly a compression level, or just the compression format/algorithm?

panda/src/putil/bamReader.cxx
```diff
@@ -133,6 +133,10 @@ init() {
     _file_stdfloat_double = scan.get_bool();
   }
 
+  if(_file_minor >= 46) {
+    _tex_compression_level = (BamTexCompressionLv)scan.get_uint8();
```
Moguri, line 137: We should throw a meaningful error if the compression format/level requested is not supported.

panda/src/putil/bamEnums.h
ChinYing-Li: In my opinion, the enums should both indicate the compression algo and the compression level in the future e.g. `BTC_zlib_5`. But you are right, the comment is misleading.

panda/src/gobj/texture.cxx
ChinYing-Li: Do we have to throw errors in both `texture.cxx` and `bamReader.cxx`? Throwing error in `bamReader::init()` should be enough.

ChinYing-Li:
Now writing bam with compressed texture rawdata works as expected, here's the test script:
```
from panda3d.core import loadPrcFile
loadPrcFile("/home/cyli/repo/panda3d_testscript/Config.prc")
from panda3d.core import ConfigVariableManager

class Test(ShowBase):
    def __init__(self):
        ShowBase.__init__(self)
        model = self.loader.loadModel("/home/cyli/repo/panda3d/dmodels/src/misc/smiley.egg")
        model.reparentTo(self.render)
        self.render.write_bam_file("texturecompress.bam")
        loadedmodel = self.loader.loadModel("texturecompress.bam")

if __name__ == "__main__":
    testapp = Test()
    # cvMgr = ConfigVariableManager.getGlobalPtr()
    # cvMgr.listVariables()
    testapp.run()

```

Looking around on Discourse, it seems like there's no one liner to check if `model` and `loadedmodel` are equivalent. Did I miss anything?

ChinYing-Li:
@Moguri Here are the results for three models from gltf-sample repo, using zlib and the compression level is set to 6 (time unit is second, file size unit is MB):
- MetalSphere

| empty  | with texture compression | w/o texture compression |
| ------------- | ------------- | ------------- |
| File size | 15.4 | 23.7 |
| write time |  0.125151634216309| 0.092681964238485 |
| load time  | 0.217965364456177 | 0.163658459981283 |

- Lantern

| empty | with texture compression | w/o texture compression |
| ------------- | ------------- | ------------- |
| File size | 9.3 |54.8 |
| write time | 1.91940752665202 | 0.19003359476725 |
| load time  |  2.18315649032593| 0.402493715286255|

- Corset

| empty  | with texture compression | w/o texture compression |
| ------------- | ------------- | ------------- |
| File size | 14.4 | 38.4 |
| write time |  2.02993027369181| 0.129932324091507 |
| load time  | 2.35624472300212 | 0.237783988316854|

panda/src/putil/bamReader.cxx
```diff
@@ -133,6 +133,18 @@ init() {
     _file_stdfloat_double = scan.get_bool();
   }
 
+  if(_file_minor >= 46) {
+    _tex_compression_format = (BamTexCompressionFormat)scan.get_uint8();
+  }
+
+  if(_tex_compression_format == BTC_zlib) {
+#ifndef HAVE_ZLIB
```
darktohka, line 141: Maybe the guard should be moved to encompass the entire `if(_tex_compression_format == BTC_zlib)` block?

panda/src/putil/bamEnums.h
darktohka: Can you add a space before the `{` here, to adhere to the current code style?

Moguri:
Yeah, those load time hits are not ideal, but I think they are okay for an initial implementation. I believe we can look into lz4 in a separate PR (as discussed before).

panda/src/gobj/texture.cxx
Moguri: I would say yes. This keeps the code better de-coupled. If something changes in `bamReader::init()` that allows us to get to texture.cxx with bad data, then we'll still want to error. Or, if someone finds some way to call into the texture.cxx code without first going through `bamReady::init()`.

rdb (changes requested):
Looks great overall, thanks so much for working on this!

Some minor points:
* In a docstring you refer to a variable called `compress-texture-to-bam`, but this doesn't exist.
* Don't abbreviate "texture" to "tex".  It's not abbreviated in other variables either, so it would seem inconsistent.
* There needs to be a space between `if` and `(`
* According to Panda coding convention, all if/else blocks need `{}`

Thanks!

ChinYing-Li:
@kamgha Sorry for getting back late. By "w/o texture compression + pzip -6", do you mean you want to see "the time to make a .bam with and withoug compressing texture" + "the time to use pzip to further compress the bam"?

kamgha:
> 
> 
> @kamgha Sorry for getting back late. By "w/o texture compression + pzip -6", do you mean you want to see "the time to make a .bam with and withoug compressing texture" + "the time to use pzip to further compress the bam"?

We can probably achieve a similar result (uncompressed model data + zlib compressed texture data) by saving the texture data into a separte file (.txo) and `pzip` it (.txo.pz) and model data (.bam) uncompresed alongside it. We could, if we wanted to have both in a single file, put them (.txo + .bam) into a MultiFile (.mf) and specify to compress the .txo only, but not the .bam.

The questions are (which would add two columns in that table above from 23 Aug):
1. How the above combination with the already existing tools performs (.txo.pz + .bam, or .mf with compression on .txo and w/o on .bam).
2. Whether leaving the model data uncompressed makes a noticeable difference over compressing the whole .bam including both texture and model data (.bam.pz, or putting into a .mf with compression enabled).

Overall, it'd test whether the feature request in #745 is worth the added complexity and/or whether the desired effect can already be accomplished (considering zlib).


> see "the time to make a .bam with and withoug compressing texture" + "the time to use pzip to further compress the bam"?

The decompression times are of a higher interest here. You could try zlib with the highest compression level of 9 as well. It may give faster decompression times.

PR #995, opened 2020-08-24
putil: Make UniqueIdAllocator safe for multithreaded usage

darktohka:
## Issue description

UniqueIdAllocator is currently unsafe to use in multithreaded environments.

This class is used by default both by server implementations found in the direct package and third-party server implementations created by the community. When coupled with a multi-threaded network task chain, usage of the UniqueIdAllocator as-is can result in IDs being allocated twice as the result of a race condition.

This actually happened in practice, hence why this PR was created.

## Solution description

This pull request adds a simple Mutex to UniqueIdAllocator, which is held by all functions that access the `_table` used for keeping track of unique IDs and all functions that modify the `_table` or any unique ID state at all. This includes the following functions: `allocate`, `initial_reserve_id`, `is_allocated`, `free` and `write`. Two functions are not protected by this mutex: `fraction_used` and `output`, as they are simple enough as to not cause any problems.

I have confirmed that none of these functions are used within each other, so the possibility of a deadlock occurring is zero.

## Checklist


darktohka:
Since UniqueIdAllocator did not have any Python unit tests so far, I've added some.

`panda/src/gutil/test_uniqueIdAllocator.cxx` seems unused, so I've removed that file as well.

rdb (changes requested):
Thanks!

Can you switch this to use LightMutex and LightMutexHolder?  These compile down to nothing in the case of SIMPLE_THREADS, which looks safe because appears to be no chance for thread yields to occur.  This means that people who compile with SIMPLE_THREADS don't pay for the overhead of locking.

darktohka:
Thanks rdb! I'll make that change right now.

I've also asked you a question on Discord, which I'll crosspost here as well:

I intend to change the signature of free from `void free(uint32_t index)` to `bool free(uint32_t index)` so that free could be used as an atomic operation (returning true if freeing the ID succeeded, and false if freeing the ID failed).

This is in order to change code like this:

```
allocator = UniqueIdAllocator()

if allocator.is_allocated(500):
    allocator.free(500)
    print('Freed ID')
```

to

```
allocator = UniqueIdAllocator()

if allocator.free(500):
    print('Freed ID')
```

as the allocated state might be changed by a separate thread in the meantime.

Should I:
1. Create a commit that encapsulates this change, alongside with the changed unit tests, and push it to the PR (so you can merge it to master and cherry-pick the first two commits for release/1.10.x)
2. or create a separate PR once this PR is pushed?

I'm asking because the function signature change ensures that this would be a release/1.11.x specific change, compared to the current commits that can be backported to release/1.10.x with ease.

rdb:
In case of doubt, if the change is separate, a separate PR would be good.

This change cannot be backported to 1.10 because it changes the memory layout of UniqueIdAllocator.

darktohka:
I'll create a separate PR once this one is reviewed for those changes. Thanks rdb!

rdb:
If you absolutely need this fix in a 1.10 release, you could make a separate PR against the 1.10 branch that changed the `uint32_t *_table` argument to instead be a pointer to a struct that stores both a mutex and the table.  That would not change the memory layout and still be safe because the table is only allocated once in the constructor.

rdb (approved):
Approved pending test completion.  Thanks for your diligence!

darktohka:
It's not that important to me, as I always use master. Like you said, it sounds like an ugly hack to me as well, so it's not a change worth pursuing right now.

darktohka:
On the other hand, 18a28292cc2379c8fad630e63a0faedb32aad611 (the UniqueIdAllocator tests) might be a good idea to backport to release/1.10.x, as the behavior should be consistent.

Issue #997, opened 2020-08-25
ClockObject tests fail spontaneously on Windows


darktohka:
## Description
Sometimes, `test_clock_get_long_time` errors when running the test suite on Windows, more often on the GitHub CI.

This is a Windows-specific issue. given that GetTickCount is unreliable and only offers 10 milliseconds of accuracy (see https://github.com/panda3d/panda3d/blob/master/panda/src/express/trueClock.cxx#L74 and https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-gettickcount)

## Steps to Reproduce
Create a pull request, and wait for the tests to finish.

Sometimes, the `windows-2016` build will fail as such:

```
================================== FAILURES ===================================
__________________________ test_clock_get_long_time ___________________________

clockobj = <panda3d.core.ClockObject object at 0x00000172F14AF900>

    def test_clock_get_long_time(clockobj):
        current_time = clockobj.get_long_time()
        time.sleep(0.4)
>       assert clockobj.get_long_time() - current_time >= 0.39
E       assert (0.3899999999999988 - 0.0) >= 0.39
E        +  where 0.3899999999999988 = <built-in method get_long_time of panda3d.core.ClockObject object at 0x00000172F14AF900>()
E        +    where <built-in method get_long_time of panda3d.core.ClockObject object at 0x00000172F14AF900> = <panda3d.core.ClockObject object at 0x00000172F14AF900>.get_long_time

tests\putil\test_clockobject.py:25: AssertionError
```

## Environment
 * Operating system: Windows Server 2016
 * System architecture: AMD64
 * Panda3D version: 1.11.0
 * Installation method: Built from source
 * Python version (if using Python): Python 3.8
 * Compiler (if using C++): MSVC

darktohka:
Solved by #998

PR #998, opened 2020-08-25
tests: Account for clock inaccuracies in all ClockObject tests

darktohka:
## Issue description
This pull request intends to solve issue #997.

Sometimes, `test_clock_get_long_time` errors when running the test suite on Windows, more often on the GitHub CI.

This is a Windows-specific issue. given that GetTickCount is unreliable and only offers 10 milliseconds of accuracy (see https://github.com/panda3d/panda3d/blob/master/panda/src/express/trueClock.cxx#L74 and https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-gettickcount).

Therefore, we must account for clock inaccuracies when conducting ClockObject tests.

## Solution description
This PR solves the problem by introducing clock inaccuracies into the test.

On Windows, the clock inaccuracy is perceived as 16 milliseconds, as per the Microsoft documentation. This is a worst case scenario, as the best case scenario would be a 10 millisecond inaccuracy.

On Unix systems, the actual clock inaccuracy is 10^-6 seconds (1 microsecond), but to prevent any potential timing issues I have set it as 1 millisecond for the time being.

`math.isclose` has been available since Python 3.5, so its usage should not be a problem. It also accounts for floating point inaccuracies, with an epsilon of 1e-09.

## Checklist

### DIFF
--- a/tests/putil/test_clockobject.py
+++ b/tests/putil/test_clockobject.py
@@ -5,13 +5,13 @@ import sys
 
 # We must account for clock inaccuracy
 if sys.platform == 'win32' or sys.platform == 'cygwin':
-    # Assume 17 milliseconds inaccuracy on Windows (worst case)
-    # 16 milliseconds plus 1 millisecond for execution time
-    CLOCK_INACCURACY = 0.017
+    # Assume 19 milliseconds inaccuracy on Windows (worst case)
+    # 16 milliseconds plus 3 milliseconds for execution time
+    CLOCK_INACCURACY = 0.019
 else:
-    # On other platforms, assume 1 milliseconds, allowing for execution time
-    # (1000 times higher than their actual 1 microsecond accuracy)
-    CLOCK_INACCURACY = 0.001
+    # On other platforms, assume 5 milliseconds, allowing for execution time
+    # (5000 times higher than their actual 1 microsecond accuracy)
+    CLOCK_INACCURACY = 0.005
 
 
 def test_clock_get_frame_time(clockobj):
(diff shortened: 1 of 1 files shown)


tests/putil/test_clockobject.py
```diff
@@ -16,13 +29,13 @@ def test_clock_jump_frame_time(clockobj):
...
     time.sleep(0.4)
-    assert clockobj.get_real_time() - current_time >= 0.39
+    assert math.isclose(clockobj.get_real_time() - current_time, 0.4, abs_tol=CLOCK_INACCURACY)
 
 
 def test_clock_get_long_time(clockobj):
     current_time = clockobj.get_long_time()
     time.sleep(0.4)
-    assert clockobj.get_long_time() - current_time >= 0.39
+    assert math.isclose(clockobj.get_long_time() - current_time, 0.4, abs_tol=CLOCK_INACCURACY)
```
rdb, line 38: The reason this was a `>=` is because on a build server under load it's always possible for the CPU scheduler to switch to another, higher-priority process in the middle of executing this code.  It can therefore take quite a lot longer in reality.

Note that pytest also has an approx function.  I don't know whether one has more advantage than the other.

rdb (changes requested):

tests/putil/test_clockobject.py
darktohka: You're right about that. I replaced math.isclose with explicit calculations instead.

rdb (approved):

PR #999, opened 2020-08-27
putil: Turn UniqueIdAllocator's free into an atomic operation

darktohka:
## Issue description
Currently, using the "free" method on an UniqueIdAllocator is relatively dangerous.

If you free an ID that has not been allocated yet, you will trigger an `_table[index] == IndexAllocated`  assert on debug builds. (I presume it causes a crash on release builds.)

A previous commit, fb0807d7696142d403bb6b18e14acabbdf2b9308, attempted to solve this issue by adding an `is_allocated` function, which could be called before freeing an ID.

However, we haven't considered the fact that UniqueIdAllocators might be used by more than one thread at once!

What happens if you check if an ID is allocated, then attempt to free it while another thread frees it in the background?

Even though you've checked for the allocation state already, since some other thread has already freed that ID, you'll run into the same assert, even with the `is_allocated` safeguard.

## Solution description
I propose that we turn `void UniqueIdAllocator::free(uint32_t id)`'s function signature into `bool UniqueIdAllocator::free(uint32_t id)` instead.

When a unique ID free attempt is made, the method will now return `false` if the ID could not be freed (either it's out of bounds or already freed) or `true` if the ID was freed successfully.

This way, `free` turns into an atomic operation that encapsulates the `is_allocated` safeguard without any risk of other threads interfering.

This turns code like:

```python
allocator = UniqueIdAllocator()

if allocator.is_allocated(500):
    allocator.free(500)
    print('Freed ID')
```

into code like this:

```python
allocator = UniqueIdAllocator()

if allocator.free(500):
    print('Freed ID')
```

which looks so much safer and cleaner.

We should still keep the `is_allocated` method. It certainly has other uses besides being a safeguard - it could be used for statistics in an admin panel, for example.

## Checklist

### DIFF
--- a/panda/src/putil/uniqueIdAllocator.cxx
+++ b/panda/src/putil/uniqueIdAllocator.cxx
@@ -195,7 +195,9 @@ is_allocated(uint32_t id) {
  * Free an allocated index (index must be between _min and _max that were
  * passed to the constructor).
  *
- * Returns true if the index has been freed successfully.
+ * Since 1.11.0, returns true if the index has been freed successfully
+ * or false if the index has not been allocated yet, instead of
+ * triggering an assertion.
  */
 bool UniqueIdAllocator::
 free(uint32_t id) {
--- a/panda/src/putil/uniqueIdAllocator.cxx
+++ b/panda/src/putil/uniqueIdAllocator.cxx
@@ -194,20 +194,32 @@ is_allocated(uint32_t id) {
 /**
  * Free an allocated index (index must be between _min and _max that were
  * passed to the constructor).
+ *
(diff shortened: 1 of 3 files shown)


rdb (changes requested):
This looks good to me.  An alternative would have been to create a new operation for this (eg. `try_free()`) and rewrite `free()` to use it but I'm fine with just changing the existing behaviour as well, since the proposed behaviour does make sense.

Since it is a change of behaviour, though, it would be good if the docstring explicitly mentioned that the behaviour has changed.

rdb (approved):

PR #1000, opened 2020-08-29
direct: Clean up broken usage of "types" module with Python 3

darktohka:
## Issue description
During the migration to Python 3, some functionality in old direct scripts has been broken.

Some of these breakages pertain to old scripts that are no longer being used (by anyone, I believe), mostly `DirectMySQLdbConnection` and `Finder`.

For starters, a lot of the types that have been defined in Python 2 through the `types` module are now removed.

Some new types have also been added that, if not ignored, throw off the results of the `GarbageReport` class (types.CellType in particular after Python 3.8).

`types.MethodType`'s constructor now only accepts two arguments, rather than three. The functionality related to the third argument has been completely removed in Python 3.

## Solution description
This PR fixes some of the issues that were brought upon by the migration to Python 3.

All incorrect usage of the `types` module have been corrected to reflect the current state of the language.

My only concern is that `DirectMySQLdbConnection`'s behavior regarding strings has technically changed. When encountering a `string` object, Panda3D will now try to decode the string object into the encoding used by the database. This is also the new default strategy of the `mysqlclient` Python library. It should not break any existing code (since this class has never worked in Python 3), but there is always the possibility of the end user completing their own Python 2->3 migration incorrectly.

I would like to propose the removal of the `DirectMySQLdbConnection` class, however. It depends on `MySQLdb` (provided by `mysqlclient`) and is an old, copy pasted version of the `Connection` class (with VR Studio-specific changes to solve an old mysqlclient bug).

We have three options:
1. Keep `DirectMySQLdbConnection` as-is for now (not a good idea, IMO)
2. Remove `DirectMySQLdbConnection` and have users just use `mysqlclient`
3. Keep `DirectMySQLdbConnection`, but replace it with the following import: `from MySQLdb.connections import Connection as 
DirectMySQLdbConnection`

As somebody who has been using `direct` for a long time, I seriously do not believe anyone is relying on `DirectMySQLdbConnection` at this point. It's been broken for a while and nobody has complained so far, either. I'd recommend we remove the class in the 1.11.0 release, together with `DirectMySQLdb`.

## Checklist

### DIFF
--- a/direct/src/showbase/GarbageReport.py
+++ b/direct/src/showbase/GarbageReport.py
@@ -7,7 +7,6 @@ from direct.showbase.PythonUtil import fastRepr
 from direct.showbase.PythonUtil import AlphabetCounter
 from direct.showbase.Job import Job
 import gc
-import types
 
 GarbageCycleCountAnnounceEvent = 'announceGarbageCycleDesc2num'
 
@@ -212,7 +211,7 @@ class GarbageReport(Job):
                     startIndex = 0
                     # + 1 to include a reference back to the first object
                     endIndex = numObjs + 1
-                    if type(objs[-1]) is types.InstanceType and type(objs[0]) is dict:
+                    if hasattr(objs[-1], '__dict__') and type(objs[0]) is dict:
                         startIndex -= 1
                         endIndex -= 1
 
@@ -221,7 +220,8 @@ class GarbageReport(Job):
                             numToSkip -= 1
                             continue
                         obj = objs[index]
-                        if type(obj) is types.InstanceType:
+                        if hasattr(obj, '__dict__'):
+                            # Instance of a class
                             if not objAlreadyRepresented:
                                 cycleBySyntax += '%s' % obj.__class__.__name__
                             cycleBySyntax += '.'
--- a/direct/src/directutil/DirectMySQLdbConnection.py
+++ b/direct/src/directutil/DirectMySQLdbConnection.py
@@ -68,7 +68,6 @@ class DirectMySQLdbConnection(Connection):
         """
 
         ### DCR: use methods rather than inline-defined functions to prevent memory leaks
-        string_literal = self._get_string_literal(db)
         self.unicode_literal = unicode_literal = self._get_unicode_literal(db)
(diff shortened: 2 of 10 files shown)


rdb:
Congrats on issue #1000!

Just curious, are you actually using `direct.directutil.DirectMySQLdbConnection`?  I've always wondered what it was doing in the Panda source.

rdb:
Oops, I shoudn't reply before reading all of the issue text.  I would be in favour of removing this class, especially if we cannot verify whether the change works or not.  Connecting to MySQL is very out-of-scope for Panda.  Otherwise, I do need some confirmation that the change is effective (via unit test or otherwise).

direct/src/showbase/GarbageReport.py
```diff
@@ -212,7 +211,7 @@ def run(self):
                     startIndex = 0
                     # + 1 to include a reference back to the first object
                     endIndex = numObjs + 1
-                    if type(objs[-1]) is types.InstanceType and type(objs[0]) is dict:
+                    if hasattr(objs[-1], '__dict__') and type(objs[0]) is dict:
```
rdb, line 214: I'm not at all certain about this replacement.  I guess the goal was to identify old-style vs. new-style classes.  So wouldn't the right thing to do just be to remove this `if` block altogether, since all classes are new-style classes?

I'm not sure what having a `__dict__` indicates.

darktohka: This is now removed in 3ac4f3a1988ee304f5c1fa415a0fd701a9826b04 as it is actually a no-op, thank you!

rdb: It's not a no-op, it's meant to just reorder the list of classes in the cycle.  A very common occurrence in a cycle is "instance -> its dict", and if the cycle is reported with the break across this boundary it's just making sure it's still showing the instance first, since that makes the report a smidge easier to read.

I'll be restoring this line when I check in the fix.  Initially I thought it was about identifying old-style classes, but I think all classes should be handled this way.

rdb:
Merged in 24c14296c016e34bef97104a7b7ddf7962a8b0b8

PR #1004, opened 2020-08-31
mayaegg: Fix animations not being created during egg traversal

darktohka:
## Issue description
The maya2egg utilities do not convert any animations from the EGG format back to the Maya format when used.

This even happens when the "-a" flag is provided to enable animation translations.

## Solution description
This is a regression that was introduced by the commit 754344906c657c703de43f2032b254ad2637f72b.

Simply put, a debug print function was removed, but to prevent a compiler warning because of an unused variable, the "GetAnim" method was also removed.

But this is the method that creates the animation entry in the table in the first place!

I've reintroduced this piece of code, and now animations are now converted back to the Maya format correctly.

I've also silenced another compiler warning.

## Checklist


rdb (approved):
Whoops, who knew carelessly commenting out code to squelch warning messages could be a [bad idea](https://www.schneier.com/blog/archives/2008/05/random_number_b.html)?

Thanks a lot for catching this!

Issue #1001, opened 2020-08-30
macOS: Fn+Key no longer work


el-dee:
## Description

On MacBook keyboards, several keys are emulated using the combination Fn+Something. Since https://github.com/panda3d/panda3d/commit/5d6ff0ecc859e7dc59f70f9d5b12670b7eb8e44b (From PR #967) this combination no longer work.

## Steps to Reproduce

This is a trivial code to test : 

```python
#!/usr/bin/env python3

from panda3d.core import *

from direct.showbase.ShowBase import ShowBase

def test_f1():
    print("Got F1")

base = ShowBase()

base.accept('f1', test_f1)
base.run()
```

Press Fn+the key mapped to F1, it should print "Got F1", however with latest 1.10.x it does not print anything.

## Environment
 * Operating system: macOS 10.15
 * System architecture: amd64
 * Panda3D version: 1.10.x https://github.com/panda3d/panda3d/commit/ab6bf5f4f75e9aa16e71a5fed7c03f1ac8ce2762
 * Installation method: built from source
 * Python version (if using Python): 3.7
 * Compiler (if using C++): N/A

el-dee:
Just tested that  reverting https://github.com/panda3d/panda3d/commit/5d6ff0ecc859e7dc59f70f9d5b12670b7eb8e44b on top of latest release/1.10.x restores the functionality, so it's indeed a regression introduced by this commit.

rdb:
I could use some help taking care of this before 1.10.7, as I don't have a keyboard with a Fn key that can plug into a Mac.

el-dee:
After instrumenting the code and lots of trials, I discovered two things : First the Fn+... combination works for many keys, i.e. the navigation keys are correctly modified when used with Fn : Fn+Up gives PgUp. However for the Function keys, UCKeyTranslate() always returns 0x10, whatever the function key is!

Digging around, I unearthed a similar bug in GTK+ (https://xamarin.github.io/bugzilla-archives/35/3505/bug.html), in fact it seems that UCKeyTranslate() does not know how to map F1 to a Unicode value (even though NS has reserved values for them) and so return 0x10.

I tested a fix where if UCKeyTranslate() returns 0x10, we fallback to the previous code flow use [event charactersIgnoringModifiers] to retrieve the actual key code, and it seems to work (even Fn+Shift+F1 is detected as shift-f1). I will clean up and make a PR later.

rdb:
Fixed in #1007.

PR #1007, opened 2020-09-01
cocoadisplay: Use [event charactersIgnoringModifiers] when UCKeyTranslate() can not translate the input key

el-dee:
## Issue description

This PR aim to fix the regression described in #1001, causing functions key to no longer work on MacBook where they are used in conjunction with the Fn key.

## Solution description

If UCKeyTranslate could not map the key into a valid unicode character or reserved symbol (See NSEvent.h), it returns 0x10 as translated character. This happens e.g.e with the combination Fn+F1.. keys (but for weird reasons not for the arrow keys).
In that case, as fallback, we use [event charactersIgnoringModifiers] to retrieve the character without modifiers.

## Checklist

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -1762,6 +1762,19 @@ handle_key_event(NSEvent *event) {
     return;
   }
 
+  // If UCKeyTranslate could not map the key into a valid unicode character or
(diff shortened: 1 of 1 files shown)


el-dee:
You may have to re-run the CI jobs: the base branch was wrongly master and when I updated it the CI was already running and it didn't like the change mid flight...

el-dee:
The CI is successful, but the old CMake jobs are still referenced and failing obviously

rdb (approved):
Looks good, thank you so much for the quick fix!

PR #1009, opened 2020-09-03
Incorporating lz4 to compress texture (WIP)

ChinYing-Li:
## Issue description

This PR is aimed at solving #745, by adding lz4 to compress raw texture in bam files.
## Solution description

Refactored zStream and zStreamBuf. 
This is still a work in progress, and it would be great if there's any feedback.
Right now I plan to add an extra parameter to all functions in `compress_string.h` e.g. `compress_string(const std::string &source,  int compression_level)` becomes `compress_string(const std::string &source, CompressionAlgorithm compression_algo, int compression_level)`. 
However, this requires a base class (say, `OCompressStreamBase`) as an interface for `IDecompressStream(Zlib/Lz4)` and `OCompressStream(Zlib/Lz4)`.
`OCompressStreamBase` and `IDecompressStreamBase` inherit from `std::ostream` and `std::istream`, respectively.
 The reason for creating such an abstract base class is to use `std::shared_ptr<OCompressStreamBase/IDecompressStreamBase>` in `compress_string`.
To do so, we need `std::enabled_shared_from_this`, which become part of the c++11 standard.
Any thought on my approach is greatly appreciated!
## Checklist


Moguri:
My main concern is this looks like an API break. Changing function signatures like this can and will break peoples applications.

ChinYing-Li:
@Moguri Thank you for reminding. I will try to revert to the original function signature.
Does the overall refactor look reasonable?

rdb (commented):
Thanks a lot for working on this!  This looks like a great start.  It would be great to see some benchmark comparisons between lz4, gzip, and uncompressed.

We shouldn't break the API as @Moguri pointed out.  Either make it a separate overload (though disambiguating it requires making CompressionAlgorithm a strongly typed enum).

What is the goal of using `shared_ptr<ostream>` and dynamic allocation in `compress_string`?  It looks like its lifetime doesn't need to extend beyond the scope of the function.

What is the intent behind splitting out create_Istream and create_Ostream, which are only used in one location, and contain only a handful of lines of code?

StreamBufBase needs a less generic name.

Your thoughts are welcome on this, but if we end up with a virtual StreamBufBase that has been specialised for each compression algorithm, maybe we don't need to specialise IDecompressStream/OCompressStream?  Perhaps we could keep it simpler with a single IDecompressStream/OCompressStream that allocates the streambuf with `new`, and allocates one or the other depending on the algorithm passed in?

ChinYing-Li:
@rdb Thank you very much for reviewing. 

1. What is the goal of using `shared_ptr<ostream>` and dynamic allocation in `compress_string`?
The goal is to make `compress_string` a generic function, and users can specify whether they want to use `zlib` or `lz4` to compress the string.  `create_Istream` and `create_Ostream` does seem to be an overkill  since we only have two compression algorithms to choose from. I was thinking that in case panda3d incorporates more algorithm in the future, perhaps having these functions makes the code more readable. 

2. if we end up with a virtual `StreamBufBase` that has been specialised for each compression algorithm, maybe we don't need to specialise` IDecompressStream`/`OCompressStream`?

That was actually what I wanted to achieve. The member `StreamBuf _buf` is needed whenever we initialize a `OCompressStream` object or `IDecompressStream` object, because `_buf` is required in the initialization of the base `std::istream`/`std::ostream`. To implement the generic `OCompressStream` / `IDecompressStream`, change member `StreamBuf _buf` to `std::shared_ptr<StreamBufBase> _buf_ptr`. However, we can't initialize `_buf` by calling `_buf_ptr = std::make_shared<StreamBufZlib>` before `std::istream(*_buf_ptr)` is called.

ChinYing-Li:
I have yet to rename `streamZlib.*` files and remove the original `streamBase.*` files. After installing `lz4` on my linux machine by cloning the `lz4` repo and `make` `make install`, the headers is in `/use/local/include` and `liblz4.so` can be found in `/usr/local/lib`, however panda still could not find the library even if it correctly locates the header `lz4frame.h`. I took reference from how zlib is linked, but maybe I did something wrong in `Package.cmake` or `makepanda.py`? @rdb

ChinYing-Li:
Sorry for coming back late to this; Closing the PR for now as I don't have a good solution to this.

PR #1012, opened 2020-09-09
glgsg: fix identical states never short-circuiting set_state_and_transform

lachbr:
## Issue description
Passing the same state already on the GSG to set_state_and_transform never causes an early-out.

## Solution description
`_inv_state_mask` needed to be initialized to all 1s and accurately reflect what state was actually cared about, for instance if using fixed-function pipeline or not. Also, the ShaderAttrib bit needed to be set on `_state_mask` to signify there is a known shader on the state.

## Checklist


rdb:
Thanks!  I'm quite curious about the extent of the performance benefit that we will achieve from re-enabling this mechanism.

I see this has the potential to affect other back-ends; have these been tested?

lachbr:
Whoops, I forgot about DX9. I've made sure everything works over there as well.

PR #1015, opened 2020-09-12
build_apps: Disable the pip version check when downloading wheels

Moguri:
## Issue description
When downloading wheels as part of `build_apps`, pip will issue warnings about a newer version of pip being available. If the user is not using the latest version of pip, they will likely be
spammed by pip enough as is. There is not much need to remind them about this for each built platform.

## Solution description
Pass `--disable-pip-version-check` to pip when downloading wheels.

## Checklist

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -425,12 +425,13 @@ class build_apps(setuptools.Command):
                     os.remove(os.path.join(whldir, whl))
 
         pip_args = [
+            '--disable-pip-version-check',
             'download',
             '-d', whldir,
             '-r', self.requirements_path,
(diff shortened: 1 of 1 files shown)


rdb:
When was this flag introduced?  And more to the point, do we still support pip versions that don't have this flag?

Moguri:
This flag seems quite old. I cannot find when it was added, but [this  commit](https://github.com/pypa/pip/commit/4a6b32e558e429284969b5a85371a7f9dd040330#diff-3ef137fb9ffdd400f117a565cd94c188) mentions it, and that commit was included as part of 6.1.0.

rdb (approved):

PR #1016, opened 2020-09-12
makepanda: MSVC /O2 for opt4; remove flags implied by /O2

kamgha:
## Issue description
The /O2 flag is more "aggressive" for speed optimization and recommended for release builds over /Ox, see: https://docs.microsoft.com/en-us/cpp/build/reference/ox-full-optimization?view=vs-2015  /Ox can be useful for debugging.

/O2 has already been in use for opt3 and it looks like it has not caused any issues for debugging at that level, so it looks like there is currently no need to switch opt3 to /Ox.

Additionally, flags already implied by /O2 have been removed in this PR, see: https://docs.microsoft.com/en-us/cpp/build/reference/o1-o2-minimize-size-maximize-speed?view=vs-2015

## Solution description
The generated opt4 binaries are slightly smaller with /O2 over /Ox.

## Checklist

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -1122,10 +1122,10 @@ def CompileCxx(obj,src,opts):
 
             if (optlevel==1): cmd += " /MDd /Zi /RTCs /GS"
             if (optlevel==2): cmd += " /MDd /Zi"
-            if (optlevel==3): cmd += " /MD /Zi /GS- /O2 /Ob2 /Oi /Ot /fp:fast"
+            if (optlevel==3): cmd += " /MD /Zi /GS- /O2 /fp:fast"
             if (optlevel==4):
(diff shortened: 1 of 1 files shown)


Moguri (approved):
Looks good to me. I am not an MSVC/Windows user, but the changes make sense to me.

rdb:
Fine with me, but we should make sure we are also making these changes on the CMake side, if applicable.

kamgha:
> 
> 
> we should make sure we are also making these changes on the CMake side, if applicable.

This is applicable, but I'd like to review it in a separate PR or discussion. The flags between _ppremake_ and _makepanda_ have differed slightly, afaik. The flags for _makepanda_ and _CMake_ do differ a bit more.

At a first glance, for example, _CMake_ does not define `/GS- /fp:fast /DFORCE_INLINING /Zp16` for `Release` (which is the replacement for `opt4` minus .pdb),

PR #1018, opened 2020-09-13
makepanda: MSVC fix generating non-SSE2 code for x86

### LINKED ISSUE #1017: makepanda: MSVC 2015+ defaults to /arch:SSE2 for x86 32bit
## Description
According to the MSVC 2015/'17/'19 docs, `/arch` for the x86 (32 bit) target defaults to `/arch:SSE2` when not specified, see: https://docs.microsoft.com/en-us/cpp/build/reference/arch-x86?view=vs-2015

https://github.com/panda3d/panda3d/blob/master/makepanda/makepanda.py#L1114-L1115

would need to be rewritten in order to check whether we are building for x86 (not x64, there can be an ARM target), and whether we are not building for SSE2 to explicitly state `/arch:IA32`.

## Environment
 * Operating system: earlier than Windows 7 (W7 requires SSE2, afaik)
 * System architecture: x86 32bit
 * Panda3D version: git master 66ac3be604b6bf8cd26ebadb2103fbd3235bbd25
 * Installation method: from source
 * Python version (if using Python): 3.7.9
 * Compiler (if using C++): MSVC 2015

kamgha:
## Issue description
Fix for #1017

## Solution description
We are now passing in /arch:IA32 for x86 when no SSE2 is requested.

## Checklist

Fixes #1017

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -1111,8 +1111,12 @@ def CompileCxx(obj,src,opts):
             cmd += "/DWINVER=0x600 "
 
             cmd += "/Fo" + obj + " /nologo /c"
-            if GetTargetArch() != 'x64' and (not PkgSkip("SSE2") or 'SSE2' in opts):
-                cmd += " /arch:SSE2"
+            if GetTargetArch() == 'x86':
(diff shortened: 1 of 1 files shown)


rdb (approved):
Thanks!

PR #1025, opened 2020-09-30
makepanda: Build maya2egg server and client program

darktohka:
## Issue description

Sometimes, Maya models must be converted into egg files in a batch. This means converting more than 1 model at a time.

However, on each maya2egg invocation, the Maya client libraries must be initialized, the Maya license must be checked, and so on...

To solve this, two Panda3D utilities exist: maya2egg_server and maya2egg_client. maya2egg_server opens a server socket on port 4242, and waits for filenames to convert. Meanwhile, maya2egg_client can be used to send filenames for the server to start processing. This way, Maya is only initialized once.

These two utilities can be used by custom pipelines to convert Maya models to EGG files more efficiently.

The problem is that, these utilities are not built by makepanda at all! It would be really important to have those built, so that they can actually be used.

Thanks to @lachbr for bringing the existence of this tool to attention!

## Solution description

This commit creates three new Panda3D tool binaries for each Maya version. For example, for Maya 2016:
- maya2egg2016_client.exe - Used to send work to the maya2egg server
- maya2egg2016_server.exe - A wrapper program that searches for the installed Maya environment, sets up the required environment variables and runs maya2egg2016_server_bin.exe
- maya2egg2016_server_bin.exe - The actual server that converts the Maya models.

## Checklist


rdb:
How much extra space do these binaries take up?

darktohka:
maya2egg2016_client.exe: 251 KB
maya2egg2016_server.exe: 173 KB
maya2egg2016_server_bin.exe: 575 KB

darktohka:
An alternative approach would be to refactor the client and server code in order to make it possible to run maya2egg with a "--server" or a "--client" flag. This way, no new extra binaries would be created.

The downside to this approach would be that the maya2egg client would now rely on Maya being present, even though it shouldn't be necessary. While everybody who would use those tools will most likely have Maya on their computer for sure, it's still something to keep in mind.

darktohka:
An even better idea: Merge the maya2egg server into the maya2egg utility (with a --server flag). This would get rid of two binaries per Maya build. Meanwhile, keep the client as a separate exe, but don't rebuild it for each Maya version (it is version independent.)

This would create one new 173 KB binary, and the overhead would be severely cut down, also avoiding unnecessary linking.

rdb:
Yeah, I was about to suggest merging the server binary into the regular binary with a special flag.

I also like the idea of a single version-independent client.  Good thinking.

darktohka:
Crosspost from commit description:

The maya2egg_server utility is no longer a separate utility. To run the maya2egg_server (now Maya conversion server), use `maya2egg -server` or `egg2maya -server`.

Two new utilities have been added: `maya2egg_client` and `egg2maya_client`. Both of them are Maya version independent, and do not rely on any Maya libraries. However, they're only built with SDKs that already come with the Maya binaries, as the server makes its own assumptions about the current working directory (the client and the server are expected to be on the same machine.)

A new feature has been added to ProgramBase: an `exit_on_complete` flag, which allows you to parse the command line without exiting the program upon a failure or a help command.

In addition to this, a getopt bug has been fixed in ProgramBase: It is now possible to clear the state of the Panda implementation of getopt, making it possible once again to parse command line arguments twice in a row.

darktohka:
PR has been updated for the recent Maya build changes on macOS ARM64.

rdb (approved):
Thanks for your work on this!  I am unable to test it myself but the code looks good.

PR #1028, opened 2020-10-01
makepanda: Add missing YY_NO_UNISTD_H to built Flex sources

darktohka:
## Issue description

About a week ago, somebody reported this error in the Panda3D Discord while building the engine: `/tmp/dcLexer.lxx.cxx can't open unistd.h.`

This is an error that has occurred to me many times before as well. It's caused by the missing `FLEX` option which defines `YY_NO_UNISTD_H`. This error wasn't noticed before, as really really old Flex versions do not make use of the YY_NO_UNISTD_H flag in the first place.

rdb found the solution to this a week ago, but I think we should mainline it.

## Solution description

If we add the `FLEX` option to all built Flex source files in makepanda, `YY_NO_UNISTD_H` will be defined at build time, preventing the `<unistd.h>` include that causes the error `/tmp/dcLexer.lxx.cxx can't open unistd.h.`

## Checklist

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -1434,7 +1434,7 @@ def CompileFlex(wobj,wsrc,opts):
             oscmd(flex +    " -P" + pre + " -o"+wdst+" "+wsrc)
 
     # Finally, compile the generated source file.
-    CompileCxx(wobj,wdst,opts)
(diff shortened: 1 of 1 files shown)


rdb (approved):
Great!

PR #1031, opened 2020-10-07
openal: Fix paused sounds not resuming from where they were paused

lachbr:
## Issue description
Calling `setActive(False)` on a sound is intended to pause the sound, and a future call to `setActive(True)` is intended to resume the paused sound.  Paused sounds were being restarted instead of resumed where they left off.

## Solution description
Before pausing the sound, store the current time as the start time for when the sound is resumed.  This is what the Miles implementation did, and what FMOD _tries_ to do, but the start time is mistakenly overridden to 0 (I'm working on improving FMOD).

## Checklist

### DIFF
--- a/panda/src/audiotraits/openalAudioSound.cxx
+++ b/panda/src/audiotraits/openalAudioSound.cxx
@@ -831,6 +831,8 @@ set_active(bool active) {
         if (_loop_count==0) {
           // ...we're pausing a looping sound.
           _paused=true;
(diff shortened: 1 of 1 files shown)


rdb (approved):
I take it you've tested this with looped sounds—that it still loops around from the beginning?

lachbr:
Yes, looped sounds still loop around from the beginning even if they've been paused and resumed.

Issue #667, opened 2019-06-05
panda3d audio: adding support for OpenAL extension


brightening-eyes:
hi,
as you know, panda3d doesn't have support for configureFilters() in OpenAL implementation of the audio library, while OpenAL soft has an extension called EFX extension that can handle this, which it's documentation is available [here](https://kcat.strangesoft.net/misc-downloads/Effects%20Extension%20Guide.pdf)
if i can start implementing it, that would be cool.
any ideas on how it could be implemented?

Moguri:
To add support for EFX, `OpenALAudioManager::configure_filters()` needs to be implemented in `panda/src/audiotraits/openalAudioManager.cxx`. The implementation in `panda/src/audiotraits/fmodAudioManager.cxx` can be used as an example.

rasmaxim:
Last 3 days I've been implementing OpenAL EFX extension support in Panda3D audio manager. The main idea is to retain compatibility with FMOD, so that the user would get close (but not 100%) sounding results when switching audio backends.

There are the some differences in how FMOD and OpenAL Soft work though. TL;DR: FMOD system is much more flexible, though Panda almost doesn't use this flexibility. This imposes some limitations, but  they can be removed upgrading to the latest OpenAL soft version (at least 1.20). Panda3D uses 1.17 if I am not mistaken.

### General differences

1. FMOD: has channels and channel groups. 
OpenAL: no channels and channel groups.

2. FMOD: has a rather high limit on the number of effects. 
OpenAL: lets to use only 4 effect slots and 2 sends. That means that a sound source can have only 2 simultaneous effects.

The limit can be changed since [this commit](https://github.com/kcat/openal-soft/commit/d3cc867bd42759cb8e294d691354187984f96ff4) (version 1.18)

3. FMOD: has control over dry/wet component of the signal. OpenAL: no control over the separate components of the signal. There's a hack to achieve the same effect: put an open lowpass/highpass filter and send it gain to 0.0 to block the signal path.

4. FMOD: additionaly, effects have a dry/wet mix parameter. 
OpenAL: effects don't have this parameter. Can try to use the previous hack to imitate this.

5. FMOD: filters are a subset effects. 
OpenAL:, filters (lowpass, highpass, bandpass) and effects are separate entities.

[...]

PR #1032, opened 2020-10-08
Update the FMOD audio system, DSP system

lachbr:
This PR updates the FMOD audio implementation to use the modern FMOD Core API, makes general improvements and fixes to the FMOD implementation, and implements a new interface for creating and applying DSP filters.

More specific change list:
- Updated to use FMOD Core API
- CMake/makepanda look for the FMOD Core libraries and link it with libp3fmod_audio
- `set_play_rate()` on a MIDI sequence changes the tempo instead of the frequency, keeping the pitch unchanged
- Pause/resume of a looped sound correctly resumes it from where it was paused
- Added 7.1.4 speaker configuration
- Sample rate of software mixer can be changed via a PRC variable
- Implemented the concurrent sound limit system that OpenAL has
- Fixed sound memory leaks by following what OpenAL does and keep a reference to sounds that are playing on the manager, and release the reference when the sound is finished during the manager's `update()` method.
- Only update the FMOD `System` once per frame
- `set_speaker_mix()` takes a single speaker ID and mix value instead of a mix value for all speakers, so speaker mix values can be changed one at a time instead of all at once.
- Cleaned up the code to better follow Panda's coding style and conventions
- Replace `FilterProperties` with a cleaner and more flexible system, described below

Info about DSP system:
- Each type of DSP filter is a subclass of `DSP`, each with specific parameters that have documentation and can be set with Python properties
- DSP filters can be individually inserted and removed from an audio manager
- Changing the values of a DSP filter's parameters after adding it will automatically reflect in the audio manager
- The same DSP filter can be added to multiple audio managers, and changes to parameters will automatically reflect in all audio managers the DSP has been added to.
- Each DSP filter has default values for all its parameters
- The `SFXReverbDSP` has preset configurations for different types of environments, such as city, hallway, underwater, etc.

At the moment the DSP filters, parameters, and units come directly from FMOD since there isn't another audio implementation that supports DSP filters.  There does appear to be a mapping to SoLoud for some of the filters, however.

Let me know what you think and if there's anything that should be done differently. Thanks!


Moguri:
Nice work! As a heads up there was some work on getting DSPs working with OpenAl using EFX in #667. The comments contain a lot of information on trying to map the FMOD parameters to OpenAl.

It would be really nice to have the DSPs not be FMOD centric (e.g., with their names and parameters). However, I also understand finding a common set of names and parameters that maps well to multiple/any audio backends is likely a fool's errand.

lachbr:
Thanks.  Yeah, I agree that it would be better to have the filters and parameters not be tied to a specific audio library.  However it's not quite clear how to do that without imposing limits on the filters that can be taken advantage of in one audio library vs another.

rdb:
It'd be good to have a core set of filters that will work with every API.

Moguri:
I am not sure if I like this or not, but we could do something like generic DSP classes that specialized, backend-specific DSP classes could inherit from to add additional features/parameters/etc. So, something like ChorusDSP and ChorusFmodDSP.

panda/src/audio/config_audio.cxx
```diff
@@ -99,6 +114,11 @@ ConfigVariableEnum<FmodSpeakerMode> fmod_speaker_mode
  PRC_DESC("Sets the speaker configuration that the FMOD sound system will use. "
           "Options: raw, mono, stereo, quad, surround, 5.1 and 7.1. "));
```
kamgha, line 115: Does this text need "7.1.4" to be added in the list?

kamgha:
The filters could be handled similar to how the different physics libraries are organized: No generic classes, and each filter can expose the parameters that are available for their implementation. #667 mentions shotcomings of a unification, there is a tradeoff for configurability  (e. g. parameters missing for the same effect) and the output may not be the same on a given input for each backend (which may not be the case depending on the used audio backend). Also, from a technical view, the question would arise why one backend would be chosen over the other when the offered features would be the same.

A generalization could be done at a higher level, e. g. in `direct` or something alike and it'd be up to the user whether they need the increased/higher grained flexbility that each implementation offers.

lachbr:
I think you make a good point.

The design of Panda's audio system is that the implementations for different back-ends are plug-and-play.  You can just choose one and not really have to think about it, the API is the same.  This works fine for basic things such as loading sounds, playing them, changing the volume, etc, but as you get into features that are specific to a particular back-end, how to expose these features under a unified API becomes a problem.  For instance, FMOD has a wide variety of DSP filters and a feature-rich sound spatialization system (including geometry occlusion, reverb zones, etc).  SoLoud and OpenAL have a smaller number of DSP filters with different parameters, and more primitive spatialization systems.  If a Panda user wants to use FMOD for their project, they should be aware of that and should be able to take advantage of FMOD specific features.  Same goes for OpenAL and theoretically SoLoud.  If they can't, then there's no reason to have different options, aside from licensing.

A universal API for basic sound loading and playback is fine, but I think there should be back-end specific APIs for back-end specific features.

Moguri:
I like the idea of a high-level, common, shared API for the basic stuff with more direct access for more complicated/advanced things.

paustian:
Is there a downside of having a base audio class that implements the common API and then specific subclasses for each audio library? That seems like a pretty straightforward solution.

kamgha:
Do we have two terms, `DSP` and `Filter`, meaning the same thing now? Looks like FMOD calls them "DSP", while OpenAL and SoLoud call the effects "Filter", like we had before. Would it be sensible, to rename the occurences of `DSP` (back to) `Filter`? I see in the code of this PR some variables still being named filter, whereas the type is DSP, which could confirm this idea.

LuigiFan25:
merged this in another branch of my custom panda3d and it gave an error on mac os when compiling
I used the latest version  api of fmod 
On the latest version of Mac OS
```
In file included from panda/src/audiotraits/fmod_audio_composite1.cxx:3:
panda/src/audiotraits/fmodAudioManager.cxx:989:30: error: use of undeclared
      identifier 'FMOD_DSP_NORMALIZE_THRESHHOLD'; did you mean
      'FMOD_DSP_NORMALIZE_THRESHOLD'?
      dsp->setParameterFloat(FMOD_DSP_NORMALIZE_THRESHHOLD, norm_conf->g...
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                             FMOD_DSP_NORMALIZE_THRESHOLD
thirdparty/darwin-libs-a/fmod/include/fmod_dsp_effects.h:129:5: note: 
      'FMOD_DSP_NORMALIZE_THRESHOLD' declared here
    FMOD_DSP_NORMALIZE_THRESHOLD,
    ^
In file included from panda/src/audiotraits/fmod_audio_composite1.cxx:4:
panda/src/audiotraits/fmodAudioSound.cxx:842:21: warning: unused variable 'time'
      [-Wunused-variable]
        PN_stdfloat time = get_time();
                    ^
In file included from panda/src/audiotraits/fmod_audio_composite1.cxx:2:
In file included from panda/src/audiotraits/config_fmodAudio.cxx:15:
In file included from panda/src/audiotraits/config_fmodAudio.h:18:
In file included from built/include/notifyCategoryProxy.h:19:
In file included from built/include/notifyCategory.h:20:
In file included from built/include/configVariableEnum.h:18:
In file included from built/include/configVariable.h:18:
In file included from built/include/configVariableBase.h:19:
In file included from built/include/configVariableCore.h:19:
In file included from built/include/configPageManager.h:19:
In file included from built/include/dSearchPath.h:19:
In file included from built/include/filename.h:19:
In file included from built/include/typeHandle.h:19:
In file included from /Volumes/Mac-Volume/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/set:427:
/Volumes/Mac-Volume/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__tree:1094:14: error: 
      constructor for 'std::__1::__tree<FMODAudioManager *,
      std::__1::less<FMODAudioManager *>, pallocator_single<FMODAudioManager *>
      >' must explicitly initialize the member '__pair1_' which does not have a
      default constructor
    explicit __tree(const value_compare& __comp)
             ^
/Volumes/Mac-Volume/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/set:565:11: note: 
      in instantiation of member function 'std::__1::__tree<FMODAudioManager *,
      std::__1::less<FMODAudioManager *>, pallocator_single<FMODAudioManager *>
      >::__tree' requested here
        : __tree_(__comp)
          ^
built/include/pset.h:57:97: note: in instantiation of member function
      'std::__1::set<FMODAudioManager *, std::__1::less<FMODAudioManager *>,
      pallocator_single<FMODAudioManager *> >::set' requested here
  ...init, TypeHandle type_handle = pset_type_handle) : base_class(std::move(...
                                                                   ^
panda/src/audiotraits/fmodAudioManager.cxx:1044:26: note: in instantiation of
      member function 'pset<FMODAudioManager *, std::__1::less<FMODAudioManager
      *> >::pset' requested here
    _dsp_managers[dsp] = { mgr };
                         ^
/Volumes/Mac-Volume/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__tree:1036:56: note: 
      member is declared here
    __compressed_pair<__end_node_t, __node_allocator>  __pair1_;
                                                       ^
/Volumes/Mac-Volume/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:2267:7: note: 
      'std::__1::__compressed_pair<std::__1::__tree_end_node<std::__1::__tree_node_base<void
      *> *>, pallocator_single<std::__1::__tree_node<FMODAudioManager *, void *>
      > >' declared here
class __compressed_pair : private __compressed_pair_elem<_T1, 0>,
      ^
1 warning and 2 errors generated.
The following command returned a non-zero value: clang++ -std=gnu++11 -ftemplate-depth-70 -fPIC -c -o built/tmp/fmod_audio_fmod_audio_composite1.o -Ibuilt/tmp -Ibuilt/include -Ithirdparty/darwin-libs-a/eigen/include -Ithirdparty/darwin-libs-a/fmod/include -Ipanda/src/audiotraits -fvisibility=hidden -Wno-deprecated-declarations -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk -mmacosx-version-min=10.9 -stdlib=libc++ -pthread -fno-exceptions -msse2 -fno-strict-aliasing -ffast-math -fno-stack-protector -fno-finite-math-only -fno-unsafe-math-optimizations -O2 -Wall -Wno-unused-function -Wno-reorder -DBUILDING_FMOD_AUDIO panda/src/audiotraits/fmod_audio_composite1.cxx
Storing dependency cache.
[T1] Waiting for tasks...
```

panda/src/audiotraits/fmodAudioManager.cxx
```diff
@@ -732,76 +725,450 @@ audio_3d_set_drop_off_factor(PN_stdfloat factor) {
...
+      LowpassDSP *lp_conf = DCAST(LowpassDSP, dsp_conf);
+      dsp->setParameterFloat(FMOD_DSP_LOWPASS_CUTOFF, lp_conf->get_cutoff());
+      dsp->setParameterFloat(FMOD_DSP_LOWPASS_RESONANCE, lp_conf->get_resonance());
+    }
+    break;
+  case DSP::DT_normalize:
+    {
+      NormalizeDSP *norm_conf = DCAST(NormalizeDSP, dsp_conf);
+      dsp->setParameterFloat(FMOD_DSP_NORMALIZE_FADETIME, norm_conf->get_fade_time());
+      dsp->setParameterFloat(FMOD_DSP_NORMALIZE_THRESHHOLD, norm_conf->get_threshold());
```
LuigiFan25, line 989: This seems to error out 
```
In file included from panda/src/audiotraits/fmod_audio_composite1.cxx:3:
panda/src/audiotraits/fmodAudioManager.cxx:989:30: error: use of undeclared
      identifier 'FMOD_DSP_NORMALIZE_THRESHHOLD'; did you mean
      'FMOD_DSP_NORMALIZE_THRESHOLD'?
      dsp->setParameterFloat(FMOD_DSP_NORMALIZE_THRESHHOLD, norm_conf->g...
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                             FMOD_DSP_NORMALIZE_THRESHOLD
```
Should be revised to FMOD_DSP_NORMALIZE_THRESHOLD?

lachbr: Is this on the latest version of the FMOD Engine?  If so, it must've been renamed to that, because the version of FMOD I have has FMOD_DSP_NORMALIZE_THRESHHOLD.

LuigiFan25: Yeah must be a latest version thing

LuigiFan25: Can you let me know which version of API you used so I can implement this in my custom panda3d version?

rdb:
What's the status on this PR?

It would probably be really useful to have the FMOD update split out from the DSP system changes

rdb:
Still interested in splitting this, or should someone else take over the work?  Would be nice to have the FMOD update in 1.11.

PR #1035, opened 2020-10-13
collide: Expose CollisionPolygon's setup_points and verify_points to Python

darktohka:
## Issue description
I would like to set a CollisionPolygon's list of points. Unfortunately, the sequence wrapper over `_points` does not allow this to happen. Fortunately, CollisionPolygon has a `setup_points` method, but this method's visibility is set to private.

## Solution description
This PR sets `setup_points` visibility to public, and adds a Python wrapper for both `setup_points` and `verify_points`, to allow these functions to be called from Python code.

## Checklist

### DIFF
--- a/panda/src/collide/collisionPolygon_ext.h
+++ b/panda/src/collide/collisionPolygon_ext.h
@@ -1,0 +1,43 @@
+/**
+ * PANDA 3D SOFTWARE
+ * Copyright (c) Carnegie Mellon University.  All rights reserved.
+ *
+ * All use of this software is subject to the terms of the revised BSD
+ * license.  You should have received a copy of this license along
+ * with this source code in a file named "LICENSE."
+ *
+ * @file collisionPolygon_ext.h
+ * @author Derzsi Daniel
+ * @date 2020-10-13
+ */
+
+#ifndef COLLISIONPOLYGON_EXT_H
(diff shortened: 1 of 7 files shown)


rdb (changes requested):
Looks good, except this won't work in double-precision mode as-is because your code uses LPoint3f whereas the function takes LPoint3.

You may need to disambiguate between Dtool_LPoint3f and Dtool_LPoint3d based on STDFLOAT_DOUBLE since there is no Dtool_LPoint3.

It should not be hard to add unit tests for this.

panda/src/collide/collisionPolygon_ext.h
```diff
@@ -0,0 +1,43 @@
...
+ * This class defines the extension methods for CollisionPolygon, which are called
+ * instead of any C++ methods with the same prototype.
+ *
+ * @since 1.11.0
+ */
+template<>
+class Extension<CollisionPolygon> : public ExtensionBase<CollisionPolygon> {
+public:
+    bool verify_points(PyObject *points);
+    void setup_points(PyObject *points);
```
rdb, line 35: Too much indent

darktohka:
The requested changes have been made. Unit tests have been added for both methods, the distinction between Point3d and Point3f is now made, and verify_points is now static.

rdb (approved):
LGTM

PR #1038, opened 2020-10-15
dist: Add platform name to NSIS installer filename and generate it into the dist directory

el-dee:
## Issue description

When generating a NSIS installer, the platform name is missing from the installer filename. If more than one platform is build at the same time, only the installer for the latest one is available.

Also the installer is generated in the build/ directory, not in the dist/ directory, this is because makensis consider OutFile relative to the NSI script file.

## Solution description

Use basename instead of fullname to include the platform, also the path specified for OutFile is concatenated with dist_dir.

## Checklist

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -1369,7 +1369,6 @@ class bdist_apps(setuptools.Command):
             for i in apps
         ]
 
-        fullname = self.distribution.get_fullname()
         shortname = self.distribution.get_name()
 
         # Create the .nsi installer script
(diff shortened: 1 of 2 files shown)


Moguri (approved):
Looks good to me!

rdb (approved):
Looks good, though redirecting to master since (if I'm not mistaken) this is technically a change of behaviour, and appears to be a non-critical usability issue.

el-dee:
Well, if you try and build win32 and win_amd64 distributions at the same time, one of them will be erased by the other, so it could be classified as a bug :) Otherwise, it's just an benign annoyance.

PR #1039, opened 2020-10-19
cocoadisplay: Test all the fullscreen modes to find the valid one

el-dee:
## Issue description

I discovered recently that sometime the switch to fullscreen fails on macOS (tested on 10.15), after testing and debugging, it seems that macOS returns fullscreen modes that look valid but are not accepted!

It is obviously not documented, but I found a similar issue on the SDL tracker with some explanation : macOS keeps secret some flags about the modes and so it's not possible to detect the invalid mode from user code (see https://bugzilla.libsdl.org/show_bug.cgi?id=4822)

## Solution description

I took a similar approach than SDL ad now find_display_modes() returns a list of the potential valid modes. Each mode is tested until a valid mode is found.

I did not update the code for 10.6, I no longer have a machine with that SDK installed and as this issue arrises only with HiDPI support it's not needed on 10.6. The only downside is that some code is duplicated for 10.6, but it won't be needed in 1.11

## Checklist

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -629,10 +629,36 @@ open_window() {
   if (_properties.get_fullscreen()) {
     // Change the display mode.
 #if __MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
-    CGDisplayModeRef mode;
+    CFMutableArrayRef modes;
+
+    modes = find_display_modes(_properties.get_x_size(),
+                               _properties.get_y_size());
+
+    if (CFArrayGetCount(modes) > 0) {
+      bool switched = false;
+      for (CFIndex i = 0; i < CFArrayGetCount(modes); i++) {
+        CGDisplayModeRef mode = (CGDisplayModeRef) CFArrayGetValueAtIndex(modes, i);
+        if (do_switch_fullscreen(mode)) {
+          switched = true;
+          break;
+        }
+      }
+      CFRelease(modes);
+
+      if (!switched) {
+        cocoadisplay_cat.error()
+          << "Failed to change display mode.\n";
+        return false;
+      }
+
+    } else {
+      cocoadisplay_cat.error()
+        << "Could not find a suitable display mode!\n";
+      CFRelease(modes);
+      return false;
+    }
 #else
     CFDictionaryRef mode;
-#endif
 
     mode = find_display_mode(_properties.get_x_size(),
                              _properties.get_y_size());
@@ -647,6 +673,7 @@ open_window() {
         << "Failed to change display mode.\n";
       return false;
     }
+#endif
   }
 
   // Make the context current.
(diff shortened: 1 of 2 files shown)


Moguri:
@treamology Any thoughts on this PR? Following along with what SDL does makes sense to me. However, I do not know enough about macOS to know if the 10.6 stuff mentioned is problematic.

panda/src/cocoadisplay/cocoaGraphicsWindow.mm
```diff
@@ -626,13 +626,40 @@
     }
   }
 
+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
```
Moguri, line 629: Does this change mean we completely ignore fullscreen on versions prior to 10.6?

el-dee: Oh, good catch! That check is at the wrong place and so breaks the 10.6 build :(

el-dee:
> @treamology Any thoughts on this PR? Following along with what SDL does makes sense to me. However, I do not know enough about macOS to know if the 10.6 stuff mentioned is problematic.

On 10.6 there is no HiDPI support, so there can't be more than one full screen mode per resolution, so the 10.6 code has no problem when using the first corresponding mode.

treamology (commented):
This looks good to me. Having to copy fixes from SDL so much does make an SDL display backend more appealing...

panda/src/cocoadisplay/cocoaGraphicsWindow.mm
```diff
@@ -875,11 +945,38 @@
...
+      } else {
+        cocoadisplay_cat.error()
+          << "Failed to change display mode.\n";
+      }
+
+    } else {
+      cocoadisplay_cat.error()
+        << "Could not find a suitable display mode with size " << width
+        << "x" << height << "!\n";
+    }
```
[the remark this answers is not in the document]
el-dee: Oops, corrected !

rdb:
Oops, I missed this when preparing 1.10.8.  How critical is it for this to get into the next 1.10 release?  To what extent are users of our public 10.6/10.9 builds affected by this bug?

el-dee:
I never encountered the problem with the native retina screen from Apple, it did occur with an external monitor (a BenQ 4K HDR10 screen). The consequences are not critical, you just not enter fullscreen mode; there is no crash or weird behaviour.

el-dee:
Bumping this as a new release is mentionned. Would you consider it for inclusion in 1.10.9 (or later) or should it be retargeted for 1.11 ?

rdb (approved):
I was going back-and-forth about whether I'm confident enough for this to go into 1.10.x, but having re-reviewed the code, I think this should be fine.  Thank you very much!

rdb:
Actually, if the check that the current mode is adequate matches, won't the `mode` object be retained twice, since we first copy it and then  `CFArrayAppendValue`?  I will see if I can test this.

I was a little concerned whether `kCFTypeArrayCallBacks` was appropriate, given that we use `CGDisplayModeRelease` instead of `CFRelease` to release a display mode, but it appears that these functions are equivalent (according to the documentation).

Incidentally, the case you were worried about duplicating is for pre-10.6, so we didn't strictly need that, but I suppose at this point it is harmless to keep it.

PR #1045, opened 2020-11-02
add sRGB support to egg-palettize, sluminance to egg

lachbr:
## Issue description
This PR adds sRGB texture support to egg-palettize, and adds the sluminance and sluminance_alpha formats to egg.  Previously, there was no way to tell egg-palettize that a texture is in sRGB space.

## Solution description
A new `srgb` keyword is made available in `textures.txa`.  This specifies that a texture is in the sRGB color space and the format of the texture should be changed to the appropriate sRGB equivalent based on the number of image channels or requested number of channels.  The resulting palette will also end up with an sRGB texture format.

```
hands.rgb : 256 256 mipmap srgb
```

## Checklist

### DIFF
--- a/panda/src/egg/eggTexture.cxx
+++ b/panda/src/egg/eggTexture.cxx
@@ -504,6 +504,7 @@ has_alpha_channel(int num_components) const {
   case F_green:
   case F_blue:
   case F_luminance:
+  case F_sluminance:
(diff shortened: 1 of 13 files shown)


rdb (changes requested):
Looks good in principle, needs rebase

PR #1050, opened 2020-11-10
minimal cmake changes for android

pmp-p:
first steps trying to build Panda3D with official vendor ndk instead of termux toolchain

### DIFF
--- a/cmake/macros/Python.cmake
+++ b/cmake/macros/Python.cmake
@@ -48,8 +48,11 @@ function(add_python_target target)
   string(REPLACE "." "/" slash_namespace "${namespace}")
 
   add_library(${target} ${MODULE_TYPE} ${sources})
-  target_link_libraries(${target} PKG::PYTHON)
-
+  if (ANDROID)
+    target_link_libraries(${target} "python${PYMAJOR}.${PYMINOR}" "ffi" "z"  "bz2" "lzma")
+  else()
+    target_link_libraries(${target} PKG::PYTHON)
+  endif()
   if(BUILD_SHARED_LIBS)
     set(_outdir "${PANDA_OUTPUT_DIR}/${slash_namespace}")
 
--- a/panda/src/gles2gsg/CMakeLists.txt
+++ b/panda/src/gles2gsg/CMakeLists.txt
@@ -13,10 +13,13 @@ set(P3GLES2GSG_SOURCES
 )
 
 composite_sources(p3gles2gsg P3GLES2GSG_SOURCES)
-add_component_library(p3gles2gsg SYMBOL BUILDING_PANDAGLES2
-  ${P3GLE2SGSG_HEADERS} ${P3GLES2GSG_SOURCES})
-target_link_libraries(p3gles2gsg p3glstuff panda
-  PKG::GLES2)
+add_component_library(p3gles2gsg SYMBOL BUILDING_PANDAGLES2 ${P3GLE2SGSG_HEADERS} ${P3GLES2GSG_SOURCES})
+
+if (ANDROID)
+    target_link_libraries(p3gles2gsg p3glstuff panda PKG::GLES2)
+else()
+    target_link_libraries(p3gles2gsg p3glstuff panda PKG::GLES2)
+endif()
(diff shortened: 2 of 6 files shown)


cmake/macros/Python.cmake
```diff
@@ -48,8 +48,11 @@ function(add_python_target target)
   string(REPLACE "." "/" slash_namespace "${namespace}")
 
   add_library(${target} ${MODULE_TYPE} ${sources})
-  target_link_libraries(${target} PKG::PYTHON)
-
+  if (ANDROID)
+    target_link_libraries(${target} "python${PYMAJOR}.${PYMINOR}" "ffi" "z"  "bz2" "lzma")
+  else()
+    target_link_libraries(${target} PKG::PYTHON)
+  endif()
```
rdb, line 55: I don't think this belongs here, is there a way to make find_package do the right thing for Python?

panda/src/gles2gsg/CMakeLists.txt
```diff
@@ -13,10 +13,13 @@ set(P3GLES2GSG_SOURCES
...
-  ${P3GLE2SGSG_HEADERS} ${P3GLES2GSG_SOURCES})
-target_link_libraries(p3gles2gsg p3glstuff panda
-  PKG::GLES2)
+add_component_library(p3gles2gsg SYMBOL BUILDING_PANDAGLES2 ${P3GLE2SGSG_HEADERS} ${P3GLES2GSG_SOURCES})
+
+if (ANDROID)
+    target_link_libraries(p3gles2gsg p3glstuff panda PKG::GLES2)
+else()
+    target_link_libraries(p3gles2gsg p3glstuff panda PKG::GLES2)
+endif()
```
rdb, line 22: These cases are identical?  This change seems to change nothing

panda/src/glstuff/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
+if(ANDROID)
+else()
 if(NOT HAVE_GL)
   return()
 endif()
+endif()
```
rdb, line 6: `if(NOT ANDROID AND NOT HAVE_GL)`… actually do we want to pick this up when building GLES on other platforms as well?  So `if(NOT HAVE_GLES AND NOT HAVE_GLES2 AND NOT HAVE_GL)`?

rdb (changes requested):

cmake/macros/Python.cmake
pmp-p: Well, it's not only python linking, there you could have locale fix too for ndk and old api unless you have a better place for it.  i'm also not sure find_package in cmake ndk toolchain does even know what python is a thing :unsure:

panda/src/gles2gsg/CMakeLists.txt
pmp-p: indeed it's a leftover

panda/src/glstuff/CMakeLists.txt
pmp-p: i've thought if(android) elif(webgl) elif(etc ... ) would be easier to read than a  big too much NOT HAVE_*  if at once

cmake/macros/Python.cmake
pmp-p: maybe it would be simpler to pass whole python link command via a cmake parameter. what do you think ?

rdb: There probably already is a setting for this, such as PYTHON_LINK_LIBRARIES, or something like that

pmp-p: That would be a setting containing  sysconfig.get_config_vars('LIBS')  but i've found nothing like it even in latest cmake https://cmake.org/cmake/help/v3.19/module/FindPython3.html.

rdb (changes requested):
Requested changes still pending.

pmp-p:
python cmake autodetection seems like a total mess. If you don't mind i'll retry later when webgl port is merged. working cmake through android/wasi/emsdk at once should sort things out.

PR #1053, opened 2020-11-19
panda: Remove PyEval_InitThreads calls when compiling against Python 3.9+

darktohka:
## Issue description

When compiling on Python 3.9, the following compiler warning is shown:
```
[T3] Building C++ object built/tmp/cpython-39-x86_64-linux-gnu/p3event_pythonTask.o
panda/src/event/pythonTask.cxx: In constructor ‘PythonTask::PythonTask(PyObject*, const string&)’:
panda/src/event/pythonTask.cxx:77:22: warning: ‘void PyEval_InitThreads()’ is deprecated [-Wdeprecated-declarations]
   77 |   PyEval_InitThreads();
      |                      ^
```

Browsing the Python C API reference yields the following result: https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads

According to the API reference, in Python 3.9+ PyEval_InitThreads is now a no-op and will be removed in Python 3.11.

## Solution description

PyEval_InitThreads calls are no longer built into Panda3D when compiling against Python 3.9 and older.

## Checklist

### DIFF
--- a/panda/src/event/pythonTask.cxx
+++ b/panda/src/event/pythonTask.cxx
@@ -71,11 +71,11 @@ PythonTask(PyObject *func_or_coro, const std::string &name) :
 
   __dict__ = PyDict_New();
 
-#ifndef SIMPLE_THREADS
+#if !defined(SIMPLE_THREADS) && defined(WITH_THREAD) && PY_VERSION_HEX < 0x03090000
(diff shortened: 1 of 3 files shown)


rdb (approved):
Thanks, LGTM!

PR #1059, opened 2020-11-24
build_apps: Add hook for packages to modify options

Moguri:
## Issue description
Packages such as [panda3d-simplepbr](https://github.com/Moguri/panda3d-simplepbr/) require users to modify `build_apps` options to handle data files. It would be nice if such packages could automatically handle this for users.

## Solution description
This PR adds a 'panda3d.build_apps.modify_options' pkg_resources entry point. This allows packages to automatically handle setting/modifying any build_apps options.

## Checklist

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -400,14 +400,6 @@ class build_apps(setuptools.Command):
         if self.default_prc_dir is None:
             self.default_prc_dir = '<auto>etc' if not self.embed_prc_data else ''
 
-        num_gui_apps = len(self.gui_apps)
-        num_console_apps = len(self.console_apps)
-
-        if not self.macos_main_app:
-            if num_gui_apps > 1:
-                assert False, 'macos_main_app must be defined if more than one gui_app is defined'
-            elif num_gui_apps == 1:
-                self.macos_main_app = list(self.gui_apps.keys())[0]
 
         use_pipenv = (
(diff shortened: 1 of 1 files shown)


rdb (commented):
Could you show an example how a package will use this to modify options?

Moguri:
I used this snippet in panda3d-simplepbr to test this with:

```python
def modify_options(build_apps_cmd):
    if 'simplepbr' not in build_apps_cmd.package_data_dirs:
        build_apps_cmd.package_data_dirs['simplepbr'] = [
            ('simplepbr/*.vert', '', {}),
            ('simplepbr/*.frag', '', {}),
        ]
```
And the following in `setup.py`:

```python
from setuptools import setup

setup(
    # ...
    entry_points={
        'panda3d.build_apps.modify_options': [
            'simplepbr=simplepbr.buildapps:modify_options'
        ]
    }
)
```

Moguri:
@rdb is this something we'd like to continue to pursue? I no longer need this for simplepbr since it just bundles shaders into Python files. In other words, if we're on the fence on this, I'll probably just abandon this PR.

Moguri:
This can actually be done with a `setuptools.finalize_distribution_options` entrypoint. Here is an example: https://github.com/Moguri/pman/blob/20b4ae113e57c6fb4bfb62b0d95c08997660f5aa/pman/setuptools.py

Issue #1049, opened 2020-11-09
System cursors switching


PapyChacal:
## Description

  I'd like to see Panda3D offer an interface to switch between system cursors, like the usual arrow cursor, the text cursor, or the finger cursor, for example.

## Use Case
  I'm working on a dear ImGui backend for Panda3D, and it would allow consistent cursor behavior in the dear ImGui interface while using only the Panda3D's interface. It would help for any Panda3D-based tool's GUI, I guess.
  Dear ImGui is great for such GUIs, that's why I'm working on a backend for Panda3D, hoping for integration on the Panda3D's side later, which would be easy once the backend is ready. If this is not considered useful, integrating dear ImGui in any Panda3D project will still be easier.

rdb:
I would suggest that we amend WindowProperties to take a new "cursor type" enumeration property, which would contain a "default" value, a "hidden" value (replacing the hidden cursor property), and a list of built-in cursors such as the one [here](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor).

PapyChacal:
I'm working on this, I can switch between two cursor types on X11 for the moment. I have questions:
- What about the "cursor filename" property: should it be used with a given cursor type, as hidden, or should it override the cursor type ?
- I'm new to the overall Panda3D's architecture: I found usages of cursor_filename in WinGraphicsWindows and X11GraphicsWindows only: are those the only places where I should implement this new property, or am I missing some?

rdb:
Thanks for working on this!

Good question.  We could make a cursor type we call "custom" which, when set, uses the specified cursor_filename; or we could just say that the cursor_filename always takes precedence and Panda falls back to the specified cursor type if the cursor file can't be loaded.  I am at the moment leaning towards the latter option.

The cursor_filename is defined in windowProperties.h and implemented in the various *GraphicsWindow.h classes.  Ideally we'd also add it to cocoaGraphicsWindow.h for macOS but perhaps someone else can add it if you are unable to.

PapyChacal:
My pleasure :slightly_smiling_face: 

I agree with the latter idea about cursor_filename and will work on this.

I don't have a macOS environment to properly check what I implement in there. I'll write something first and let someone test it!

PapyChacal:
While working on this, I've seen that making these choices lead to a complication or inconsistency.
[...]

PR #1061, opened 2020-11-26
Cursor_type property to use standard cursors

PapyChacal:
## Issue description
  See #1049 

## Solution description

  I added a cursor_type property. It takes value in a CursorType enum, and have four values representing cursor type at the moment, as well as CT_hidden, which replaces the cursor_hidden property, and CT_custom, which tells Panda to use cursor_filename instead.

  I tried to replace every use of the cursor_hidden property accordingly. It seems to work alright on X11 and Win32, but I have doubts about the cocoa implementation as I don't have access to a macOS environment. This should break some existing use of the Panda3D API, as all the cursor_hidden-related functions are removed.
## Checklist
## Notes
  The way cursor_type and cursor_filename (and eventually cursor_hidden if it should instead be kept) still is to be discussed.

  Before integrating this, the offered cursors should be discussed too. I used the following sources to see which cursors were exposed by which platform:
- X11 : [https://tronche.com/gui/x/xlib/appendix/b/](url)
- Win32 : [https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursora](url)
- Cocoa : [https://developer.apple.com/documentation/appkit/nscursor](url)

As one can see, those platforms offer different cursors. So for this draft, I simply used the set of cursors which clearly were the same on every platform (this reduces to default arrow, open hand, text beam, and crosshair...)

  I could include some more with fallbacks: for example, Win32 and Cocoa both offer bidirectional horizontal and vertical resizing cursors, but X11 only offers unidirectional versions. I could make panda offer a unidirectional cursor_type and fall back to the bidirectional one in this case)
  I could offer a union of those cursor types too, and return a flag or print a Warning or whatever if a platform doesn't support the asked type?
  Finally, Panda could offer its own cursor theme, to get a consistent look and behavior on every platform.
  Any other idea is welcome!


rdb:
I see this is still marked Draft, is this ready to receive review and feedback?

PR #1065, opened 2020-12-08
VFSimporter: solve import can fail on _bootlocale module

pmp-p:
When on systems with missing or incomplete locale support ( eg android , emscripten ) and not using PYTHONUTF8=1 , the C module _bootlocale can fail to be imported lazily after VFS importer is already set up.
https://github.com/python/cpython/blob/3.9/Lib/locale.py#L658

### DIFF
--- a/direct/src/showbase/VFSImporter.py
+++ b/direct/src/showbase/VFSImporter.py
@@ -15,6 +15,10 @@ import marshal
 import imp
 import types
 
+# make sure modules are loaded for default encodings before hooking import system
+# file.py use TextIOWrapper that relies on the default preferred encoding
+__import__('locale').getpreferredencoding()
+
 #: The sharedPackages dictionary lists all of the "shared packages",
 #: special Python packages that automatically span multiple directories
 #: via magic in the VFSImporter.  You can make a package "shared"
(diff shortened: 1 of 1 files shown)


rdb:
I need to better understand the problem this is solving.

pmp-p:
it's really a corner case or 3.9 locale fallback, and it has been fixed in 3.10. i did not check < 3.9

rdb:
Could you explain the problem, though?

pmp-p:
well if you do an import after vfsimporter is installed, the Textiowrapper from direct.std.file will ask the locale module the default encoding ( which should be C.UTF-8 ) but to give that reply the locale module import _bootlocale.py which will use the VFSimporter which will ....... and finally you end up with "ascii" locale because of circular import and it will fail to load encoding too.
i know that is  **not** very clear, but it will happen in the browser with a single  direct.showbase.VFSImporter.register()

rdb:
OK, I understand.  A circular dependency.

However, getpreferredencoding shouldn't be called in the first place, since source files are never using the preferred encoding in Python 3, but rather using UTF-8.

I just checked in a change that reads the source file as binary and uses the tokenize module to detect the encoding (which properly respects `# coding=` comments and defaults to UTF-8 otherwise).  That should implicitly fix this issue as well.

Issue #669, opened 2019-06-06
Suggestion: More verbose deprecation prints


mfwass:
When tracking down deprecations in a large project, it can be difficult to be able to locate the specific line and file in which a deprecation print originates.

For example, the `NodePath.remove()` deprecation print shown below will spam in the program output. It would be extremely helpful to be able to see which file and line in particular is raising this warning.

```
Warning: NodePath.remove() is deprecated.  Use remove_node() instead.
```

A potential improved output could be the following:

```
Warning (fakepath.YourNodePathFile:L32): NodePath.remove() is deprecated.  Use remove_node() instead. 
```

This could be accomplished using `inspect`.

```python
from inspect import currentframe, getframeinfo
frameinfo = getframeinfo(currentframe())
print(frameinfo.filename, frameinfo.lineno)
```

Source: https://stackoverflow.com/questions/3056048/filename-and-line-number-of-python-script

rdb:
I'd be OK with some sort of a traceback protected under an `if __debug__:` block.

Another option is to use the `warnings` module which allows you to run Python with `-W error` to turn every warning into a fatal error (or create a Panda config variable with similar effect).

paustian:
This one looks like some simple line switching and would be a good first effort for me. Just to make sure I am understanding, all the Warnings for deprecations in the code would need to just have the file and line no. where they occur added. This should be pretty easy to do. 

For example in NodePath_extenstions.py on line 17 it says

print("Warning: NodePath.id() is deprecated.  Use hash(NodePath) or NodePath.get_key() instead.")

Would be changed to:

print("NodePath_extentions.py:L17 - Warning: NodePath.id() is deprecated.  Use hash(NodePath) or NodePath.get_key() instead.")

I think I would prefer to make these changes semi-manually since there are a ton of them.
Let me know if this is what's desired and I can get at it.

rdb:
I think we should look at leveraging the Python warnings module, or something else that can look at the call stack and figure out which file it's being called from.  It would be a maintenance burden if we had to update the line number every time we moved around some code.

paustian:
True. I will look into the python warning module.

paustian:
So it looks like every file with a warning would need a bit different code. The old code is:

`print("Warning: NodePath.id() is deprecated. Use hash(NodePath) or NodePath.get_key() instead.")`

You would need at the top of the file:

`import warnings`

Then later in the file:
[...]

PR #1067, opened 2020-12-08
Python warning

paustian:
## Issue description
Fixes Suggestion: [More verbose deprecation prints #669](https://github.com/panda3d/panda3d/issues/669) by replacing print warnings with the warnings module.

## Solution description
The warnings module is used to send notices that function calls are depreciated. The warning reports back the place in the code where the function call originated, instead of the line of the function call itself, as this will be more useful to the developer who receives the warning.


direct/src/dist/FreezeTool.py
```diff
@@ -1957,7 +1958,7 @@ def generateRuntimeFromStub(self, target, stub_file, use_console, fields={},
 
         if append_offset:
             # This is for legacy deploy-stub.
-            print("WARNING: Could not find blob header. Is deploy-stub outdated?")
+            warnings.warn("Could not find blob header. Is deploy-stub outdated?", DeprecationWarning)
```
rdb, line 1961: I don't think this is a deprecation, so this should not be flagged as a deprecation warning.

direct/src/dist/commands.py
```diff
@@ -26,14 +26,16 @@
...
-    print("=================================================================")
-    print("WARNING: You are using Python 2, which has reached the end of its")
-    print("WARNING: life as of January 1, 2020.  Please upgrade to Python 3.")
-    print("=================================================================")
+    # Do we want to leave this as this type of warning or should we switch to warnings class?
+    warnings.warn("=================================================================\n \
+                 WARNING: You are using Python 2, which has reached the end of its\n \
+                 WARNING: life as of January 1, 2020.  Please upgrade to Python 3.\n \
+                 =================================================================",
+                 DeprecationWarning)
```
rdb, line 38: I think this should be a FutureWarning, since it needs to be displayed by default to end-users, especially considering that DeprecationWarning doesn't display by default in Python 2.

That said, I'm not sure why we still have this warning because we no longer support Python 2 to begin with; there should probably be an error message here.  That could also happen in a separate change though.

rdb (changes requested):
Thank you, great work!  I added some inline comments.

As of Python 3.7, DeprecationWarnings are displayed [by default](https://www.python.org/dev/peps/pep-0565/), so I am happy with this—it does mean that Python 3.6 users don't get these warnings unless they explicitly enable them, but that's not a huge problem, I think.

I think there are still a bunch of cases missing.  For example, the use of the legacy global `run()` method, which is defined in ShowBaseGlobal.py.  It currently uses the Notifier system to display a warning, but that mechanism doesn't give the location where it's used, so the DeprecationWarning is better for that case.

There are a couple of deprecation warnings that are excellent candidates for this mechanism in direct/src/showbase/Loader.py as well.  Same as loadSfx and loadMusic in ShowBase.py.

I'm also seeing some deprecated methods in direct/src/extensions_native/NodePath_extensions.py that aren't included in this PR.

I don't think you intended to add `get-pip.py` to this PR?

paustian:
> I think there are still a bunch of cases missing. For example, the use of the legacy global `run()` method, which is defined in ShowBaseGlobal.py. It currently uses the Notifier system to display a warning, but that mechanism doesn't give the location where it's used, so the DeprecationWarning is better for that case.

I did a quick search with the string notify.warning and found 170 results. In all cases, I think the warning system might be more useful than notify.warning. 

On the other hand, searching for deprecated brought up 370 instances that should be changed. Things are brought Loader.py  like:

```
def cancelRequest(self, cb):
        """Cancels an aysynchronous loading or flatten request issued
        earlier.  The callback associated with the request will not be
        called after cancelRequest() has been performed.

        This is now deprecated: call cb.cancel() instead. """

        cb.cancel()
```


 Thoughts?

direct/src/dist/FreezeTool.py
paustian: I made the change to a UserWarning

direct/src/dist/commands.py
paustian: This was updated to a FutureWarning.

rdb:
Yes, I think many of the things that are currently only deprecated in the docstring should well be issuing a real deprecation warning (or a PendingDeprecationWarning for cases where we don't yet want to nag users about it; should be reviewed on a case by case basis).

> I did a quick search with the string notify.warning and found 170 results. In all cases, I think the warning system might be more useful than notify.warning.

I wonder about that.  I don't necessarily think all warnings are directly due to a deprecation or even due to particular line in the stack frame that the developer should change; but if they were, then we ought to patch `notify.warning` to do something similar to the warnings module (perhaps just defer to warnings with stacklevel=3).

direct/src/dist/commands.py
rdb: If I'm not mistaken, the indents are included in the string here?  You could avoid this by putting quotation marks around each line.

paustian:
> Yes, I think many of the things that are currently only deprecated in the docstring should well be issuing a real deprecation warning (or a PendingDeprecationWarning for cases where we don't yet want to nag users about it; should be reviewed on a case by case basis).

Ok. I will do a search for deprecated and then start converting over to the warnings module. I will make my best judgment on UserWarning, DeprecationWarning, or PendingDeprecationWarning

paustian:
> 
> > I did a quick search with the string notify.warning and found 170 results. In all cases, I think the warning system might be more useful than notify.warning.
> 
> I wonder about that. I don't necessarily think all warnings are directly due to a deprecation or even due to particular line in the stack frame that the developer should change; but if they were, then we ought to patch `notify.warning` to do something similar to the warnings module (perhaps just defer to warnings with stacklevel=3).

I am curious as to why have a separate notify.warning vs. the warnings module? What it the purpose of the notify module? Let me know how you want me to handle these cases.

rdb:
The Notifier system is a full-fledged logging system: it can handle messages of various severities (info, error, debug), supports different categories, write them to a log, add timestamps, etc.  Very useful for runtime logging.

The Python warnings module doesn't have most of those features.  But it does have the ability to show on which line a warning occurred.  I think it's great for showing deprecation warnings in particular, which are primarily of interest during development.

paustian:
> 
> > I did a quick search with the string notify.warning and found 170 results. In all cases, I think the warning system might be more useful than notify.warning.
> 
> I wonder about that. I don't necessarily think all warnings are directly due to a deprecation or even due to particular line in the stack frame that the developer should change; but if they were, then we ought to patch `notify.warning` to do something similar to the warnings module (perhaps just defer to warnings with stacklevel=3).

I am curious as to why have a separate notify.warning vs. the warnings module? What it the purpose of the notify module? Let me know how you want me to handle these cases.


> The Notifier system is a full-fledged logging system: it can handle messages of various severities (info, error, debug), supports different categories, write them to a log, add timestamps, etc. Very useful for runtime logging.
> 
> The Python warnings module doesn't have most of those features. But it does have the ability to show on which line a warning occurred. I think it's great for showing deprecation warnings in particular, which are primarily of interest during development.

In looking at def warning() in Notifier.py it would be pretty simple to add a warning here, most of these are userWarnings and it could list the line number. There is a switch for showTime, I could add a switch to Notifier that says show line number. If the switch is on, then send the warning using the warnings module. I am a bit concerned that these may not log to the same place as other notifications in the Nontifier class. Is there a way to make sure the warnings module write to the same file as the the function call self.__log()? Here is the code I am thinking of modifying...
```
 def warning(self, warningString):
        """
        Issue the warning message if warn flag is on
        """
        if self.__warning:
            message = str(warningString)
            if Notifier.showTime.getValue():
                string = (self.getTime() + self.__name + '(warning): ' + message)
            else:
                string = (":" + self.__name + '(warning): ' + message)
            self.__log(string)
#The warnings.warn() would go here
            self.__print(string)
```

rdb:
I'll give it some thought.  Arguably, if showing the line number in the Notifier is useful for warnings, then it's also useful for other messages (eg. error, maybe even info).  Another consideration is that notifier warnings may occur deeply nested inside Panda code, which might suggest the need for Notify.warning to accept an explicit stacklevel parameter.

rdb:
Some additional thoughts.

* The warnings module displays each warning only once when it occurs on a particular line.  It is very specifically meant to indicate that the developer did something wrong on a particular line of their code.  This is not the same as a warning message that would appear in a log file, such as that the game couldn't connect to a server, or something like that.
* The deprecation warnings that we intend to be served by `warnings` module are only useful while someone is developing.  They should be stripped when deploying a build.  It's not really useful to have these in a log file part of a distributed game (this might even reflect poorly on the developer if the user sees use of deprecated APIs).  The Notifier warnings, on the other hand, should not be stripped, because they might contain key information about why a game displays a certain behaviour.

With that in mind I don't think that we should be trying to merge the Notifier and the deprecation warning system.  The desired and implemented behaviour is different enough that "dev warnings" and "general warnings" ought to be considered separate.

I'm a little unsure about this, but we could even write warnings like so:
```
assert not warnings.warn("blah blah", DeprecationWarning)
```
Or like so:
```
if __debug__:
    assert not warnings.warn("blah blah", DeprecationWarning)
```
Either one is automatically recognized by the Python optimizer and removed from the bytecode when running it in optimized mode, such as would be the case in a deployed application.  (The `not` keyword after `assert` would be required for making sure that it doesn't actually trigger an assertion, because the return value is always falsey (`None`)).

Thoughts?

paustian:
Thanks, I have a better understanding of the notify and warning systems now. I agree we should not merge them, since as you said, they serve different purposes. I think I prefer the second method (if __debug__) since it makes it clear these are warnings meant for the developer and the code makes more intuitive sense. The assert not just seems counter intuitive. I will proceed with adding the if __debug__ to the code I changed.

rdb:
Sounds good.

We could also just have the deployment system replace the warnings module with a stub that does nothing.  This may make the code look cleaner without having `if __debug__:` all over the place, although with this method, the warning code wouldn't be entirely optimized out of the Python bytecode, so it would still take up a *bit* of space.

Up to you.

paustian:
I have no problem with if __debug__ in the code. Up to you though. Let me know which way you want to go and I can make the changes.

rdb:
If you have no problem adding them, it would not hurt to have them. :)

rdb (changes requested):
Let's not have both `if __debug__` and `assert not`; I thought the idea was to have just `if __debug__` so we wouldn't need `assert not`.

Please also update this branch from latest master (to pick up the latest deprecations and removals—for example, the commands.py warning no longer exists) and add in warnings for the remaining deprecations (although if you have no time to add more, I can accept it like this and a later PR can add additional ones)

Thank you for working on this!

paustian:
> Let's not have both `if __debug__` and `assert not`; I thought the idea was to have just `if __debug__` so we wouldn't need `assert not`.
> 
Ah, got it. I was going by your example above. I fixed them now.

> Please also update this branch from latest master (to pick up the latest deprecations and removals—for example, the commands.py warning no longer exists) 

I think I did this right. I am still pretty new to using git on a big project like panda3d. I will 

>and add in warnings for the remaining deprecations (although if you have no time to add more, I can accept it like this and a later PR can add additional ones)

I will continue to work on this. I have a better handle on what I am doing (I hope). I enjoy coding. It's like crossword puzzles for me, but it actually does something useful, so win-win.

direct/src/extensions_native/NodePath_extensions.py
```diff
@@ -168,8 +185,9 @@ def pPrintString(self, other = None):
         """
         Deprecated.  pretty print
         """
-        print("NodePath.pPrintString() is deprecated.")
+        print("")
```
rdb, line 188: There doesn't need to be an empty print statement here

direct/src/extensions_native/NodePath_extensions.py
```diff
@@ -200,7 +218,8 @@ def pPrintString(self, other = None):
 
 def printPos(self, other = None, sd = 2):
         """ Deprecated.  Pretty print a node path's pos """
-        print("NodePath.printPos() is deprecated.")
+        if __debug__:
+                warnings.warn("NodePath.printPos() is deprecated.", DeprecationWarning, stacklevel=2)
```
rdb, line 222: This line and other warnings lines in this file are indented too much (I don't know why the functions themselves are also indented too much, but let's use 4 space indents for these)

direct/src/extensions_native/NodePath_extensions.py
```diff
@@ -460,7 +489,9 @@ def showCS(self, mask = None):
         CameraBitmask) that indicates which particular collision
         solids should be made visible; otherwise, all of them will be.
         """
-        print("NodePath.showCS() is deprecated.  Use findAllMatches('**/+CollisionNode').show() instead.")
+        if __debug__:
+                warnings.warn("NodePath.showCS() is deprecated.  Use findAllMatches('**/+CollisionNode').show() instead.", DeprecationWarning, stacklevel=2)
+        print()
```
rdb, line 494: Unnecessary empty print()

direct/src/gui/OnscreenText.py
```diff
@@ -300,10 +301,12 @@ def setTextX(self, x):
         self.setTextPos(x, self.__pos[1])
 
     def setX(self, x):
-        """
+         """
```
rdb, line 304: I think this is the erroneous space causing the checks error

direct/src/showbase/ShowBase.py
```diff
@@ -71,6 +71,7 @@
     from direct.showbase import GarbageReport
     from direct.directutil import DeltaProfiler
     from . import OnScreenDebug
+import warnings
```
rdb, line 74: If you indent this line by four spaces it falls under the `if __debug__:` as well :)

I don't generally suggest putting the `import warnings` under `if __debug__` but in this case it's serendipitous that there is already a block of imports under such a check.

direct/src/showbase/AppRunnerGlobal.py
```diff
@@ -12,7 +12,7 @@
 """
 
 if __debug__:
-    print('AppRunner has been removed and AppRunnerGlobal has been deprecated')
+    warnings.warn("AppRunner has been removed and AppRunnerGlobal has been deprecated.", DeprecationWarning, stacklevel=2)
```
rdb, line 15: Put an `import warnings` right above this line, please

rdb (changes requested):
Please note on the "Checks" tab that the build is currently failing due to a bad indent in OnscreenText.py:
```
Coverage/direct/showbase/OnScreenDebug.py:7: in <module>
    from direct.gui import OnscreenText
E     File "/Users/runner/work/panda3d/panda3d/build/Coverage/direct/gui/OnscreenText.py", line 308
E       if __debug__:
E                   ^
E   IndentationError: unindent does not match any outer indentation level
```

direct/src/dist/commands.py
```diff
@@ -28,6 +28,7 @@
 from . import installers
 from .icon import Icon
 import panda3d.core as p3d
+import warnings
```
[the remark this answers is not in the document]
paustian: Fixed.

direct/src/dist/FreezeTool.py
```diff
@@ -1961,7 +1962,8 @@ def generateRuntimeFromStub(self, target, stub_file, use_console, fields={},
 
         if append_offset:
             # This is for legacy deploy-stub.
-            print("WARNING: Could not find blob header. Is deploy-stub outdated?")
+            if __debug__:
+                warnings.warn("Could not find blob header. Is deploy-stub outdated?")
```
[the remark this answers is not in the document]
paustian: So remove the change? Let me know.

direct/src/extensions_native/NodePath_extensions.py
paustian: Removed

direct/src/extensions_native/NodePath_extensions.py
paustian: Hmm. My editor is set to 4 spaces per tab. Looking at the file, it appears there were 8 spaces for each one. I will fix them all.

paustian: Fixed it. It was very strange. Not sure when this happened.

direct/src/extensions_native/NodePath_extensions.py
paustian: Done

direct/src/gui/OnscreenText.py
paustian: fixed

direct/src/dist/FreezeTool.py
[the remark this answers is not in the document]
rdb: Yes, please undo.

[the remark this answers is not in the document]
paustian: I took out the __debug__ check. Seems to me it fits as a UserWarning.

direct/src/extensions_native/NodePath_extensions.py
paustian: Yeah, weird. Fixed.

direct/src/extensions_native/NodePath_extensions.py
paustian: Deleted

.gitignore
```diff
@@ -70,3 +70,8 @@ __pycache__/
 /.settings/
 /.cproject
 /.project
+.Rproj.user
+
+.Rhistory
+
+panda3d.Rproj
```
rdb, line 77: Did you commit these changes intentionally?  Could you explain them if so?

direct/src/extensions_native/NodePath_extensions.py
```diff
@@ -10,309 +11,330 @@
...
-               formatString % hpr[0] + ', ' +
-               formatString % hpr[1] + ', ' +
-               formatString % hpr[2] + ', ' +
-               formatString % scale[0] + ', ' +
-               formatString % scale[1] + ', ' +
-               formatString % scale[2] +
-               ')\n'))
+    """ Deprecated.  Pretty print a node path's pos, hpr, and scale """
+    if __debug__:
+            warnings.warn("NodePath.printPosHprScale() is deprecated.", DeprecationWarning, stacklevel=2)
```
rdb, line 306: This line is indented too much, and there are several more like that in this file.

direct/src/showbase/DConfig.py
```diff
@@ -4,9 +4,11 @@
 
 from panda3d.core import (ConfigFlags, ConfigVariableBool, ConfigVariableInt,
                           ConfigVariableDouble, ConfigVariableString)
-
+import warnings
 
 def GetBool(sym, default=False):
+    if __debug__:
+        warnings.warn("This is deprecated. use ConfigVriableBool instead", DeprecationWarning, stacklevel=2)
```
rdb, line 11: This is a good addition, but:
* Typo here, should be "ConfigVariableBool" instead of "ConfigVriableBool"
* There other methods in this file are also deprecated, not just GetBool

rdb (changes requested):
Great work, thanks!

See review comments for some remaining nitpicks.  If you can address those and do a merge from master to resolve any conflicts that occurred in the meantime, this should be just about good to merge.

.gitignore
paustian: That was some information that snuck in when I was playing around with using R as my interface for coding and interacting with git. I have since learned better ways to do this using VSCode and GitKraken

direct/src/extensions_native/NodePath_extensions.py
paustian: Indents removed. This should be the last of that.

direct/src/showbase/DConfig.py
paustian: Updated as you suggested.

PR #1069, opened 2020-12-11
cmake tweaks for emsdk

pmp-p:
apart from an annoying recurring `warning: overriding currently unsupported use of floating point exceptions on this target [-Wunsupported-floating-point-opt]` build went very well and was a bit faster than with makepanda.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,16 @@ unset(_s)
 
 enable_testing()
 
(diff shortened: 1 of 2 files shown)


rdb:
Thanks, merged (and merged in latest master).  I restructured some of the CMake changes a bit, and I think the SSE2 change is no longer necessary with the latest master changes.

PR #1078, opened 2020-12-19
shaderpipeline: add support for specialization constants

lachbr:
## Issue description
Adds support for specialization constants to the shaderpipeline branch.  Can be used in place of preprocessor definitions to set values that are constant for the entire shader, without requiring a complete recompilation of the shader from source.

## Solution description
Each `ShaderModule` contains a list of `SpecializationConstants`, which individually store the name, type, and constant ID.  `ShaderModuleSpirV` fills in that list when parsing the SPIR-V instructions.  In `Shader`, external values given to the constants are stored per-module in two lists with matching indices: the constant IDs and the values.  To set a constant from application code, the usage is `Shader.set_constant(name, value)`.  The matchup is done by name, so multiple shader modules containing a spec constant with the same name are given the same value.  This also requires spec constants with the same name across multiple modules to have the same type.  Changing the value of a specialization constant will force the shader to re-prepare itself.

## Checklist

### DIFF
--- a/panda/src/glstuff/glShaderContext_src.cxx
+++ b/panda/src/glstuff/glShaderContext_src.cxx
@@ -3425,7 +3425,11 @@ attach_shader(const ShaderModule *module) {
                                 (const char *)spv->get_data(),
                                 spv->get_data_size() * sizeof(uint32_t));
       }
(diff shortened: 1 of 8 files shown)


rdb:
Thanks!

PR #1084, opened 2020-12-27
missing include for subdivideCollision()

janEntikan:
When running NodePath.subdivideCollisions(some_int) there is an import error.
Added the import, fixed the problem.


rdb (approved):
Thank you!

PR #1093, opened 2021-01-02
direct: Fix actor initialization

darktohka:
## Issue description

Actors were acting up in extraordinary ways!

Turns out that the Actor `__init__` method was broken in be9dde1eee22e8d83c44399f042c1a12aa18942a.

## Solution description

If Actor_initialized is set, actors will not be re-initialized anymore (this has always been the default behavior)

## Checklist


rdb (approved):
Whoops.  Thanks.

That said, it's pretty bizarre that we allow double-initializing Actor objects to begin with.

PR #1098, opened 2021-01-25
WIP: Cube map support for TexturePeeker

Moguri:
## Issue description
TexturePeeker does not support cube map textures

## Solution description
This PR adds support for cube maps and reduces some duplicate code. Marking this as draft to get the code out, but there is still some testing that should be done (e.g., checking if wrapping works as intended).

## Checklist


rdb:
Thanks for your work on this!  Going to check this in for 1.10.9.  I did find one mistake in the calculation of the UVs for the Y direction, but I've resolved that.

I don't think we need to worry about wrapping.  `fetch_pixel` doesn't do wrapping to begin with, and `lookup` takes a vector, so no wrapping can happen. :)

Issue #1047, opened 2020-11-07
Incorrect merging of texture data.


serkkz:
When exporting an object with different values of the same texture. Panda3D merge them into one by file name, and the rest will be lost.

Test egg:
```
<CoordinateSystem> { Z-Up }

<Texture> Texture1 {
 "panda.jpg"
 <Scalar> format { RGB }
 <Scalar> alpha { OFF }
 <Scalar> compression { OFF }
 <Scalar> envtype { MODULATE }
 <Scalar> anisotropic-degree { 16 }
 <Scalar> wrapu { REPEAT }
 <Scalar> wrapv { REPEAT }
 <Scalar> minfilter { LINEAR_MIPMAP_LINEAR }
 <Scalar> magfilter { LINEAR_MIPMAP_LINEAR }
}

<Texture> Texture2 {
 "panda.jpg"
 <Scalar> format { RGB }
 <Scalar> alpha { ON }
 <Scalar> compression { DXT5 }
 <Scalar> envtype { MODULATE }
 <Scalar> anisotropic-degree { 0 }
 <Scalar> wrapu { REPEAT }
 <Scalar> wrapv { BORDER_COLOR }
 <Scalar> minfilter { LINEAR_MIPMAP_LINEAR }
 <Scalar> magfilter { LINEAR_MIPMAP_LINEAR }
}

<Group>  Plane {
 <VertexPool> Plane {
  <Vertex> 0 { -1 -1 0 
  <UV> { 0 0 }
  }
  <Vertex> 1 { 1 -1 0 
  <UV> { 1 0 }
  }
  <Vertex> 2 { 1 1 0 
  <UV> { 1 1 }
  }
  <Vertex> 3 { -1 1 0 
  <UV> { 0 1 }
  }
  <Vertex> 4 { -1 1.40914 0 
  <UV> { 0 0 }
  }
  <Vertex> 5 { 1 1.40914 0 
  <UV> { 1 0 }
  }
  <Vertex> 6 { 1 3.40914 0 
  <UV> { 1 1 }
  }
  <Vertex> 7 { -1 3.40914 0 
  <UV> { 0 1 }
  }
 }

 <Polygon> 0 { 
  <TRef> { Texture1}
  <Normal> { 0 0 1 }
  <VertexRef> { 0 1 2 3 <Ref> { Plane } }
  }
  
 <Polygon> 1 { 
  <TRef> { Texture2 }
  <Normal> { 0 0 1 }
[...]

PR #1105, opened 2021-02-06
Textture 1047

paustian:
## Issue description
A first attempt at solving [issue 1047](https://github.com/panda3d/panda3d/issues/1047), the incorrect merging of texture data. 

## Solution description
I was able to change the functions calls such that the modifications detect the format, compression, and wrap parameters of all the textures that are loaded and can compare values between them. The implementation was able to successfully differentiate the textures in the test code and print out the parameters correctly. Here is a brief summary of how this was implemented:

Textures for models are stored in texturePool. These are kept in a map template, with the key differentiating them. Previously, the LookupKey only differentiated textures by their file path and the number of channels. Thus, if a file had more than one texture in it, the second would get ignored. To solves this I added variabkes to the LookupKey for each scalar. These are then added to LookupKey as thus:
```
struct LookupKey {
    Filename _fullpath;
    Filename _alpha_fullpath;
    int _primary_file_num_channels = 0;
    int _alpha_file_channel = 0;
    int _texture_format;
    int _texture_quality;
    int _texture_compress;
    SamplerState _texture_sampler;
    Texture::TextureType _texture_type = Texture::TT_2d_texture;
  };
```
When the .egg file is read, the textures are parsed into an EggTexture.  During load_texture all the options can be explored. The final piece of the implementation is to change the inline < operator in texturePool.I such that it also checks the new variables.

## Checklist
* [ ] …the changed code is adequately covered by the test suite, where possible. I have not implemented a test suite yet because I want the ok that this is an acceptable solution.


paustian:
I have reimplemented this using the SamplerState to encapsulate many of the properties of the texture. I do have a few questions about the implementation (I apologize they are a bit noobish)

1. SamplerState has to persist after the function exits, thus I decided to allocate memory for it. This then gets passed to the texutrePool which is responsible for deallocating it. The pointer in SamplerState in LoaderOptions gets transferred to LookupKey. If the item gets erased in the map object, then the sampler state is deleted. See the changed methods in texturePool.cxx
2. I altered apply_texture_attributes in eggLoader.cxx. I know it's a public function, but it is not called outside for eggLoader.cxx so I assume the is ok. If not let me know.
3. In apply_texture_attributes I call tex->set_default_sampler(*sampler) using the pointer of the sampler I allocated. set_default_sampler makes a copy of this object, so I think this is safe if I destroy the sampler later. Right?
4. I want to make sure the deallocation of the SamplerState is working correctly. How can I trigger an erase event so that a texture gets removed from the texture map?
In my tests it's working as expected. I will make a more robust PytTest

paustian:
There may still be an issue with ns_add_texture which keys against the file path only, and dumps it. Then does not consider the new key structure. It would therefore add the texture to the texture pool and not have the extra sampler data.

paustian:
After conversations with the head developers, SamplerState was switched back to a objet instead of a pointer and then passed as a direct parameter to TexturePool. This eliminated a circular reference and dependency for EggTexture in LoaderOptions.

panda/src/gobj/texturePool.h
```diff
@@ -175,6 +185,10 @@ class EXPCL_PANDA_GOBJ TexturePool {
     Filename _alpha_fullpath;
     int _primary_file_num_channels = 0;
     int _alpha_file_channel = 0;
+    int _texture_format;
+    int _texture_quality;
+    int _texture_compress;
```
rdb, line 190: Any particular reason these are ints instead of the appropriate enum from texture.h?

paustian: The enums are defined in eggTexture.h. If I add that header file to LoaderOptions.h doesn't that cause the LoaderOptions to now be required in the core when it is supposed to be optional?

rdb: I was talking about the enums in texture.h, but I realised that that's in gobj whereas LoaderOptions is in putil, so you're right that that wouldn't work unless we moved the enums.

paustian: I am not sure if that is worth doing. The values are for comparison only the way I use them. However, others might have a use for them.

paustian: Any idea why the Buildsystem checks are failing?

paustian: I have some time this weekend. Let me know if there is more to do on this.

rdb: The build failures are unrelated to your changes; if you merge in from latest master they should go away.

panda/src/putil/loaderOptions.h
```diff
@@ -48,16 +48,23 @@ class EXPCL_PANDA_PUTIL LoaderOptions {
...
   INLINE void set_flags(int flags);
   INLINE int get_flags() const;
   MAKE_PROPERTY(flags, get_flags, set_flags);
-
   INLINE void set_texture_flags(int flags);
   INLINE int get_texture_flags() const;
+  INLINE void set_texture_format(int format);
+  INLINE int get_texture_format() const;
+  INLINE void set_texture_compress(int  compress);
+  INLINE int get_texture_compress() const;
```
rdb, line 63: I'm not really a fan of abbreviating an existing property, so I would prefer if this were named `texture_compression`

panda/src/putil/loaderOptions.I
```diff
@@ -55,6 +58,44 @@ get_texture_flags() const {
...
+
+INLINE void LoaderOptions::
+set_texture_quality(int quality){
+  _texture_quality = quality;
+}
+
+INLINE int LoaderOptions::
+get_texture_quality() const {
+  return _texture_quality;
+}
```
rdb, line 97: The `/**` blocks are used to generate the API reference, so there should ideally be one for every method, with an appropriate message (though getters/setters are usually pretty obvious, so it's not that pressing)

panda/src/putil/loaderOptions.I
```diff
@@ -18,6 +18,9 @@ constexpr LoaderOptions::
 LoaderOptions(int flags, int texture_flags) :
   _flags(flags),
   _texture_flags(texture_flags),
+  _texture_format(0),
+  _texture_compress(0),
+  _texture_quality(1),
```
rdb, line 23: I would prefer if texture_quality defaulted to 0 (`QL_default`), which tells Panda to use whatever is specified in the configuration file.

panda/src/gobj/texturePool.I
```diff
@@ -327,5 +327,22 @@ operator < (const LookupKey &other) const {
...
+
+  if(_texture_compress != other._texture_compress){
+    return _texture_compress < other._texture_compress;
+  }
+
+  if(_texture_quality != other._texture_quality){
+    return _texture_quality < other._texture_quality;
+  }
+
+  if(_texture_sampler != other._texture_sampler){
```
rdb, line 343: The nitpickiest nitpick: `if(` should be `if (`

panda/src/gobj/samplerState.I
```diff
@@ -14,6 +14,7 @@
 /**
  * Creates a new SamplerState initialized to the default values.
  */
+
```
rdb, line 17: Should not add newline between docstring and method (this may have been added by mistake)

rdb (changes requested):
This is good, thank you for working on this!  Having looked a little more closely at it, I have one concern though.

We now have the additional parameters added to the LoaderOptions/TexturePool API so we can add them to the lookup key, and that's a great step.  But to someone using the TexturePool/LoaderOptions API directly, these parameters don't appear to do anything.  Someone using those calls would reasonably expect that, say, passing in a certain compression mode would actually have a certain effect on the texture.  It would be nice if TexturePool actually directly applied these parameters to the texture, instead of doing it in `apply_texture_attributes`.

This would also result in a cleaner architectural separation between the "core" loader/texture modules and the "egg" module, rather than having EggTexture enums appear (even as ints) inside a "core" class, and make this feature useful beyond just the egg loader.

That would require:
* Having the format¹/quality/compress enums in LoaderOptions be set to the Texture equivalents, not to the EggTexture version of those enums.  So instead of doing:
```cpp
  options.set_texture_compress(egg_tex->get_compression_mode());
``` 
you would be doing:
```cpp
  options.set_texture_compress(convert_compression_mode(egg_tex->get_compression_mode()));
```
…and similar for format and quality.
* Having TexturePool apply the parameters at the end of the load calls, rather than doing it in `apply_texture_attributes`.  The latter will become almost¹ vestigial.

I think this would result in a more coherent API—people using the more central TexturePool/LoaderOptions API (which underpins the Python Loader class, but is also directly accessible to both Python and C++ users) would be able to pass their desired parameters in through LoaderOptions/SamplerState and have them behave as they might expect.

What do you think?

——
¹ I do foresee some difficulty with the format setting, since `apply_texture_attributes` does some checks/warnings based on the number of components of the actual loaded texture, which we can't do if we're determining the converted format before even loading the texture! 
 This logic would need to be restructured a bit—we can determine the converted texture format assuming that the number of components will match whatever format is given, and later display any warnings *after* the texture has been loaded.

(An alternative approach would be to preload the texture header early on to determine its channel count beforehand, but I think that will be more fragile and less performant.)

panda/src/putil/loaderOptions.h
paustian: Fixed

panda/src/gobj/texturePool.I
paustian: I fixed all of them.

panda/src/putil/loaderOptions.I
paustian: This is trivial to add so I did it.

panda/src/gobj/samplerState.I
paustian: Yep, it was. Return removed. I did have modifications to that class in earlier implementations.

paustian:
>  Someone using those calls would reasonably expect that, say, passing in a certain compression mode would actually have a certain effect on the texture. It would be nice if TexturePool actually directly applied these parameters to the texture, instead of doing it in `apply_texture_attributes`.

Thanks for looking at the changes and for the feedback. So this would involve moving apply_texture_attributes out of EggLoader and putting the real code in TexturePool. That way if anyone else uses LoaderOptions to call into TexturePool, we could apply the attributes there.

rdb:
> So this would involve moving apply_texture_attributes out of EggLoader and putting the real code in TexturePool. That way if anyone else uses LoaderOptions to call into TexturePool, we could apply the attributes there. It looks like I can move the function call to texture pool and remove it from EggLoader

I think the code in TexturePool will look very simple, just something like:
```cpp
if (options.get_texture_format() != 0) {
  tex->set_format((Texture::Format)options.get_texture_format());
}
````
…or something to that effect.

> Right now it seems to apply attributes after partially loading the texture?

Right now, EggLoader loads the texture from the texture pool, and then applies the desired attributes to the loaded texture.  In the suggested approach, EggLoader will just pass the attributes along to the TexturePool, and then TexturePool will apply the attributes.

> 1. Pick out the desired texture format from whatever scalar is given and add those to the loader options
> 2. Load the texture
> 3. Do the checks after the texture is loaded and then post the warnings 

Yes, that sounds right.  An important detail is that the flags stored on the LoaderOptions will be the converted Texture::Format enumeration, not the original EggTexture::Format enum.

paustian:
Ok I think I am all done. I created a pytest (/tests/egg2pg/test_texture_loading.py) for these textures and it checked it appropriately. I think it is ready for final review.

rdb:
Thanks for your contribution.  I've finished the remaining work and pushed it to master.

paustian:
I am happy to contribute and I am glad it made it into the master branch :)

Issue #1021, opened 2020-09-19
Filled Wireframes do not behave as expected on ShaderTerrainMesh nodes


thetestgame:
## Description
with ``set_render_mode_wireframe`` on a ShaderTerrainMesh node you get the expected result of a wireframe of the terrain. https://gyazo.com/2a170966601fd5759f0e43c55a1d1005

However when using ``set_render_mode_filled_wireframe`` the wireframe is a flat sheet under the terrain and does not take into account the terrain's heightfield data. https://gyazo.com/31fdd51827dbd003cbb0d03092e66378

## Steps to Reproduce
Call ``set_render_mode_filled_wireframe`` on the Shader-Terrain sample

## Environment
 * Operating system: Windows
 * System architecture: x64
 * Panda3D version: 1.10.7
 * Installation method: pip
 * Python version (if using Python): 3.8
 * Compiler (if using C++):

rdb:
I think the problem is that we clear the shader on the wireframe part.

PR #1124, opened 2021-03-03
Fix filled wireframe for nodes with vertex transform in shader

el-dee:
## Issue description

Filled wireframe does not work on nodes which have a shader that perform vertex transform, see #1021

## Solution description

When creating the filled wireframe overlay state, we copy the shader attrib from the node and use the wireframe color as the blend color to replace the flat color mode of the non-shader version. 

The blend parameter could be wrong though, I could not configure them to have exactly the same output as with the current implementation.

I tried also to disable color write from the fragment shader, but it remove the whole wireframe altogether.

## Checklist

### DIFF
--- a/panda/src/pgraph/config_pgraph.cxx
+++ b/panda/src/pgraph/config_pgraph.cxx
@@ -373,6 +373,13 @@ ConfigVariableBool allow_live_flatten
           "only has an effect when Panda is not compiled for a release "
           "build."));
 
+ConfigVariableBool filled_wireframe_apply_shader
+("filled-wireframe-apply-shader", false,
+ PRC_DESC("Set this true to apply any shader configured on nodes onto the"
+          "filled wireframe overlay. It is helpful when the shader alters the"
+          "position of the vertices and makes the overlay wrong. This variable"
+          "has no effect on nodes without shader configured."));
+
 /**
  * Initializes the library.  This must be called at least once before any of
  * the functions or classes in this library can be used.  Normally it will be
--- a/panda/src/pgraph/config_pgraph.h
+++ b/panda/src/pgraph/config_pgraph.h
@@ -74,6 +74,8 @@ extern ConfigVariableString default_model_extension;
 
 extern ConfigVariableBool allow_live_flatten;
 
+extern ConfigVariableBool filled_wireframe_apply_shader;
+
 extern EXPCL_PANDA_PGRAPH void init_libpgraph();
 
 #endif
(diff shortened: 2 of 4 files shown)


rdb:
Interesting idea.  Can't go in 1.10.9 though due to ABI change.  It'd be fine to create an overloaded version of the method, and have the old one to redirect to the new one.
 
Unfortunately it's impossible to entirely override the incoming color using a ColorBlendAttrib.  We'd need some feature in the shader system that only replaces the fragment shader but keeps the vertex shader.  Perhaps something for when we have the SPIR-V pipeline.

el-dee:
> Interesting idea. Can't go in 1.10.9 though due to ABI change. It'd be fine to create an overloaded version of the method, and have the old one to redirect to the new one.

Oops, indeed ! I added the overloaded method and I kept it non-inline.


> Unfortunately it's impossible to entirely override the incoming color using a ColorBlendAttrib. We'd need some feature in the shader system that only replaces the fragment shader but keeps the vertex shader. Perhaps something for when we have the SPIR-V pipeline.

Yes it's only a workaround, the filled wireframe is really helpful to check complex geometry so even not perfect it still helps (at least it helped me :) )

If the new SPIR-V pipeline allows that it would be really helpful !

rdb:
Thinking more about this, since it otherwise is a breaking change, I think for 1.10.x I would rather have a Config.prc variable switch this (rather than the presence of a shader).  It would default to false in 1.10.x and true in 1.11.0.

el-dee:
I added the configuration variable, `filled-wireframe-apply-shader` which defaults to false right now. I choose to still apply the new blend parameters only if there is a shader, so the modification won't change the filled wireframe appearance for nodes without any shader.

rdb:
I don't agree with that, because whether or not it has a shader isn't relevant—rather whether the shader modifies the vertex position.  We don't have a reliable way of determining that.

In absence of that, I would rather that this change be applied consistently, so that users don't have to wonder why eg. enabling the shader generator would cause the filled wireframe appearance to be different.  They'd just choose between two filled wireframe modes, one which does work with custom shaders, and one which doesn't.

I also realised that it would be relatively easy to write some code that replaces the fragment shader with one that just writes out a flat color.  But I would rather support an easy solution that works for now and keep that option for the master branch (probably after we have the new shader pipeline so we can avoid the GLSL version mess).

I will try to modify your PR to always apply the alternate wireframe mode if the setting is provided, otherwise this'll have to wait for 1.10.10.

PR #1131, opened 2021-03-27
Task object fails 1097

paustian:
## Issue description
<!-- What is this change intended to accomplish?  What problem does it solve?
    This change fixes issue 1097 and allows calls to base.addTask with the task

## Solution description
I added a check in addTask in DirectObject.py it checks to see if the task is the only argument submitted and then grabs the other parameters that it needs to implement the task. I also added a pytest to verify that this is indeed working.

### DIFF
--- a/direct/src/showbase/DirectObject.py
+++ b/direct/src/showbase/DirectObject.py
@@ -44,6 +44,12 @@ class DirectObject:
     def addTask(self, *args, **kwargs):
         if not hasattr(self, "_taskList"):
             self._taskList = {}
+        #check to see if our first and only variable is a task, if so grab the args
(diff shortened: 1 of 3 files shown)


rdb (changes requested):
Thanks for working on this.  However, I think this is entirely the wrong place to address this.  `DO.add_task` just forwards the args and kwargs to `taskMgr.add(...)`, which with your PR still has the problem.  If it were instead addressed in taskMgr.add (specifically in `__setupTask`) by changing the handling of the extraArgs argument, I think it would solve the problem more universally.

paustian:
Yeah I wondered about that. I always hesitate to hack on things that are more universal. I need to just go for it.

paustian:
I am going to close this pull request and start over since it is completely different files.

Issue #1097, opened 2021-01-23
Adding task object with extraArgs fails


K9Kraken:
## Description
When adding a taskObject back in to the task manager Panda3D crashes if the task has extraArgs.

## Steps to Reproduce
**This code works fine:**
```
def test(task):
    ship.y += 0.01
    return task.cont
task = base.add_task(test, 'test_task')
base.remove_task(task)
base.add_task(task)
```

**This code crashes Panda3D:** **(Results are the same when using task_mgr.remove and task_mgr.add)**
```
def test(ship, task):
    ship.y += 0.01
    return task.cont
task = base.add_task(test, 'test_task', extraArgs=[ship], appendTask=True)
base.remove_task(task)
base.add_task(task)
```

Console output:
> TypeError: test() missing 1 required positional argument: 'task'
> :task(error): Exception occurred in PythonTask test_task
> EXIT
> Traceback (most recent call last):
>   File "/home/will/Sync/projects/panda/game/main.py", line 44, in <module>
>     ez.run()
>   File "/home/will/.local/lib/python3.9/site-packages/direct/showbase/ShowBase.py", line 3325, in run
>     self.taskMgr.run()
>   File "/home/will/.local/lib/python3.9/site-packages/direct/task/Task.py", line 541, in run
>     self.step()
>   File "/home/will/.local/lib/python3.9/site-packages/direct/task/Task.py", line 495, in step
>     self.mgr.poll()
> TypeError: test() missing 1 required positional argument: 'task'

## Environment
 * Operating system: Manjaro
 * System architecture: x64
 * Panda3D version: 1.10.8
 * Installation method: pip
 * Python version: 3.9.1

K9Kraken:
A work around for this is to make my own function:
```
def add_task( task ):
    base.add_task(task, task.name, extraArgs=task.get_args())
```

rdb:
ShowBase has no method called `add_task`.  I take it you meant to write `base.task_mgr.add` and `base.task_mgr.remove`?

When you add it back, you need to pass the same arguments:
```python
task = base.task_mgr.add(test, 'test_task', extraArgs=[ship], appendTask=True)
base.task_mgr.remove(task)
base.task_mgr.add(task, extraArgs=[ship], appendTask=True)
```

rdb:
I suppose we could default both arguments to `None` and have it use the existing values if they are `None`.  Do others have thoughts on this?

K9Kraken:
[...]

PR #1132, opened 2021-03-28
task: Fix re-adding task object removing extraArgs

paustian:
## Issue description
This change resolves issue 1097 correctly implementing a task without any arguments. The task gets created on a previous call, removed, and then later added back to the task manager. It is convenient for the programmer to be able to just send the task, instead of all the other arguments again. This change allows that.
#1097 
## Solution description
A few changes were made in Task.py __setupTask to check for the task object coming in and harvest the arguments needed.


Moguri (approved):
Looks good to me.

PR #1134, opened 2021-03-31
maya: Allow more than three eggObjectTypes in a Maya model

darktohka:
## Issue description
As of right now, a Maya group node can contain three egg object types: `eggObjectTypes1`, `eggObjectTypes2` and `eggObjectTypes3`. Some nodes may have more than three object types set. These are lost during the maya2egg conversion process.

## Solution description
Instead of checking `eggObjectTypes1`, `eggObjectTypes2` and `eggObjectTypes3` for object type information, we will now check all attributes matching `eggObjectTypes[1...inf]`. As soon as we find an egg object type that is not set in this fashion, we will stop looking for object types to add to the group. This approach is backwards compatible.

The `eggObjectFlags.mel` Maya script has also been changed to allow more than 3 object types to be added. `eggObjectFlags.mel` is responsible for adding one egg object type to a group node. Before this commit, it would print a warning if more than 3 object types are present, but now, it will find the first empty object type slot to insert an object type into, removing this 3 object type limitation.

## Checklist

### DIFF
--- a/pandatool/src/maya/eggObjectFlags.mel
+++ b/pandatool/src/maya/eggObjectFlags.mel
@@ -1,6 +1,6 @@
 //
-// This is a sample mel script to flag a Maya node with one, two, or
-// three eggObjectTypes* attributes.  These attributes are used to tag
+// This is a sample mel script to flag a Maya node with at least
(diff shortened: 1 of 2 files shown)


rdb (approved):
Thanks, looks good.

PR #1147, opened 2021-04-11
Properly handle update fields in Python if ConnectionRepositoryBase does not have owner view.

Souzooka:
## Issue description
Currently, if DO object updates are handled in Python in ConnectionRepositoryBase, and the repository does not have owner view, then the engine will crash as `__doUpdateOwner` will attempt to access an undefined dictionary `doId2ownerView`.

## Solution description
This pull request simply adds a check in `__doUpdateOwner` to see if the repository has owner view and returns early if not, so the nonexistent property is not accessed.

## Checklist

### DIFF
--- a/direct/src/distributed/ClientRepositoryBase.py
+++ b/direct/src/distributed/ClientRepositoryBase.py
@@ -468,6 +468,9 @@ class ClientRepositoryBase(ConnectionRepository):
                         "Asked to update non-existent DistObj " + str(doId) + "and failed to find it")
 
     def __doUpdateOwner(self, doId, di):
+        if not self.hasOwnerView():
+            return False
+
         ovObj = self.doId2ownerView.get(doId)
         if ovObj:
             odg = Datagram(di.getDatagram())


rdb:
I need a second opinion on this from someone with in-depth knowledge of the distributed system.

loblao:
This is a good contribution. As the docstring states, "this method is exactly duplicated by the C++ method", which is not true without the proposed fix.

On the other hand, the docstring also states that `handleUpdateField` only exists for "documentation purposes". In practice, this is not true.

This Python handler will be called when either `_handle_datagrams_internally` or `_handle_c_updates` is false. However, `setDeferInterval` also does `setHandleCUpdates(False)` when the config `deferred-generate-interval` is non-zero. [The default value is 0.2 so using the Python handler is indeed what happens by default in clients](https://github.com/kestred/panda3d/blob/16bfd3750f726a8831771b81649d18d087917fd5/direct/src/distributed/ClientRepositoryBase.py#L44). Handling updates in C++ is the default behavior for non-client processes (e.g., AI and UD servers), which [usually inherit from ConnectionRepository directly](https://github.com/Astron/panda3d/blob/88b66f63d462d35665cf45dccaf354f397c9c1fb/direct/src/distributed/AstronInternalRepository.py#L89). In those processes `_has_owner_view` will also be false.

As a side note, having both `_handle_datagrams_internally` and `_handle_c_updates` seems redundant: they have the very same effect.

Disabling OV support (`_has_owner_view = false`) is [the default behavior in AI processes](https://github.com/Astron/panda3d/blob/88b66f63d462d35665cf45dccaf354f397c9c1fb/direct/src/distributed/AstronInternalRepository.py#L107) but uncommon in clients. Apparently, having it disabled serves primarily as an optimization: it skips a dict lookup (compare `handle_update_field` and `handle_update_field_owner`).

In summary, this is the usual setup:

| Process Type | _handle_datagrams_internally | _handle_c_updates  | _has_owner_view  |
|--------------|---|---|---|
| Client | _don't care_ | false (due to `setDeferInterval`) | true |
| AI | [true](https://github.com/panda3d/panda3d/blob/cbfd8f45450b0af296500d9f071f44dc4d91fc58/direct/src/distributed/config_distributed.cxx#L47) | true | false |

What the OP seems to be doing, i.e., disabling OVs + using the Python handler is a very exotic combination.

rdb:
Thanks for weighing in!  I appreciate it.

PR #1161, opened 2021-05-04
bullet clang compilation with ndk

pmp-p:
seems to fix clang compilation about "ambiguous declaration" on ndk with bullet git for me. I have no idea possible side effects on other builds.


rdb (approved):
LGTM, let's wait for the CI.

Issue #1202, opened 2021-10-31
Bug related to Sequence.play_rate property


Epihaius:
## Description
Hi,

Recently I discovered an inconsistency between the behaviour of `Sequence.set_play_rate` and assigning to the `Sequence.play_rate` property.
There seems to be an issue with the latter, which becomes apparent when the Sequence contains a `Func` object. In this case, I observe the following problem:

when `Sequence.start` is called, followed by an assignment to the `Sequence.play_rate` property, the Sequence plays all of the contained intervals, until a `Func` object is encountered. At that point, the callable wrapped by the Func doesn't get called and the Sequence either:

* stops (if the `Func` object is last in the Sequence);
when closing the Panda window, the following is output to the console:
  ```
  :interval(warning): Losing outstanding events for Sequence-1 dur 4
  ```

* or crashes (if the `Func` object is followed by a non-`Func` object);
in this case an assertion error is thrown:
  ```
  File "D:\Panda3D-1.10.10-x64\direct\showbase\ShowBase.py", line 3325, in run
    self.taskMgr.run()
  File "D:\Panda3D-1.10.10-x64\direct\task\Task.py", line 546, in run
    self.step()
  File "D:\Panda3D-1.10.10-x64\direct\task\Task.py", line 500, in step
    self.mgr.poll()
  File "D:\Panda3D-1.10.10-x64\direct\showbase\ShowBase.py", line 2097, in __ivalLoop
    IntervalManager.ivalMgr.step()
  File "D:\Panda3D-1.10.10-x64\direct\interval\IntervalManager.py", line 95, in step
    self.__doPythonCallbacks()
  File "D:\Panda3D-1.10.10-x64\direct\interval\IntervalManager.py", line 121, in __doPythonCallbacks
    index = self.getNextEvent()
  AssertionError: (def._flags & F_external) != 0 at line 311 of c:\buildslave\sdk-windows-amd64\build\direct\src\interval\cIntervalManager.cxx
  ```

So it seems there's a difference in the handling of `Func` objects between the `Sequence.play_rate` property and the `set_play_rate` method, as the latter causes expected behaviour.

## Steps to Reproduce
In the following code sample, a `Func` object is placed at the end of a Sequence; although `Sequence.start` doesn't lead to an AssertionError in this case, calling `Sequence.loop` instead does:

```python
from direct.showbase.ShowBase import ShowBase
from panda3d.core import *
from direct.interval.IntervalGlobal import *


base = ShowBase()

[...]

PR #1204, opened 2021-11-01
Redefine play_rate property in MetaInterval class

Epihaius:
## Issue description
The `play_rate` property of classes derived from `MetaInterval` invokes the underlying C++ `set_play_rate` method, which leads to a bug (#1202).

## Solution description
The property is now redefined in `MetaInterval` to make it invoke the overriding `MetaInterval.set_play_rate` method instead.

### DIFF
--- a/direct/src/interval/MetaInterval.py
+++ b/direct/src/interval/MetaInterval.py
@@ -348,9 +348,13 @@ class MetaInterval(CMetaInterval):
     def getManager(self):
         return self.__manager
 
+    manager = property(getManager, setManager)
+
     def setT(self, t):
         self.__updateIvals()
         CMetaInterval.setT(self, t)
+
+    t = property(CMetaInterval.getT, setT)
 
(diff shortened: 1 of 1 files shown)


Epihaius:
Ah, indeed, there are these methods as well:

* `get_manager` and `set_manager`;
* `set_t`;
* `get_duration`.

The last two call `MetaInterval.__updateIvals`, while the first two access `MetaInterval.__manager`, so the corresponding properties should be redefined as well, it looks like.

This should now be done.

Epihaius:
Now I do see in `Interval.py` that the `Interval.isPlaying` method exists which looks like this:

```python
    def isPlaying(self):
        return taskMgr.hasTaskNamed(self.getName() + '-play')
```

There doesn't seem to be a `playing` property defined for this in the `Interval` class; should it be added as well?

rdb:
Interval doesn't inherit from CInterval, so I think that's out of the scope of this issue.

Epihaius:
Ah you're right, sorry!

PR #1205, opened 2021-11-04
Add ability to get name from AICharacter

Clockwork-Muse:
## Issue description
Removing characters from an AIWorld requires passing the character name; however, there's currently no way to get the name of a given character.

## Solution description
Add the ability to get the name of a character.

## Checklist

### DIFF
--- a/contrib/src/ai/aiCharacter.cxx
+++ b/contrib/src/ai/aiCharacter.cxx
@@ -13,6 +13,8 @@
 
 #include "aiCharacter.h"
 
+using std::string;
+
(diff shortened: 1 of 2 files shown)


Clockwork-Muse:
The other way to do this would be to add a method to remove a character directly, instead of using a name.
The catch is that this could actually be _different_ behavior, since remove-by-name removes only one instance of the name (probably the first one)...

rdb:
Thanks, checked in with a slight change: it is accessed in Python via a `.name` property instead of `.get_name()` method.

PR #1217, opened 2021-12-12
showbase: Add type annotations to ShowBase/ClientRepositoryBase attributes

Souzooka:
## Issue description
Currently, a lot of attributes contained on ShowBase have properties with non-inferable types, and as such these attributes nor their methods/attributes are hinted in IDEs. 

## Solution description
This MR adds type annotations to most properties defined during instance initialization for ShowBase (and a few for ClientRepositoryBase), so that developers which access these properties can determine easily determine the type of the attribute and the properties of its methods (if type information exists for said type, i.e. `panda3d.core` types).

A slight caveat is that I did not add typehinting for a few attributes declared in ShowBase, (e.g. `ShowBase.texmem`) because their types are imported via `importlib`, and I did not want to break the functionality that is provided via using that import method. 
Another caveat is that if the names used in type annotations are removed (e.g. the `from panda3d.core import *` is changed to import a list of specific names), types imported for annotations (i.e. if they are moved into a `if TYPE_CHECKING:` block) will need to be modified so that they are a string literal (e.g. `Optional["NodePath"]`) unless `annotations` is imported from `__future__` (Python 3.7+) or direct is running on Python 3.11+.

I've done my best to verify these type annotations are correct, but please correct me if I have gotten anything wrong.

## Checklist


direct/src/showbase/ShowBase.py
```diff
@@ -197,69 +200,69 @@ def __init__(self, fStartDirect=True, windowType=None):
...
         #: `makeDefaultPipe()`.
-        self.pipe = None
+        self.pipe: Optional[GraphicsPipe] = None
         #: The full list of :class:`~panda3d.core.GraphicsPipe` objects,
         #: including any auxiliary pipes.  Filled by `makeAllPipes()`.
-        self.pipeList = []
-        self.mouse2cam = None
-        self.buttonThrowers = None
-        self.mouseWatcher = None
+        self.pipeList: List[Optional[GraphicsPipe]] = []
```
rdb, line 216: Does this need to be Optional?  Will there ever be a None added to the list?

direct/src/showbase/ShowBase.py
```diff
@@ -289,11 +292,11 @@ def __init__(self, fStartDirect=True, windowType=None):
...
         #: need to call :meth:`~panda3d.core.CollisionTraverser.traverse()`
         #: yourself every frame.
-        self.cTrav = 0
-        self.shadowTrav = 0
+        self.cTrav: Optional[CollisionTraverser] = None
+        self.shadowTrav: Optional[CollisionTraverser] = None
         self.cTravStack = Stack()
         # Ditto for an AppTraverser.
-        self.appTrav = 0
+        self.appTrav: Optional[CollisionTraverser] = None
```
rdb, line 299: I don't think appTrav is meant to be a CollisionTraverser, but I don't know what it's meant to be - I would suggest leaving this Any for now

direct/src/showbase/ShowBase.py
```diff
@@ -275,7 +278,7 @@ def __init__(self, fStartDirect=True, windowType=None):
 
         #: The global :class:`~panda3d.core.GraphicsEngine`, as returned by
         #: GraphicsEngine.getGlobalPtr()
-        self.graphicsEngine = GraphicsEngine.getGlobalPtr()
+        self.graphicsEngine: GraphicsEngine = GraphicsEngine.getGlobalPtr()
         self.graphics_engine = self.graphicsEngine
```
rdb, line 282: Do we need to add an annotation to the underscored version as well?

direct/src/distributed/ClientRepositoryBase.py
```diff
@@ -14,6 +14,11 @@
 from .ClockDelta import *
 import time
 
+# Typing annotations
+from typing import Optional, TYPE_CHECKING
+if TYPE_CHECKING:
+    from direct.distributed.TimeManager import TimeManager
```
rdb, line 20: Not sure about such an optional import - and is it even necessary since you referred to it by string?

rdb (changes requested):
- Needs rebase
- See inline comments
- Do the annotations all work with Python 3.6?

direct/src/showbase/ShowBase.py
```diff
@@ -197,69 +200,69 @@ def __init__(self, fStartDirect=True, windowType=None):
         self.requireWindow = ConfigVariableBool('require-window', True).value
 
         #: This is the main, or only window; see `winList` for a list of *all* windows.
-        self.win = None
-        self.frameRateMeter = None
-        self.sceneGraphAnalyzerMeter = None
+        self.win: Optional[GraphicsWindow] = None
```
rdb, line 203: Actually, this may also be a GraphicsBuffer, in offscreen mode. But it should in any case be a GraphicsOutput.

rdb:
Ping.  Still interested in this?

PR #1220, opened 2021-12-14
Add rpmversion for packaging.

Clockwork-Muse:
## Issue description
The current way the package version is set for rpm builds is insufficient - although there is an `rpmrelease` field, this is insufficient for setting things like pre-release or post-release package versions.  Given conventions [in these guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/), a fuller versioning scheme is required.

Note that a "complete" rpm version is required if an effort is being made to build C++ modules for python against manylinux systems.

## Solution description
This PR just adds an `--rpmversion` option to mimic the `--debversion` option.

Adding additional versions this way may be unsustainable, though, if we anticipate more being added in the future.

## Checklist

### DIFF
--- a/makepanda/makepackage.py
+++ b/makepanda/makepackage.py
@@ -276,7 +276,7 @@ def MakeDebugSymbolArchive(zipname, dirname):
     zip.close()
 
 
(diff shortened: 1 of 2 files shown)


rdb (approved):
LGTM

Issue #1234, opened 2022-01-08
direct.gui.DirectGuiBase file missing snake_case aliases


Max-Rodriguez:
## Description
Panda3D is aiming to convert its naming convention from CamelCase named methods to snake_case names.
I found that the `DirectGuiBase` class in `direct.gui.DirectGuiBase` has no snake case aliases, so I'm requesting to add snake case aliases to this object. (will see if I can send in my first PR myself)

## Use Case
Adding the aliases will help converting to the snake case naming convention.

Max-Rodriguez:
Looking at the file, DirectGuiBase actually doesn't even use CamelCase aside from one of its methods.
It's mostly DirectGuiWidget that needs snake case aliases.

Max-Rodriguez:
I've sent a PR with my commit that adds the aliases. Please let me know if I need to change anything, I'm a newbie to this. 👍🏼

PR #1235, opened 2022-01-08
Added snake-case aliases to DirectGuiBase

maxrdz:
## Issue description
The `DirectGuiWidget` class was missing its snake-case aliases.
This resolves issue #1234 and helps Panda to convert to this naming convention.

## Solution description
This PR includes the references to the class methods for both classes in the DirectGuiBase.py file to include snake-case aliases for these methods.

## Checklist
* [ ✓ ] …I have familiarized myself with the CONTRIBUTING.md file
* [ ✓ ] …this change follows the coding style and design patterns of the codebase
* [ ✓ ] …I own the intellectual property rights to this code
* [ ✓ ] …the intent of this change is clearly explained
* [ ✓ ] …existing uses of the Panda3D API are not broken
* [ ✓ ] …the changed code is adequately covered by the test suite, where possible.


Max-Rodriguez:
This is my first PR, and my first 'real' PR so please let me know if anything needs to be change to the commit itself or the way im pushing my commit. 👍🏼

Max-Rodriguez:
Checks passed 👍🏼

Moguri (approved):

rdb:
I'm going to unapprove this PR.  Many of these, like guiScaleTask, setFrameSize, frameInitialiseFunc, etc. are actually internal methods and not meant for public use, so we shouldn't be making aliases for them.  A few of them, like get_bounds and set_state, are a good idea in principle but conflict with NodePath methods, so we need to consider whether we need to rename these methods anyway.

Issue #1182, opened 2021-08-03
Add option to bam-info for listing texture info


rdb:
## Description
It would be useful to add an option to bam-info to list all the referenced textures in the bam file.

## Use Case
The use case is similar to for egg-list-textures: list externally referenced textures, to be able to verify there are no absolute paths, and to know which textures need to be copied alongside the bam file.

Pranjal-dew:
hi there, can I contribute to this issue?

paustian:
So basically you are wanting someone to write a C++ program that does what eggListTextures.cxx does, but for bam files.

rdb:
Yes, precisely.

PR #1249, opened 2022-01-29
Bam info textures

paustian:
## Issue description
Added the capability to list the texture paths of .bam files to the baminfo program. When the -ltp option is used, any textures that contain full paths will be printed out on the command line. This resolves [issue #1182](https://github.com/panda3d/panda3d/issues/1182).

## Solution description
First, the _textures private variable in the class SceenGraphAnalyzer needed a getter so that I could graph the textures of any scene graphs and print them out. This involved making the map definition public and creating a getting function. 

Second, a -ltp option was created for baminfo and a write_texture function was added.


panda/src/pgraphnodes/sceneGraphAnalyzer.I
```diff
@@ -229,7 +229,6 @@ get_texture_bytes() const {
   return _texture_bytes;
 }
 
-
```
Moguri, line 232: This whitespace change is unnecessary and is the only change to this file. It should be removed to make the PR more concise.

panda/src/pgraphnodes/sceneGraphAnalyzer.h
```diff
@@ -118,7 +118,15 @@ class EXPCL_PANDA_PGRAPHNODES SceneGraphAnalyzer {
...
   UniqueVADatas _unique_prim_vadatas;
+
+public:
+  typedef pmap<Texture *, int> Textures;
+
+private:
   Textures _textures;
+  
+public:
+  INLINE Textures get_textures() const {return _textures;}
```
Moguri, line 129: The implementation should either be in the source file or the .I file

pandatool/src/bam/bamInfo.cxx
```diff
@@ -36,7 +38,7 @@ BamInfo() {
      "and Models native binary format--and describes their contents.");
 
   clear_runlines();
-  add_runline("[opts] input.bam [input.bam ... ]");
+  add_runline("[-ls -t -g -ltp] input.bam [input.bam ... ]");
```
Moguri, line 41: Is this the style used by other Panda cli tools? It looks like if one option is specified, then they all need to be there. I think leaving it as `[opts] ` is fine

Moguri (changes requested):

Moguri:
Closing due to inactivity. Please feel free to reopen the PR  if/when you continue working on this.

PR #1256, opened 2022-02-02
Stop polluting builtins from PythonUtil.

Maxwell175:
## Issue description

This PR prevents PythonUtil from polluting the builtins with it's classes and functions.
Polluting builtins can cause many unintended bugs or oddities and is considered bad practice.

## Solution description

This PR removes the section in PythonUtil that adds all the functions and classes to builtins.
No utilities are being removed in this PR.
If you would like to use one of these classes, you would simply need to import them.

## Checklist


Moguri:
While I agree with the spirit of the changes, this looks like an API-breaking change.

loblao:
Messing with PythonUtil is guaranteed to upset those guys maintaining old Disney games. Though, given that it has happened a couple of times, I suspect they have stopped relying on PythonUtil by now, but who knows.

IIRC `getBase()` and `getRepository()`, for example, are used all over the code as builtins. TIL they are defined in PythonUtil.

If anything, I'd recommend removing the most obscure ones first.

Maxwell175:
Yeah, I think that the overall health of panda3d is more important. Polluting builtins with commonly used names is a recipe for bugs.

Once again this PR does not _remove_ anything. if anyone really wants to keep using their builtins they can add them manually in their own code, Or, you know, just `from direct.showbase.PythonUtil import *` or something of the sort. it is very trivial to workaround and it would improve panda3d code health.

Moguri:
1.11 is not the place for breaking changes. Breaking changes should only occur in a major version bump (e.g., 2.0).

However, I agree that some of those names are problematic and ripe for unexpected name collisions. I do not know @rdb's opinion on the matter, but I would prefer we at least move these builtin assignments to a new module so the behavior can be restored with one line of code (importing the new module).

Maxwell175:
> 1.11 is not the place for breaking changes. Breaking changes should only occur in a major version bump (e.g., 2.0).

This is by no means a documented API and there have been plenty of much more significant changes between say 1.9 and 1.10.

> [...] so the behavior can be restored with one line of code (importing the new module).

    builtins.__dict__.update(__import__('direct.showbase.PythonUtil', fromlist=['*']).__dict__)

PR #1261, opened 2022-02-05
makepanda: Create 7-zip debug symbol archives by default, if available

darktohka:
7-zip archives will only be created if 7-zip is available during the build phase. When 7-zip is unavailable, ZIP archives will be created as a fallback.

Benchmarks:

- Default ZIP compression: ~23.5 seconds, 162 MB
- 7-zip compression: ~7.5 seconds, 108 MB
- 7-zip compression, `--lzma` set: ~44 seconds, 88 MB
- 7-zip compression, solid archive: ~5 minutes, 83 MB (not implemented)

## Issue description

The makepanda system takes up quite some time creating debug archives on Windows. For how much time it's spending creating ZIP archives, the end result is quite disappointing: 23.5 seconds on average yields a 162 MB archive.

If we were packing these PDB files into a 7-Zip archive instead, we could save on both disk space and CPU cores.

## Solution description

When 7-zip is available, makepanda will now build 7-zip archives. There are two modes available:
- No `--lzma` set: an LZMA2 7-Zip archive will be created with compression level 3. An average archive will take around 7.5 seconds to create, and will weigh around 108 megabytes - a 33.3% size reduction (on my testing machine)
- `--lzma` set: an LZMA2 7-Zip archive will be created with default settings. An average archive will take around 44 seconds to create, and will weigh around 88 megabytes, a 46% size reduction (on my testing machine)

If 7-zip is unavailable, a simple ZIP archive will be created like before.

## Checklist

### DIFF
--- a/makepanda/makepackage.py
+++ b/makepanda/makepackage.py
@@ -189,29 +189,63 @@ def MakeInstallerNSIS(version, file, title, installdir, compressor="lzma", **kwa
     oscmd(cmd)
 
 
-def MakeDebugSymbolArchive(zipname, dirname):
-    outputdir = GetOutputDir()
-
+def MakeDebugSymbolZipArchive(zipname):
     import zipfile
 
-    zip = zipfile.ZipFile(zipname, 'w', zipfile.ZIP_DEFLATED)
+    outputdir = GetOutputDir()
+    zip = zipfile.ZipFile(zipname + '.zip', 'w', zipfile.ZIP_DEFLATED)
 
     for fn in glob.glob(os.path.join(outputdir, 'bin', '*.pdb')):
-        zip.write(fn, dirname + '/bin/' + os.path.basename(fn))
+        zip.write(fn, 'bin/' + os.path.basename(fn))
 
     for fn in glob.glob(os.path.join(outputdir, 'panda3d', '*.pdb')):
-        zip.write(fn, dirname + '/panda3d/' + os.path.basename(fn))
+        zip.write(fn, 'panda3d/' + os.path.basename(fn))
 
     for fn in glob.glob(os.path.join(outputdir, 'plugins', '*.pdb')):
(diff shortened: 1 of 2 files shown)


makepanda/makepackage.py
```diff
@@ -189,31 +189,65 @@ def MakeInstallerNSIS(version, file, title, installdir, compressor="lzma", **kwa
...
+def MakeDebugSymbolZipArchive(zipname):
     import zipfile
 
-    zip = zipfile.ZipFile(zipname, 'w', zipfile.ZIP_DEFLATED)
+    outputdir = GetOutputDir()
+    zip = zipfile.ZipFile(zipname + '.zip', 'w', zipfile.ZIP_DEFLATED)
 
     for fn in glob.glob(os.path.join(outputdir, 'bin', '*.pdb')):
-        zip.write(fn, dirname + '/bin/' + os.path.basename(fn))
+        zip.write(fn, 'bin/' + os.path.basename(fn))
```
rdb, line 199: Does this remove the top-level directory from the archives?  Why?

darktohka: To stay consistent with the archive that 7-zip creates. 7-zip has no feature that I know of that allows the top-level directory name to be set, so it doesn't create one. It is possible to make 7-zip emit an archive with a top-level directory, BUT it will be called either `built_x64` or `built`, it does not match the original name unfortunately.

`dirname` in this case was `C:\Panda3D-1.11.0-x64`, but the zipfile that was created had a top level directory of `Panda3D-1.11.0-x64` anyway which is wrong behavior on its own.

darktohka: If we want to keep the top-level directory, we can use 7-Zip's `rn` option to rename the top level directory after creating the original archive, but it does require two invocations of the 7-Zip command line utility. Should the top-level directory be restored?

Issue #1087, opened 2020-12-29
swaywm (xwayland): app eats 100% CPU without drawing until window changes first time


adabru:
## Description

I start a panda3d program via command line. The started application window pops up as expected and one first frame is rendered. Then no further frames are rendered and one python thread keeps busy at 100% CPU load.

As soon as I resize the window or change the window focus from another xwayland window to the panda3d window (doesn't work with native wayland windows), further frames are drawn as expected and the CPU isn't as busy anymore.

I tried the python debugger to narrow down the problem. The line I stopped at was https://github.com/panda3d/panda3d/blob/v1.10.8/direct/src/showbase/ShowBase.py#L2153 . I embraced the line (`self.graphicsEngine.renderFrame()`) with time measuring code and observed that before the first window change, the delta was constantly around 0.00001xxx, after the first window change the delta was constantly around  0.016xxx (=60fps) seconds.

## Steps to Reproduce

Run following code inside the sway window manager:

```python
from direct.showbase.ShowBase import ShowBase


class MyApp(ShowBase):
    def __init__(self):
        ShowBase.__init__(self)
        self.setFrameRateMeter(True)


app = MyApp()
app.run()
```

- Observe the fps meter (=stuck at its first 0.0) and your CPU load (=one CPU-core at 100%).
- Alternatively resize the window or focus another x window and then the panda3d window again. 
- Observe the fps meter (=updating around 60fps) and your CPU load (~10-20% for one CPU-core).

## Environment
 * Operating system: Arch Linux
 * System architecture: x64
 * Panda3D version: 1.10.8 (same with 1.10.7) 
 * Installation method: `pip install --user panda3d`
 * Python version (if using Python): 3.9.1
 * Compiler (if using C++): -

rdb:
This might be because of the `_awaiting_configure` check in `x11GraphicsWindow::begin_frame()`, which early-outs until a ConfigureNotify event has been received.  Perhaps XWayland behaves unexpectedly in this respect.

Moguri:
I can confirm that we are stuck on `_awaiting_configure` in `x11GraphicsWindow::begin_frame()`. It seems something with either Sway and/or xwayland does not always send Configure Notify events as expected (or at least the same way as X11). Doing some googling brings up some results with OBS running into issues with Sway and xwayland.

Moguri:
[...]

PR #1267, opened 2022-02-21
x11GraphicsWindow: Only wait to draw on window reconfigures

Moguri:
## Issue description
This is a fix for #1087 where Panda gets stuck waiting on an initial X11 window configure.

## Solution description
Instead of waiting on all configure/reconfigures, we only await on true Window reconfigures and stop waiting for a ConfigNotify for the initial window config.

Tested using
* i3wm+x11
* sway+xwayland
* kwin+x11
* kwin+xwayland

## Checklist

### DIFF
--- a/panda/src/x11display/x11GraphicsWindow.cxx
+++ b/panda/src/x11display/x11GraphicsWindow.cxx
@@ -116,6 +116,7 @@ x11GraphicsWindow(GraphicsEngine *engine, GraphicsPipe *pipe,
   }
 
   _awaiting_configure = false;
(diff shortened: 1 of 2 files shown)


Moguri:
This still needs to be tested on non-tiling window managers (e.g., Kwin or Mutter).

Moguri:
Now tested on kwin using both x11 and wayland.

rdb:
Thanks!  Closing this as per comment in #1087 since I don't believe this is the right fix, we can't actually rely on ConfigureNotify being sent in any case

PR #1275, opened 2022-03-01
mayaprogs: Resolve Maya conversion server memory leak by storing managers on the heap

darktohka:
## Issue description

As @rdb has pointed out, `MayaConversionClient` is creating three objects on the heap but neglecting to clean it up in a missing destructor.

## Solution description

Instead of creating a destructor, I've decided to take the objects off the heap completely. Instead, the constructor will initialize these connection management objects directly on the object on both `MayaConversionClient` and `MayaConversionServer`.

## Checklist

### DIFF
--- a/pandatool/src/mayaprogs/mayaConversionClient.cxx
+++ b/pandatool/src/mayaprogs/mayaConversionClient.cxx
@@ -19,11 +19,8 @@
  * Initializes the Maya conversion client.
  */
 MayaConversionClient::
-MayaConversionClient()
+MayaConversionClient() : _qReader(&_qManager, 0), _cWriter(&_qManager, 0)
 {
-  _qManager = new QueuedConnectionManager();
-  _qReader = new QueuedConnectionReader(_qManager, 0);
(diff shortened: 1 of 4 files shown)


rdb (commented):
Thanks!

Would you mind also fixing the odd naming convention issue? `_qManager` should just be `_manager`, etc.  I'd do it myself but I'm not building with Maya myself so I can't validate my changes as easily.

Thanks!

darktohka:
`_qManager` is short for `_queuedManager` and `_cWriter` is short for `_connectionWriter`, would you still like that to be changed?

rdb:
I don't care very much about the precise names that these variables have, just that they use snake_case instead of camelCase convention like the rest of the Panda codebase.  If it is really necessary to distinguish a queued writer from a non-queued writer then feel free to put in a `queued_` prefix if you want.

darktohka:
Truth be told, it doesn't matter - we only use queued readers and writers anyway.

rdb (approved):
Great, thank you :)

PR #1282, opened 2022-03-17
Add getter method for positional flag to AudioSound class

### LINKED ISSUE #1257: Suggestion: Add a "getter" for the "positional" flag of the "AudioSound" class.
## Description
Simply put, this issue is to suggest that a "getter" method be added to the "AudioSound" class in order to read the value of the "positional" flag.

## Use Case
There may be cases in which one wants to know whether or not an extant "AudioSound" object is set up for use with the 3D-positional system.

Indeed, a sound that is so set up may misbehave when treated as a "2D" sound, while a sound that is _not_ so set up may not respond to attempts to position it in space, if I'm not much mistaken.

(This is _not_ urgent in the case of myself, the original poster of this issue.)

morganmliang:
## Issue description
 Fixes #1257 

## Checklist

### DIFF
--- a/panda/src/audio/audioSound.h
+++ b/panda/src/audio/audioSound.h
@@ -84,6 +84,9 @@ PUBLISHED:
   // There is no set_name(), this is intentional.
   virtual const std::string& get_name() const = 0;
 
+
+  virtual const bool get_positional() const = 0;
+
   // return: playing time in seconds.
   virtual PN_stdfloat length() const = 0;
 
@@ -130,6 +133,8 @@ PUBLISHED:
   MAKE_PROPERTY(play_rate, get_play_rate, set_play_rate);
   MAKE_PROPERTY(active, get_active, set_active);
   MAKE_PROPERTY(name, get_name);
+  MAKE_PROPERTY(positional, get_positional);
+
 
 protected:
   AudioSound();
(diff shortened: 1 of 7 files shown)


panda/src/audio/audioSound.h
```diff
@@ -84,6 +84,9 @@ class EXPCL_PANDA_AUDIO AudioSound : public TypedReferenceCount {
   // There is no set_name(), this is intentional.
   virtual const std::string& get_name() const = 0;
 
+
+  virtual const bool get_positional() const = 0;
```
rdb, line 88: The first `const` has no effect here; you should leave it out

panda/src/audiotraits/fmodAudioSound.cxx
```diff
@@ -538,6 +540,11 @@ get_name() const {
   return _file_name;
 }
 
+const bool& FmodAudioSound::
+get_positional() const {
```
rdb, line 544: The return type doesn't match the base class method it's overriding; this is returning `const bool &`, but the base class method returns `const bool`.  But I think just returning `bool` would be best.

panda/src/audiotraits/openalAudioSound.cxx
```diff
@@ -875,6 +875,14 @@ get_name() const {
   return _basename;
 }
 
+/*
+  Get positional flag
+*/
```
rdb, line 880: Formatting:
```cpp
/**
 * Get positional flag
 */
```

rdb (changes requested):
Thanks!  Some inline comments.

rdb (changes requested):
Feedback has not yet been fully addressed

PR #1299, opened 2022-05-03
Restore webcam functionality on Linux

geraldog:
## Issue description
Webcam support is broken on Linux. mmap() for the buffers always returns MAP_FAILED.

## Solution description
mmap() takes into account the mode flags of the underlying file descriptor. After auditing the code it turns out the O_RDWR mode flag is being overwritten with O_NONBLOCK, when instead it should be OR'ed.

Fixing this restores webcam functionality.

## Checklist


rdb (approved):
Thanks!

geraldog:
My pleasure. Thank you and everyone else for making Panda3D great!

PR #1308, opened 2022-05-21
cocoadisplay: Add support for high-dpi screens

el-dee:
## Issue description

This PR add High-DPI support on macOS. Until now, on macOS a Panda3D application was not "dpi-aware" and so the OpenGL view was upscaled by the system. With this PR, an application can set `dpi-aware` to `true` to disable the scaling on High-DPI/Retina screen and use the full resolution of the display.

See also the discussion here : https://discourse.panda3d.org/t/updated-there-is-no-retina-support-on-mac-catalina-big-sur/27946

## Solution description

On macOS, an application should not be aware of the actual pixel density of a screen, the application uses display points that are then transformed into actual pixel coordinates by appkit. Pixel units (or more exactly backing store points), are to be used only for some specific case. On the other hand, in the philosophy of the High-DPI support of Panda3D the unit used is always display pixels. Therefore, with this PR, the size of the window and view are converted to display points when modified by the application, and converted from display point into pixels when updated by the system.

 NSApp, NSWindow and MSView can have their own value for the `NSHighResolutionCapable` flag, which control the automatic upscaling of the application/window/view.  To have a coherent behaviour between an application that is launched using the Python interpreter (which has this flag set to `True` since macOS 10.15) and an application built using deploy_ng, and also to allow an application to programmatically decide to support high dpi or not, this PR set the `NSHighResolutionCapable` flag to `True` in the `Info.plist` generated by deploy_ng. Therefore the application and the window will always be dpi aware. The `NSHighResolutionCapable` flag of the view is instead controlled by the `dpi-aware` configuration variable. 

As usual, there are a lot of quirks, weird behaviours and undocumented features in AppKit that I had to work around (I documented their occurence in the code).
There are two important modifications to be noted here : 

first, `CGLSetFullScreenOnDisplay()` seems to not support high-dpi mode (at least I could not find to make it work) and I had to remove its usage, otherwise the fullscreen view was always upscaled. As a consequence I also had to remove call to `CGDisplayCapture()`

Also, in the events fired by `makeFirstResponder()` the internal state of the NSView was wrong and lead to invalid size of position. That's why I forced a manual call to `handle_resize_event()` after each occurence.

Finally, on macOS the backing scale factor (used as `detected_display_zoom` in Panda3D) can be different between displays. On the other hand, in Panda3D there is only one value tied to the graphic pipe. That means on a hybrid setup, a window will be upscaled when moved from a low-dpi screen to a high-dpi screen even if the `dpi-aware` flag is set to `True`.

I tested this PR with several low and high dpi displays, in various configurations. I also tested fullscreen switching (from config and from in-app), the various mouse modes, ...

## Checklist
* [ x] …I have familiarized myself with the CONTRIBUTING.md file
* [ x] …this change follows the coding style and design patterns of the codebase
* [ x] …I own the intellectual property rights to this code
* [ x] …the intent of this change is clearly explained
* [ x] …existing uses of the Panda3D API are not broken

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -717,6 +717,7 @@ class build_apps(setuptools.Command):
             'CFBundlePackageType': 'APPL',
             'CFBundleSignature': '', #TODO
             'CFBundleExecutable': self.macos_main_app,
+            'NSHighResolutionCapable': 'True',
         }
 
         icon = self.icon_objects.get(
--- a/panda/src/cocoadisplay/cocoaGraphicsPipe.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsPipe.mm
@@ -40,12 +40,36 @@ CocoaGraphicsPipe(CGDirectDisplayID display) : _display(display) {
   [thread start];
   [thread autorelease];
 
+  // If the application is dpi-aware, iterate over all the screens to find the
+  // one with our display ID and get the backing scale factor to configure the
+  // detected display zoom. Otherwise the detected display zoom keeps its
+  // default value of 1.0
+
+  if (dpi_aware) {
+    NSScreen *screen;
+    NSEnumerator *e = [[NSScreen screens] objectEnumerator];
+    while (screen = (NSScreen *) [e nextObject]) {
+      NSNumber *num = [[screen deviceDescription] objectForKey: @"NSScreenNumber"];
+      if (_display == (CGDirectDisplayID) [num longValue]) {
+        set_detected_display_zoom([screen backingScaleFactor]);
+        if (cocoadisplay_cat.is_debug()) {
+          cocoadisplay_cat.debug()
+            << "Display zoom is " << [screen backingScaleFactor] << "\n";
+        }
+        break;
+      }
+    }
+  }
+
   // We used to also obtain the corresponding NSScreen here, but this causes
   // the application icon to start bouncing, which may be undesirable for
   // apps that will never open a window.
 
-  _display_width = CGDisplayPixelsWide(_display);
-  _display_height = CGDisplayPixelsHigh(_display);
+  // Although the name of these functions mention pixels, they actually return
+  // display points, we use the detected display zoom to transform the values
+  // into pixels.
+  _display_width = CGDisplayPixelsWide(_display) * _detected_display_zoom;
+  _display_height = CGDisplayPixelsHigh(_display) * _detected_display_zoom;
   load_display_information();
 
   if (cocoadisplay_cat.is_debug()) {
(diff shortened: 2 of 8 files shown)


Moguri:
Do we make the PRC variable `cocoa-dpi-aware` until we get better HiDPI support figured out for other platforms?

@rdb do you see any conflicts here with your HiDPI vision? Assuming not, I am pretty tempted to assume the Cocoa code is good (@el-dee is very thorough) to speed up the review process.

el-dee:
> Do we make the PRC variable `cocoa-dpi-aware` until we get better HiDPI support figured out for other platforms?

That may be a good idea, I thought at first to move `dpi-aware` from `win display` into `display` and reuse it, but maybe it's better to keep each platform flag independent for the time being.

> @rdb do you see any conflicts here with your HiDPI vision? Assuming not, I am pretty tempted to assume the Cocoa code is good (@el-dee is very thorough) to speed up the review process.

I tried and sticked close to the Windows HiDPI implementation so an application would behave similarly (within the restriction of Cocoa of course...). But I believe a code review would be welcome as there are some important changes and there might be consequences I have overlooked. Also, when implementing #1321 I realised that I forgot to remove a call to `CGDisplayRelease()`... It was inconsequential and I just force-pushed an update, but that show a review would anyway be useful :)

On the other hand, having the code merged rapidly into master could let other people test it on their setup and check if everything is still working fine, and that the HiDPI mode works for them too.

rdb:
> first, `CGLSetFullScreenOnDisplay()` seems to not support high-dpi mode (at least I could not find to make it work) and I had to remove its usage, otherwise the fullscreen view was always upscaled. As a consequence I also had to remove call to `CGDisplayCapture()`

If I do this on my end, then I end up with an awkward situation when alt-tabbing, because the window stays on top but loses key status or something like that.  Are you noticing anything like this?

rdb:
Maybe what we ought to do is switch to using the "proper" fullscreen mode, ie. this one:
https://developer.apple.com/library/archive/documentation/General/Conceptual/MOSXAppProgrammingGuide/FullScreenApp/FullScreenApp.html

el-dee:
If I remember correctly, that weird behaviour is due to the fact that the window style mask was reset when switching display mode. In the PR I added the setting of the style mask to borderless window . I will confirm when I have access again to my mac

laurentderu:
There is actually still two problems when switching to fullscreen (unrelated to HiDPI but triggered by the removal of `CGLSetFullScreenOnDisplay()`), and removing `CGCaptureDisplay()` was just hiding them.

First, `CGCaptureDisplay()` can not be called on the main display, instead `CGCaptureAllDisplays()` must be used. Second, the window level of the fullscreen window must be set to the shielding level, retrieved using `CGShieldingWindowLevel()`. Instead Panda is using level `NSMainMenuWindowLevel+1` which was equivalent to the shielding level in older version of Cocoa but not anymore, resulting in a black screen.

I will update the PR with the proper fix.

el-dee:
There is actually still two problems when switching to fullscreen (unrelated to HiDPI but triggered by the removal of `CGLSetFullScreenOnDisplay()`), and removing `CGCaptureDisplay()` was just hiding them.

First, `CGCaptureDisplay()` can not be called on the main display, instead `CGCaptureAllDisplays()` must be used. Second, the window level of the fullscreen window must be set to the shielding level, retrieved using `CGShieldingWindowLevel()`. Instead Panda is using level `NSMainMenuWindowLevel+1` which was equivalent to the shielding level in older version of Cocoa but not anymore, resulting in a black screen.

I will update the PR with the proper fix.

rdb:
Great, thank you so much for figuring that out!

el-dee:
Rebased onto the latest coocadisplay changes

rdb (approved):
Thanks!  Works for me, I'm inclined to just push this and see if issues creep up.

Moguri:
I'm good with just merging this. I did a skim over the code and it seems good, but I know nothing about Cocoa.

I think keeping the PRC variable name consistent across platforms is good. I am not sure why a Panda app would want to be dpi-aware on one platform but not another.

Issue #1312, opened 2022-05-22
macOS: Mouse pointer does not move to correct coordinates after window resize


kamgha:
## Description
Given is the following reproducer:

```
import panda3d.core
from direct.showbase.ShowBase import ShowBase


class MyApp(ShowBase):

    def __init__(self):
        ShowBase.__init__(self)
        
        base.accept('1', lambda: self.centerMouse())
        
    def centerMouse(self):
        base.win.movePointer(0, base.win.getProperties().getXSize() // 2, base.win.getProperties().getYSize() // 2)
        
        pointer = base.win.getPointer(0)
        print('DATA:')
        print(f'Window Dimensions: {base.win.getProperties().getXSize()} {base.win.getProperties().getYSize()}')
        print(f'Requested:         {base.win.getProperties().getXSize() // 2} {base.win.getProperties().getYSize() // 2}')
        print(f'Returned Coords:   {pointer.getX()} {pointer.getY()}')
        print()


app = MyApp()
app.run()
```

The code shall center the mouse pointer after pressing '1'. This works on Windows as expected. On macOS, this works as long as the window has not been resized. After resizing the window, e. g. by dragging it bigger at the borders, double clicking on the title bar or clicking on the green icon in the title bar to maximize, the mouse pointer will not center but be positioned on a wrong coordinate.

## Steps to Reproduce

1. Run the sample code under macOS.
2. Press '1' on the keyboard to see the mouse pointer to be centered. Have a look at the output on the terminal.
3. Resize the window, e. g. maximize.
4. Press '1' again on the keyboard. The mouse pointer will not center to the screen, but jump to a wrong position. Have a look at the values on the terminal.

**Before resize:**
```
DATA:
Window Dimensions: 800 600
Requested:         400 300
Returned Coords:   400.0 300.0
```
The returned coordinates match the requested coordinates.

**After maximizing:**
```
DATA:
Window Dimensions: 1680 1050
Requested:         840 525
Returned Coords:   1280.0 750.0
```
The returned coordinates do not match the requested coordinates.

## Environment
 * Operating system: macOS 12.4
 * System architecture: arm64
 * Panda3D version: git master 36e34294cf4b196fddedb219b04857267f42f5c5
 * Installation method: built from source
 * Python version: CPython 3.9.13
 * Compiler (if using C++): Apple clang 13.1.6

el-dee:
Indeed, I could reproduce the problem:

<pre>
DATA:
[...]

PR #1317, opened 2022-05-31
Fix cocoadisplay mouse bugs

el-dee:
## Issue description

This PR aims to solve a couple of issues wrt the mouse behaviour on macOS.

1. Resizing a window may change window origin

When resizing a window, the origin of the window may change, for example when the window is maximized or when the window is resized using the left or top side of the window. See #1312 

2. The mouse Y-coordinates is off by one point.

When retrieving the mouse pointer position using `[event locationInWindow]` the Y-coordinate has 1 for base, however this is converted by `NSView convertPoint:fromView:nil` However in the code that offset is also compensated, resulting in a pointer position off by one point.
This can be tested by moving the pointer to position (16, -1) this point is outside the view and on the window bar and dragging the mouse from that position should move the window alongside. However this is not the case, one must move the position to (16, -2) to be on the first point of the window bar

3. Confined mouse mode is not properly working

When in mouse confined mode, when the cursor is moved against one of the border, it sorts of bounce back inside the window once it reaches the border (the effect is more visible if the mouse is moved fast). This is due to using `[event delta]` when in confined mode instead of the actual mouse pointer position. Deltas are not accurate, and so the derived mouse pointer position calculated by accumulating all the received deltas is wrong.

4. Mouse pointer freezes for half a second after being moved

When the mouse pointer is moved programmatically, either because we are in confined mode or we change the pointer position, the mouse pointer becomes unresponsive for half a second. This is an obscure mechanism inside CG that is triggered when moving (or warping) the mouse pointer.

## Solution description

1. Fix in 6bcf7d1f525892cad35cbab149419076b0d2d3fa consists of calling `handle_move_event()` also when resizing (as was expected by the comment in the code)

2. The fix in 0093bacef423636a52d2d0df61ca38ea6ba8241d simply removes the extraneous coordinates change.

3. The fix in dbc199ce47d35e2fb6baef4d7732e947f137873a uses the delta only for the relative mouse mode.

4. The fix in a61debd7aa28743abba94dd153c087e97619b62b calls `CGAssociateMouseAndMouseCursorPosition()` to disable the event suppresion interval that causes the mouse freeze.

## Checklist

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -143,9 +143,11 @@ move_pointer(int device, int x, int y) {
                           y + _properties.get_y_origin());
     }
 
-    // I don't know what the difference between these two methods is.  if
-    // (CGWarpMouseCursorPosition(point) == kCGErrorSuccess) {
-    if (CGDisplayMoveCursorToPoint(_display, point) == kCGErrorSuccess) {
+    if (CGWarpMouseCursorPosition(point) == kCGErrorSuccess) {
+      //After moving (or warping) the mouse position, CG starts an event
+      // suppression interval during which no more mouse events can occur
+      // This interval can be interupted by the following call :
+      CGAssociateMouseAndMouseCursorPosition(YES);
       // Generate a mouse event.
       NSPoint pos = [_window mouseLocationOutsideOfEventStream];
       NSPoint loc = [_view convertPoint:pos fromView:nil];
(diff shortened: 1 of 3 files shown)


Moguri:
Should this target the 1.10 branch?

Moguri (approved):
Looks good to me, but I know very little about macOS.

I also very much appreciate the thorough description and explanation of solutions.

rdb:
Thanks for your work on this!  We should probably get this into 1.10.

el-dee:
> Thanks for your work! I think I've got all your important fixes in 1.10 now, right?

Thanks :) I think #1321 could be a potential candidate for 1.10 (I the weird exit sequence can be sorted out), everything else remaining should be for 1.11

PR #1321, opened 2022-06-11
cocoadisplay: Restore desktop display mode when closing a fullscreen window

el-dee:
## Issue description

On macOS, when one closes a fullscreen window which display mode is not the desktop display mode, the desktop display mode is not properly restored, resulting in a unexpected change of layout.
## Solution description

This PR should fix this problem by restoring the desktop display mode (stored in `_windowed_mode`) when `handle_close_event()` is called (when the user click the close window button), or when the `open` window property is set to `false` (this happens when the application closes itself).

When the application is requested to close by the system, for example when using `Command-Q`, there is no event or property update received and so the mode is not restored.
To support also this scenario, I updated the application delegate to listen to `applicationWillTerminate` which is called when the application is about to be closed, and I call `remove_all_windows()` on GraphicsEngine. This cause in fine `set_properties_now()` to be called with the `open` property set to `false` like the previous scenario.

Note that I also tried to put the restore code in the destructor of `CocoaGraphicsWindow` but when the application is closed by `Command-Q` the destructor is never called.

## Checklist

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -761,6 +761,19 @@ close_window() {
  */
 void CocoaGraphicsWindow::
 set_properties_now(WindowProperties &properties) {
+
+  // This test is done first as if the application is shutting down, _pipe will
+  // be null and the rest of the method will be skipped.
+  if (properties.has_open() && !properties.get_open()) {
+    if (_windowed_mode != NULL) {
+      // Switch back to the mode we were in when we were still windowed.
+      CGDisplaySetDisplayMode(_display, _windowed_mode, NULL);
+      CGDisplayModeRelease(_windowed_mode);
+      CGDisplayRelease(_display);
+      _windowed_mode = NULL;
+    }
+  }
+
   if (_pipe == (GraphicsPipe *)NULL) {
     // If the pipe is null, we're probably closing down.
     GraphicsWindow::set_properties_now(properties);
@@ -1591,8 +1604,18 @@ handle_close_request() {
  */
 void CocoaGraphicsWindow::
 handle_close_event() {
+  WindowProperties properties;
   if (cocoadisplay_cat.is_debug()) {
     cocoadisplay_cat.debug() << "Window is about to close\n";
+  }
+
+  if (_windowed_mode != NULL) {
+    // Switch back to the mode we were in when we were still windowed.
+    CGDisplaySetDisplayMode(_display, _windowed_mode, NULL);
+    CGDisplayModeRelease(_windowed_mode);
+    CGDisplayRelease(_display);
+    _windowed_mode = NULL;
+    properties.set_fullscreen(false);
   }
 
   _window = nil;
@@ -1629,7 +1652,6 @@ handle_close_event() {
     _cursor = nil;
   }
 
-  WindowProperties properties;
   properties.set_open(false);
   properties.set_cursor_hidden(false);
   system_changed_properties(properties);
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -76,7 +76,7 @@ CocoaGraphicsWindow(GraphicsEngine *engine, GraphicsPipe *pipe,
   if (NSApp == nil) {
     [CocoaPandaApp sharedApplication];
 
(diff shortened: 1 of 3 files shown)


el-dee:
The CI build was failing because there was a missing forward declaration of `GraphicsEngine` in `CocoaPandaAppDelegate.mm`. As with `makepanda.py` everything under `cocoadisplay` is bundled inside `cocoadisplay_composite1.mm` the build was fine on my setup and so it went under the radar. I should switch to the CMake workflow one day :)

el-dee:
@rdb, do you think this could be included in 1.10.12 too ?

rdb:
What happens currently when you press Command+Q?  By what mechanism does the application actually close?  Does Cocoa just call `exit()` somewhere?

el-dee:
I did tests again, and actually the destructor of `CocoaGraphicsWindow` (and also `CocoaGraphicsPipe`)is never called when the application is closed, either using Command-Q, code button, `sys.exit()` or `ShowBase.userExit()`

I don't know the actual flow inside Cocoa when the application is about to close but it seems at first sight the clean up code or the exit hooks are bypassed.

rdb:
I'm a little on the fence about calling `GraphicsEngine::remove_all_windows()`  here.  I feel like it would be more consistent with the other implementations to instead just close the windows associated with the Cocoa app (and possibly handle any close-request events first), since closing the main window will also have the desired effect of calling userExit and friends.

rdb:
Two observations:
* Cocoa *does* close all the windows, but we don't actually implement `windowWillClose` in the window delegate, which is why close_window doesn't run
* It doesn't appear that anything else gets to run after `applicationWillTerminate`, if we want to do any custom shutdown stuff, we should probably implement `applicationShouldTerminate`.

rdb:
@el-dee I've pushed your first fix (calling `remove_all_windows()`) to the release/1.10.x branch and some additional changes so that the finalizeExit handler runs properly on Command+Q.  In fact, applicationWillTerminate will probably never be called now, since I just have applicationShouldTerminate return FALSE always, it just iterates over `[app windows]`, requests each one to be closed, then closes all of them if none of them have a close-request handler.  This should make ShowBase receive window-event and run finalizeExit et al.

I am not actually seeing any issues with fullscreen mode switches, though.  When I hit Command+Q or sys.exit in my test code, the resolution always gets set back properly, without your second change.  In fact, calling `do_fullscreen_switch(nullptr)` inside `close_window()` actually messes up my display mode on quit.

Would you mind doing some testing with the 1.10 branch?  (Buildbot builds will appear [here](https://www.panda3d.org/download.php?platform=macosx&version=devel&sdk&branch=release/1.10.x) shortly if you need them.)

rdb:
Going to un-assign this from 1.10.13 until this is verified to be either still broken or already fixed.

el-dee:
Weird, I will test again using the released 1.10.13 and see when (or if) the mode is not properly restored (perhaps it is triggered by some multi display configuration)

el-dee:
Thanks for the reminder !

I think I have found the actual root cause of the problem, I had mixed things up as I tested several fixes that I thought were independent, hoping to test faster, but actually are tightly coupled in Cocoa...

The problem is indeed still present on the latest V1.10.x but it happens rarely, however you can observe another version of the problem :

If you ask your app to enter a fullscreen mode with a lower resolution than the desktop then quit the app, when the display is leaving the fullscreen mode you can observe that all the other windows have been resized to match the size if the fullscreen mode and are resized back to their previous size.
If you have many windows opened (or heavy task running in the background, like building Panda3D), this resize back is slow and sometimes stops (either because there is a timeout or a bug in Cocoa).

The root cause is that the display is captured by the application after switching the display mode, and this mode switching triggers a resize event for all the opened window, and the reverse resize when the original mode is restored.

If the display is captured before switching mode, only the requesting application receives the resize notification, all the other ones stay as is.

For multi-display setup, this is not enough, though. You have to capture all the displays in order to avoid all the other windows to be resized.

Another issue that is solved by capturing the display before switching mode is that when leaving the fullscreen mode, the window is not repositioned at its former origin and instead stays at (0, 0) but the move event has been triggered by Cocoa, so the window is partly under the menubar of the Finder and there is an offset between the position of the mouse cursor and the position calculated by Panda, see the screenshot below. And to add to the craziness, this does not occur with a multidisplay setup

As this is no longer the topic of this PR i will create a new one with the modifications for capturing the displays (after more testing)


Capture d’écran 2023-01-16 à 21 48 52

el-dee:
Fixed in the linked PR

PR #1323, opened 2022-06-14
support android no thread build

pmp-p:
dummy threading ( -DHAVE_THREADS=No ) does not provide android specifics slots handling jni environnement.


pmp-p:
@rdb if merging that please sync webgl port, cause i use webgl port code tree for both android and wasm.

PR #1333, opened 2022-07-18
direct: Migrate legacy DConfig usage to modern configuration interface

darktohka:
## Issue description

Some modules in `direct` are still using the old DConfig interface (config.GetBool and friends). They should be migrated to the new ConfigVariableBool (and friends) interface.

## Solution description

This PR consists of three commits:
- a46dd2d: which replaces the "value" property accesses to the "getValue()" method call
- 7cb229d: which removes the old `temp-hpr-fix` setting which is no longer used, from the contributed sceneeditor module
- 52d11ca: which replaces the `config.Get*` calls with `ConfigVariable*` usages

## Checklist


rdb:
What's wrong with using the new Pythonic .value property as opposed to the getter method?

darktohka:
It's used inconsistently. Some code in `direct` uses the value property, some code uses the getter method. I personally prefer the getter method, since it doesn't imply that the "value" property can be assigned. I wouldn't be against an inverse change either, however (changing all getter method accesses to the value property).

rdb:
I see no advantage of using `.getValue()` over `.value`.  The latter is faster, less visually cluttered, and naming convention agnostic.  I reckon we will eventually deprecate the former in favour of the latter, so I would suggest the inverse change.

Not sure what you mean by "doesn't imply that it can be assigned".  It doesn't imply such a thing (read-only properties are common), but it *can* be assigned, if you wanted to.

Moguri (changes requested):
Let's leave existing ConfigVariable* getValue(), setValue(), and value calls alone for this PR to reduce clutter (the style update can go in a new PR). For new code, I prefer using the property over the methods.

PR #1337, opened 2022-07-21
dtool: Fix EncryptStreamBuf compatibility with OpenSSL 3.x 

darktohka:
## Issue description

When compiling Panda3D with OpenSSL 3.x, encrypted multifiles completely break. Attempting to create or load an encrypted multifile in Release mode results in a segmentation fault.

On Debug builds, however, some more information is found, as the following assert is hit: https://github.com/panda3d/panda3d/blob/d79709f/dtool/src/prc/encryptStreamBuf.cxx#L247

The `EVP_EncryptInit` call exits with a return code of `0`. According to the [documentation](https://www.openssl.org/docs/man3.0/man3/EVP_EncryptInit.html):

```
EVP_EncryptInit_ex(), EVP_EncryptUpdate() and EVP_EncryptFinal_ex() return 1 for success and 0 for failure.

EVP_EncryptInit(), EVP_DecryptInit() and EVP_CipherInit() behave in a similar way to EVP_EncryptInit_ex(), EVP_DecryptInit_ex() and EVP_CipherInit_ex() except they always use the default cipher implementation.
```

This means that the `EVP_EncryptInit` call failed. But why is that? The cipher is not nullptr. Panda3D does have a "valid" reference to the cipher!

Two reasons:
1. First of all, the `bf-cbc` cipher has been removed from the base OpenSSL provider. It has been moved to the legacy OpenSSL provider, which must be compiled separately from the base OpenSSL libraries: https://www.openssl.org/docs/man3.0/man7/OSSL_PROVIDER-legacy.html. This legacy provider is still shipped with some distributions, but it is NOT loaded by default and we cannot rely on its existence (`nolegacy` OpenSSL build flag).
2. Secondly, the `EVP_EncryptInit` method has been deprecated. Which shouldn't *really* be a problem, given that deprecation is not deletion, but... According to the OpenSSL 3.x documentation, the following applies to `EVP_get_cipherbyname()` and `EVP_get_cipherbynid()`, which Panda3D uses to fetch its ciphers:

```
The EVP_get_cipherbyname() function is present for backwards compatibility with OpenSSL prior to version 3 and is different to the EVP_CIPHER_fetch() function since it does not attempt to "fetch" an implementation of the cipher. Additionally, it only knows about ciphers that are built-in to OpenSSL and have an associated NID. Similarly EVP_get_cipherbynid() and EVP_get_cipherbyobj() also return objects without an associated implementation.
```

Huh. How nice that OpenSSL is breaking compatibility and giving users pointers to ciphers that seem valid but have no underlying implementation, leading to the encrypt/decrypt inits failing!

## Solution description

We have two problems to solve:
1. We need to use the new cipher fetching methods when building with OpenSSL 3.x. These will allow Panda to fetch actual ciphers with actual implementations.
2. We need to switch the default encryption algorithm from Blowfish to something else. Blowfish's days are numbered in OpenSSL. I propose AES-256-CBC as a replacement, as it has hardware support in many targets and has proven to be a solid choice in other applications. It is also supported in OpenSSL 1.0.2 (I haven't checked in earlier library versions than 1.0.2). We must also make sure to keep compatibility with older Blowfish streams, so make sure to load the legacy provider on a best-effort basis. If the legacy provider fails to load, it's not that big of a problem.

I submit the following commits:
1. c9cf803914e2700897cbfc73394e865055e52285 - This commit replaces the `EVP_get_cipherbynid` and `EVP_get_cipherbyname` methods with `EVP_CIPHER_fetch` on OpenSSL 3.x. It also switches from the deprecated `EVP_DecryptInit` to `EVP_DecryptInit_ex` (and EncryptInit as well). Also, the IV and key are now updated simultaneously with one method call rather than updating the IV and the key with two different calls.
2. bdfe1398babcbdee3d77b7d1691dd1c2012b7bea - This commit switches the default encryption algorithm from `bf-cbc` to `aes-256-cbc`.
3. aff77e37e707885912614027521acf00bc93bb8b - This commit makes Panda3D attempt to load the legacy provider on a best-efforts basis.

## Checklist


rdb:
Thanks for your work on this.

How much of this can/should go into 1.10?

rdb (changes requested):
Needs to be rebased (or closed if my previous changes were sufficient, but I didn't make the EVP_DecryptInit_ex change)

rdb:
Friendly ping. :-)

Moguri:
Closing due to inactivity. Please feel free to reopen the PR  if/when you continue working on this.

PR #1339, opened 2022-07-26
Update README.md

jakemcf22:
## Issue Description
The usage of thirdparty vs third-party is inconsistent throughout the readme.
## Solution
Modified third-party vs thirdparty so that the usage in the paragraph is consistent with the links provided in the same paragraph.


Moguri (approved):
Looks good to me

Moguri (changes requested):
Actually, on second thought, I believe we should prefer "third-party"

rdb:
Ah, well, too late `¯\_(ツ)_/¯`

I suggest leaving it like this, I think minor stylistic changes to readme files shouldn't be a priority

Moguri:
Yeah, not something to worry about. 😁

PR #1340, opened 2022-07-26
ffmpeg: Resolve segmentation fault when statically linking ffmpeg

darktohka:
## Issue description

When statically linking the ffmpeg module, a nasty segmentation fault can occur at: 
https://github.com/panda3d/panda3d/blob/0fcfb8e372c71e5c15a20bfd064b51f8db831469/panda/src/ffmpeg/ffmpegAudio.h#L42

When Panda3D tries to register the `FfmpegAudio` type handle, it fails to evaluate the parent class `MovieAudio` as `FfmpegAudio` is registered before `MovieAudio` is registered. As such, a segmentation fault happens when accessing the parent class. This is hard to track down as it happens before the `main` method of the program is ever called.

This is due to a copy-paste gone wrong. `FfmpegVideo` uses `TypedWritableReferenceCount` as its parent class, and as such `FfmpegVideo` initializes `TypedWritableReferenceCount` before `FfmpegVideo` itself. `FfmpegAudio`, however, depends on `MovieAudio`. Most likely `FfmpegAudio` was copied from the `FfmpegVideo` class, and the initialization of the parent class was not changed.

## Solution description

Initializing `MovieAudio` before initializing `FfmpegAudio` solves the problem and prevents the segmentation fault from happening.

## Checklist

### DIFF
--- a/panda/src/ffmpeg/ffmpegAudio.h
+++ b/panda/src/ffmpeg/ffmpegAudio.h
@@ -38,7 +38,7 @@ public:
     return _type_handle;
   }
   static void init_type() {
-    TypedWritableReferenceCount::init_type();
+    MovieAudio::init_type();
     register_type(_type_handle, "FfmpegAudio",
                   MovieAudio::get_class_type());
   }


rdb (approved):
Thanks for tracking this down! Will merge onto 1.10 branch.

PR #1347, opened 2022-08-04
audio: Add support for setting the loop start time of sounds

darktohka:
## Issue description
It would be nice to be able to set the loop start time of certain sounds, like music.

This would mean that the audio being played would play, from start to finish, the first time around, and when it begins looping, it wouldn't begin from the beginning, but rather from a manually set timestamp.

This feature would be very nice to have for songs that have different opening tunes that would need to be skipped when replaying the same song.

## Solution description
We add two new methods to the AudioSound class:
`set_loop_start`, which can be used to set the timestamp from which sound will resume when looping
`get_loop_start`, which can be used to get the timestamp from which sound will resume when looping

Since we have three audio backends right now (NULL, OpenAL and FMOD), three backends have to be edited to add support for setting loop start times.

The first is the NULL audio backend, which doesn't require any special explanation.

The second is the OpenAL audio backend. I've edited the code to begin playing the audio from the set `loop_start` timestamp, as well as to queue the buffer from the set timestamp when necessary.

The third is the FMOD audio backend. I'm aware that the FMOD audio backend is currently being reworked. This implementation is for the currently checked in FMOD audio backend, albeit I checked the documentation for FMODEX and it seems like the functions used exist in FMODEX aswell. The FMOD library provides the function `setLoopPoints`, which allows us to set a loop start timestamp and a loop end timestamp. To provide support for loop start timestamps, this implementation sets a loop point from the loop start timestamp to the end of the audio.

The loop start time is always checked as to not be greater than the actual length of the audio.

## Checklist

### DIFF
--- a/panda/src/audio/audioSound.h
+++ b/panda/src/audio/audioSound.h
@@ -42,6 +42,10 @@ PUBLISHED:
   virtual void set_loop_count(unsigned long loop_count=1) = 0;
   virtual unsigned long get_loop_count() const = 0;
 
+  // loop_start: 0 = beginning.  expressed in seconds.  inits to 0.
+  virtual void set_loop_start(PN_stdfloat loop_start=0) = 0;
+  virtual PN_stdfloat get_loop_start() const = 0;
+
   /**
    * Control time position within the sound, in seconds.  This is similar (in
    * concept) to the seek position within a file.  The value starts at 0.0 (the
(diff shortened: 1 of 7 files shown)


RegDogg (approved):
This would probably make things a bit easier, instead of using multiple SoundIntervals, since running 2 of the same Audio files at the same time is a bit janky

Moguri (approved):
Looks good to me.

rdb (approved):
I decided to accept this PR as-is, a new unified loop API can be added in a separate change.

Thank you!

PR #1365, opened 2022-09-02
direct: Use explicitly Vec3F in calls to addData3f to avoid crash on double precision builds

el-dee:
## Issue description

When opening the BufferViewer with Panda3D compiled with double precision float enabled, the app crashes as `Vec3` is actually `Vec3D` and `addData3f()` expect `Vec3F`

## Solution description

Trivial fix that replace `Vec3` with `Vec3F`

## Checklist


rdb:
Oops, the timing is unfortunate, since I had already reached the point of no return for the 1.10.12 release at the time you filed this PR.  It will have to go into 1.10.13, with my apologies for the regression.

Thanks!

rdb (approved):

PR #1367, opened 2022-09-08
Added e2k support (MCST Elbrus 2000)

DJs3000:
This pull request introduces support of Elbrus hardware platform. e2k (Elbrus 2000) - this is VLIW/EPIC architecture, like Intel Itanium architecture.

About this architecture:
https://en.wikipedia.org/wiki/Elbrus_2000


rdb (approved):
Thank you!

PR #1380, opened 2022-10-21
dcparser: fix uint32s being signed on Windows

NathaanTFM:
The `long` type is 32 bits on Windows and 64 bits on most linux systems. Therefore, passing an uint32 value to `PyLong_FromLong` on Windows converts it to a signed 32 bit integer.

Replacing `PyLong_FromLong` to `PyLong_FromUnsignedLong` should fix the issue


rdb (approved):

rdb:
Committed and backported to 1.10, thanks!

Issue #1412, opened 2022-12-23
Several missing HAVE_PYTHON blocks


Frangwhuf:
## Description
Several parts of the published API reverence Python types, but are not in `#ifdef HAVE_PYTHON` blocks. Typically this is with `EXTENSION` methods. For example, instances of `__reduce__`.

## Use Case
I am teaching `interrogate` how to emit JNI bindings. Python types, as part of the deep Python integration in PANDA3d, are not meaningful in this context. It is just as reasonable to compile without Python as it will be to compile without JNI. As such I am cleaning up sites where this has leaked over time.

I have a PR that I will be filing today that addresses many of these. I'll file more as I go along. As such I don't think this needs to be assigned to anyone else. But if someone feels super passionate about it, they are welcome to take what I have done so far.

PR #1413, opened 2022-12-23
add more HAVE_PYTHON

Frangwhuf:
(Also add some more #endif comments. Where it doesn't over clutter the change I try to leave things in a better state than I found them.)

## Issue description
There are missing HAVE_PYTHON blocks in the code. See [Issue #1412](https://github.com/panda3d/panda3d/issues/1412).

## Solution description
Add HAVE_PYTHON `#ifdef`s where needed. There will be more down the road, but this was already touching enough that I felt it was a good time to checkpoint.

## Checklist

### DIFF
--- a/dtool/src/dtoolbase/typeHandle.h
+++ b/dtool/src/dtoolbase/typeHandle.h
@@ -97,7 +97,9 @@ PUBLISHED:
   // its value, it  might happen after the value had already been set
   // previously by another static initializer!
 
+#ifdef HAVE_PYTHON
   EXTENSION(static TypeHandle make(PyTypeObject *classobj));
+#endif
 
   INLINE bool operator == (const TypeHandle &other) const;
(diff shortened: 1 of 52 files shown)


rdb (approved):
I'm perfectly happy with this solution, although it does make me think it might be convenient to create a specific `PY_EXTENSION` macro that automatically expands down to nothing if HAVE_PYTHON is not set, or expands to EXTENSION otherwise.

Frangwhuf:
> 

Yeah. I was considering that, but the optimist in me keeps thinking that I must be just about done. :)

I'll give that a spin over the break and if I like how it looks I'll fire it off.

rdb:
If you wish; I've merged this as-is, we can do a PY_EXTENSION change in a later cleanup PR if desirable.

Thanks!

Issue #1414, opened 2023-01-02
Create PY_* variants of EXTENSION and EXTEND that auto obliterate when !HAVE_PYTHON


Frangwhuf:
## Description
To reduce `ifdef` chaff in the source, PY_ variants of the EXTENSION macro would both streamline and help document the intent of a given extension.

## Use Case
The majority of EXTENSIONs in the PANDA3d code are specifically for the Python integration. While this is the primary (and most fully expressed) scripting integration, the intent is that it does not have to be the only one. As such it should remain reasonable to compile without Python support/integration. Many of the existing EXTENSION instances reference Python types without being in `ifdef HAVE_PYTHON` wrappers. Creating a Python specific variant of these macros should make this both clearer and safer.

## Note
I have a PR that does this, which I will stage shortly. As such this can be assigned to me.

PR #1415, opened 2023-01-02
create PY_EXTENSION and friends

Frangwhuf:
## Issue description
Reduce some `ifdef` chaff and make the scripting target for a Python extension clearer. See [Issue #1414](https://github.com/panda3d/panda3d/issues/1414)

## Solution description
Created `PY_EXTENSION`, `PY_EXTEND`, `PY_MAKE_PROPERTY`, and `PY_MAKE_SEQ_PROPERTY` (for these last two, they were the only ones currently needed). These mimic the behavior of the non-PY versions with the added condition that if HAVE_PYTHON is not defined, they drop the contents as would normally happen with CPPPARSER undefined. Further, the `PY_EXTEND` macro is expressed as variadic (unlike `EXTEND`) so that it can consume its contents properly. (It may be worth considering doing this with `EXTEND` as well.)

## Checklist

### DIFF
--- a/panda/src/gobj/geomVertexArrayData.h
+++ b/panda/src/gobj/geomVertexArrayData.h
@@ -114,11 +114,9 @@ PUBLISHED:
   static void lru_epoch();
   INLINE static VertexDataBook &get_book();
 
-#ifdef HAVE_PYTHON
-  EXTENSION(int __getbuffer__(PyObject *self, Py_buffer *view, int flags));
-  EXTENSION(int __getbuffer__(PyObject *self, Py_buffer *view, int flags) const);
-  EXTENSION(void __releasebuffer__(PyObject *self, Py_buffer *view) const);
-#endif // HAVE_PYTHON
+  PY_EXTENSION(int __getbuffer__(PyObject *self, Py_buffer *view, int flags));
+  PY_EXTENSION(int __getbuffer__(PyObject *self, Py_buffer *view, int flags) const);
+  PY_EXTENSION(void __releasebuffer__(PyObject *self, Py_buffer *view) const);
 
 public:
   virtual void evict_lru();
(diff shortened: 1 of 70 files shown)


panda/src/gobj/geomVertexArrayData.h
```diff
@@ -310,14 +308,12 @@ class EXPCL_PANDA_GOBJ GeomVertexArrayDataHandle : public ReferenceCount, public
...
-#ifdef HAVE_PYTHON
-  EXTENSION(void copy_data_from(PyObject *buffer));
-  EXTENSION(void copy_subdata_from(size_t to_start, size_t to_size,
+  PY_EXTENSION(void copy_data_from(PyObject *buffer));
+  PY_EXTENSION(void copy_subdata_from(size_t to_start, size_t to_size,
                                    PyObject *buffer));
-  EXTENSION(void copy_subdata_from(size_t to_start, size_t to_size,
+  PY_EXTENSION(void copy_subdata_from(size_t to_start, size_t to_size,
                                    PyObject *buffer,
                                    size_t from_start, size_t from_size));
```
rdb, line 316: Nitpick: alignment is wrong now

panda/src/gobj/texture.h
```diff
@@ -461,9 +461,9 @@ class EXPCL_PANDA_GOBJ Texture : public TypedWritableReferenceCount, public Nama
                             size_t page_size = 0);
   void set_ram_image_as(CPTA_uchar image, const std::string &provided_format);
 #else // !CPPPARSER || !HAVE_PYTHON
-  EXTEND void set_ram_image(PyObject *image, CompressionMode compression = CM_off,
-                            size_t page_size = 0);
-  EXTEND void set_ram_image_as(PyObject *image, const std::string &provided_format);
+  PY_EXTEND(void set_ram_image(PyObject *image, CompressionMode compression = CM_off,
+                            size_t page_size = 0));
```
rdb, line 465: Alignment

panda/src/ode/odeUtil.h
```diff
@@ -43,7 +43,7 @@ class EXPCL_PANDAODE OdeUtil {
   static PT(OdeCollisionEntry) collide(const OdeGeom &geom1, const OdeGeom &geom2,
                                       const short int max_contacts = 150);
 
-  EXTENSION(static int collide2(const OdeGeom &geom1, const OdeGeom &geom2,
+  PY_EXTENSION(static int collide2(const OdeGeom &geom1, const OdeGeom &geom2,
                                 PyObject* arg, PyObject* callback));
```
rdb, line 47: Alignment

rdb (changes requested):
Great, much better.  Thank you!  Just a few cases where the alignment is off now.

It's not really a big deal here, but for future reference, I would prefer to put unrelated cleanup changes (such as `#endif` comments) in a separate commit, so that it's easier to sort out the important parts of a change in a diff - also easier to untangle things if a change ever needs to be reverted.

Frangwhuf:
Two questions:
 
- Is the preferred mechanism for resolving the above a separate commit on the PR, or to squash the changes together?
- For future janitorial updates (endif comments, etc) is the preferred approach to make that a separate PR or just a separate commit in the PR?

rdb:
It takes only a single click for a maintainer to squash the commits in a PR, so I'd say, whatever is easiest for you. :)

As for the (uncontroversial) janitorial updates, since you are a Panda dev I think you should just push them directly to the repo.  In the general case, a separate PR would be cleanest if the changes are unrelated, so that they can be reviewed and applied separately.  That said, I would consider a separate commit on the same PR also OK if the changes are interwoven such that dealing with the merge conflicts would just be a PITA.

Issue #1418, opened 2023-01-03
Remove covariant overload on SocketStream::flush


Frangwhuf:
## Description
SSWriter declares a covariant overload for `flush()` that returns `bool`. This is a departure from the base class `ostream` which declares the return type as `void`. Such covariant return overloads can be highly problematic in C++ as they behave in unexpected/surprising ways. The semantics of the `bool` returned from `SSWriter::flush()` is the same as calling `is_closed()`, and there is no deeper atomicity at play. As such `SSWriter::flush()` can be changed to return 'void', and usage sites that depended on the return value can call `is_closed()`.

## Use Case
This was discovered as part of the development of the JNI bindings for PANDA3d. Java does not permit such covariant overloads.

## Notes
I have a PR that addresses this change, which I will push shortly. As such this issue can be assigned to me.

PR #1419, opened 2023-01-03
fix SocketStream::flush() covariant

Frangwhuf:
## Issue description
This corrects an instance of covariant overload on return type. Such functionality is generally a dangerous pattern in C++. See [Issue #1418](https://github.com/panda3d/panda3d/issues/1418)

## Solution description
I changed `SSWriter::flush()` (and derived types) to return `void` rather than `bool`, then adjusted call sites to call `is_closed()` if they needed that information.

## Checklist


rdb (approved):
As discussed on the forums.  LGTM.

PR #1428, opened 2023-01-07
webgl-port: merge test

pmp-p:
CI merge test


pmp-p:
one linking problem with ubuntu-18.04 on python3.11 - which i cannot have with em++ when linking because i remove duplicates before calling clang - is worth checking

pmp-p:
The macos failure seems unrelated to the PR, `clang: error: unable to execute command: Segmentation fault: 11` could be OOM

Moguri:
Should this be closed for now, or, maybe switched to draft?

rdb:
Closing, will revisit later, I'm still sitting on some local changes for webgl-port CI, and then we can think about merging it into master

PR #1430, opened 2023-01-08
distributed: Fix segmentation fault when ConnectionRepository is verbose by holding GIL

darktohka:
## Issue description

A segmentation fault occurs when `CConnectionRepository` is set to verbose mode, and a message is to be printed to the debug console.

This occurs because `CConnectionRepository` is running on a separate thread when threading is enabled, and Python might be running code on another thread while `describe_message` is running. As such, we need to hold the GIL before calling into any Python functions from the `describe_message` function.

## Solution description
I solved the problem by holding the GIL before calling into any Python functions in the `describe_message` function.

I've also changed `msg_type` to be an unsigned int, as it's read from the stream as an `uint16` (not an `int16`), and removed an unused Python 2.2 macro.

## Checklist

### DIFF
--- a/direct/src/distributed/cConnectionRepository.cxx
+++ b/direct/src/distributed/cConnectionRepository.cxx
@@ -687,11 +687,7 @@ handle_update_field() {
       PyObject_GetAttrString(_python_repository, "doId2do");
     nassertr(doId2do != nullptr, false);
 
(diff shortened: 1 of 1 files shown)


rdb (approved):
LGTM for 1.10.14, thanks!

Issue #1269, opened 2022-02-22
Can not use ShowBase.destroy


tzaika:
## Description

I'm trying to call the ShowBase method "destroy". I want to call this method on a certain event in the app. Of course, the listener for this event runs in a separate thread. But when I call the "destroy" method from this listener, I get the following exception:

```
2022-02-22 19:42:39.061 python[4344:205896] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'nextEventMatchingMask should only be called from the Main Thread!'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007ff80dbfff0b __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007ff80d960b9d objc_exception_throw + 48
	2   AppKit                              0x00007ff8105b1116 -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 4773
	3   libpandagl.dylib                    0x00000001118821ac _ZN19CocoaGraphicsWindow14process_eventsEv + 92
	4   libpandagl.dylib                    0x0000000111884594 _ZN19CocoaGraphicsWindow12close_windowEv + 292
	5   libpanda.1.10.dylib                 0x000000010f214392 _ZN14GraphicsWindow18set_properties_nowER16WindowProperties + 530
	6   libpandagl.dylib                    0x0000000111884664 _ZN19CocoaGraphicsWindow18set_properties_nowER16WindowProperties + 52
	7   libpanda.1.10.dylib                 0x000000010f213e89 _ZN14GraphicsWindow13set_close_nowEv + 281
	8   libpanda.1.10.dylib                 0x000000010f1ee66c _ZN14GraphicsEngine14WindowRenderer10do_pendingEPS_P6Thread + 316
	9   libpanda.1.10.dylib                 0x000000010f1ee0ba _ZN14GraphicsEngine16do_remove_windowEP14GraphicsOutputP6Thread + 314
	10  libpanda.1.10.dylib                 0x000000010f1ebc25 _ZN14GraphicsEngine18remove_all_windowsEv + 277
	11  core.cpython-38-darwin.so           0x0000000110b85dba _ZL43Dtool_GraphicsEngine_remove_all_windows_508P7_objectS0_ + 58
	12  Python3                             0x000000010ebaea12 method_vectorcall_NOARGS + 242
	13  Python3                             0x000000010ec85012 call_function + 354
[...]

PR #1436, opened 2023-01-12
ShowBase: Add a check for Destroy() being called from a non-main thread

DavidACrompton:
## Issue description
On certain platforms (MacOS in this case), closing a graphics window must be done from the main thread. Failing to do so on MacOS triggers an NSInternalInconsistencyException. 
Issue: #1269 

## Solution description
This PR solves the problem by checking if Destroy is being called from the main thread. Issuing a warning via DirectNotify if it hasn't been called from the main thread. Then posts a task in TaskMgr to call Destroy. 

## Checklist

### DIFF
--- a/direct/src/showbase/ShowBase.py
+++ b/direct/src/showbase/ShowBase.py
@@ -584,11 +584,14 @@ class ShowBase(DirectObject.DirectObject):
         exitfunc and will be called at application exit time
         automatically.
 
-        This function is designed to be safe to call multiple times."""
+        This function is designed to be safe to call multiple times.
+
+        When called from a Thread, a Task is created to run it on the main thread
+        """
 
         if threading.current_thread() is not threading.main_thread():
-            self.notify.warning("destory() was called on a thread other than the main thread.")
-            taskMgr.doMethodLater(0.1, self.destroy, "ShowBaseDestroyTask", extraArgs=[])
+            task = taskMgr.add(self.destroy, extraArgs=[])
+            task.wait()
             return
 
         for cb in self.finalExitCallbacks[:]:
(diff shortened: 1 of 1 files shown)


rdb (changes requested):
Thanks for the PR.

* Please test that it's legal to call `destroy()` within a task.  Since it shuts down the task manager, I am not sure if there will be strange behaviour.
* If it works correctly with a task then there's no need for the warning, is there? :)
* I see no reason for there to be a 100 ms delay.  I suggest just using `add()`.
* The destroy operation should still appear synchronous, so we should wait for the task to be completed.  This can be done by calling `result()` on the task object returned from `add()`, which will block the current thread while waiting for the main thread to finish the task.
* There should be a note in the docstring about what happens when not calling this from the main thread.

It also just occurred to me that people might not be calling `run()` from the main thread either, so there's no guarantee that it does indeed happen on the main thread.  But, I'm not sure what actually happens when you try to do that (it's possible that this just doesn't work anyway).

If posting a task to the main thread turns out to be too tricky or doesn't work properly, I'd be fine with just displaying an error instead.

DavidACrompton:
> * If it works correctly with a task then there's no need for the warning, is there? :)

Point taken!

> * Please test that it's legal to call `destroy()` within a task.  Since it shuts down the task manager, I am not sure if there will be strange behaviour.
> * I see no reason for there to be a 100 ms delay.  I suggest just using `add()`.
> * The destroy operation should still appear synchronous, so we should wait for the task to be completed.  This can be done by calling `result()` on the task object returned from `add()`, which will block the current thread while waiting for the main thread to finish the task.

I tested destroy() using both taskMgr.add and do_method_later. Using add() with result() caused a SystemError exception. 
```SystemError: <built-in method result of panda3d.core.PythonTask object at 0x114a895d0> returned NULL without setting an error```
I've just now tried add() and wait() which also works - that's probably the best way to do it.
 In both cases, the window gets destroyed and seems to act the same as calling destroy() in the main-thread

> It also just occurred to me that people might not be calling `run()` from the main thread either, so there's no guarantee that it does indeed happen on the main thread. But, I'm not sure what actually happens when you try to do that (it's possible that this just doesn't work anyway).

You can't run() from a non-main thread. Task.py uses signal, which has to be used on the main thread.
```ValueError: signal only works in main thread of the main interpreter```
Do you think I should add a check in run()? In run() I'm not sure what we can do except a notify.error? I believe taskMgr hasn't kicked off yet to be able to use it to call it?

Issue #1203, opened 2021-10-31
Unexpected playing of Func objects when resuming finished Sequence


Epihaius:
## Description
Hi,

Normally, I wouldn't expect a call to `Sequence.resume` to have any effect on a Sequence that has played to the end (assuming it is not looping).
Yet, it appears that any `Func` objects added to a `Sequence` get replayed upon resuming that Sequence, even after the latter has completely finished playing.

## Steps to Reproduce
Here is a code sample to demonstrate this:

```python
from direct.showbase.ShowBase import ShowBase
from panda3d.core import *
from direct.interval.IntervalGlobal import *


base = ShowBase()

base.disable_mouse()

base.camera.set_pos(5., -5., 7.)
base.camera.look_at(0., 5., 0.)

def say_hi():
    print("Hi!")

def say_hi2():
    print("Hi again!")

smiley = base.loader.load_model("smiley")
smiley.reparent_to(base.render)
ival1 = LerpPosInterval(smiley, 2., (0., 10., 0.))
ival2 = LerpPosInterval(smiley, 2., (0., 0., 0.))
seq = Sequence()
seq.append(ival1)
seq.append(Func(say_hi))
seq.append(ival2)
seq.append(Func(say_hi2))
seq.start()
#seq.finish()

def pause():
    seq.pause()

def resume():
    seq.resume()

base.accept("-", pause)
base.accept("+", resume)


base.run()
```

When running the sample, wait until the smiley stops moving and then press `+` to resume the Sequence. Both callables wrapped by Funcs are called each time the Sequence is resumed, even though it has already run its course.

This seems quite strange to me, especially since any other intervals are *not* repeated.

## Environment
 * Operating system: Windows 10 Home
 * System architecture: 64-bit
 * Panda3D version: 1.10.10
 * Python version: 3.7.9

rdb:
Is the problem resolved if you use `Func(f, openEnded=True)` ?

Epihaius:
No, sadly that doesn't seem to make any difference.

However, when I set `openEnded=False`, *then* the callable wrapped by `Func` doesn't get called anymore when resuming. But then it also *never* gets called if it's not followed by any "non-instantaneous" interval in the `Sequence`, so it's not a real solution either.
Unless I append one final, very short "dummy" interval -- e.g. `Wait(0.1)` -- to the `Sequence`, but that feels quite hacky. And if the duration of said dummy interval is really short (e.g. `0.01`) then it doesn't even work.

[...]

PR #1438, opened 2023-01-12
MetaInterval: Prevent restarting an interval if it has already finished and we don't specify a start time

JaceMayer:
## Issue description
As per issue #1203, when a finished sequence has ```resume()``` called, ```Func()``` have their callbacks called. This PR aims to mitigate this unintuitive design decision.

## Solution description
For MetaIntervals, if you call ```.resume()``` and the interval is already in the finished state, it will return out of the function, unless you have specified a ```startT```.  For consistency, this check has been placed in MetaInterval as opposed to overriding the function for just ```Sequence()```

## Checklist

### DIFF
--- a/direct/src/interval/FunctionInterval.py
+++ b/direct/src/interval/FunctionInterval.py
@@ -86,11 +86,18 @@ class FunctionInterval(Interval.Interval):
         return name
 
     def privInstant(self):
+        """
+        In versions previous to 1.11, this function did not set the state of the Interval
+        In 1.11 this has been changed to be consistent with MetaInterval
+        """
+        self.state = CInterval.SStarted
         # Evaluate the function
         self.function(*self.extraArgs, **self.kw)
         # Print debug information
         self.notify.debug(
             'updateFunc() - %s: executing Function' % self.name)
+        self.state = CInterval.SFinal
+        self.intervalDone()
 
 
 ### FunctionInterval subclass for throwing events ###
--- a/direct/src/interval/Interval.py
+++ b/direct/src/interval/Interval.py
@@ -134,6 +134,12 @@ class Interval(DirectObject):
         return self.getT()
 
     def resume(self, startT = None):
+        """
+        In versions previous to 1.11, this would resume an interval even if it had ended
+        As of 1.11 we now check if the interval has ended, and only resume it if startT has been passed
+        """
+        if self.getState() == CInterval.SFinal and startT is None:
+            return
         if startT is not None:
             self.setT(startT)
         self.setupResume()
--- a/direct/src/interval/cInterval.cxx
+++ b/direct/src/interval/cInterval.cxx
@@ -175,9 +175,14 @@ pause() {
 /**
  * Restarts the interval from its current point after a previous call to
  * pause().
+ * In versions previous to 1.11, this would resume an interval even if it had ended
+ * As of 1.11 we now check if the interval has ended, and only resume it if startT has been passed
  */
 void CInterval::
 resume() {
+    if(get_state() == S_final){
+        return;
+    }
   setup_resume();
   _manager->add_c_interval(this, false);
 }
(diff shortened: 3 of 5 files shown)


rdb (changes requested):
I think this is a fine solution, thanks!  But it is a change of behaviour, so it really needs a docstring explaining the behaviour (and explaining how it changes as of 1.11.0, which will be the next release that a behaviour-changing change will land in).

I'm wondering if we should also put a similar check in the base Interval class for consistency.  Resuming just a solitary instantaneous interval like `Func(print, 'test')` will also rerun upon calling `resume()` when it's at the end, so it should probably behave the same way as `Sequence(Func(print, 'test'))`, right?

We then also should make the change to the C++ CInterval class (ie. so that `resume()` without start_t argument doesn't do anything for a finished interval).

A unit test would be a nice-to-have.

DavidACrompton:
> I'm wondering if we should also put a similar check in the base Interval class for consistency. Resuming just a solitary instantaneous interval like `Func(print, 'test')` will also rerun upon calling `resume()` when it's at the end, so it should probably behave the same way as `Sequence(Func(print, 'test'))`, right?

Been looking into this @rdb, the behaviour for `Func(print, 'test')`  and `Sequence(Func(print, 'test'))`, are different for some reason.
```
from direct.interval.IntervalManager import ivalMgr
from direct.interval.IntervalGlobal import *
from direct.task.TaskManagerGlobal import taskMgr

sut = Func(print, 'test')
sut.start()
ivalMgr.step()
taskMgr.step()
print(sut.getState())
```
This returns 0 (S_initial), where as changing `sut = Func(print, 'test')` to `sut = Sequence(Func(print, 'test'))` will have the print statement return 3 (S_final) as expected.

Perhaps FunctionInterval should be changed to have this - so it would better match the flow of a Sequence/Parallel?
```
    def privInstant(self):
        self.state = CInterval.SStarted
        # Evaluate the function
        self.function(*self.extraArgs, **self.kw)
        # Print debug information
        self.notify.debug(
            'updateFunc() - %s: executing Function' % self.name)
        self.state = CInterval.SFinal
        self.intervalDone()
```

direct/src/interval/cInterval.cxx
```diff
@@ -175,9 +175,14 @@ pause() {
...
  * Restarts the interval from its current point after a previous call to
  * pause().
+ * In versions previous to 1.11, this would resume an interval even if it had ended
+ * As of 1.11 we now check if the interval has ended, and only resume it if startT has been passed
  */
 void CInterval::
 resume() {
+    if(get_state() == S_final){
+        return;
+    }
```
rdb, line 185: Style nitpicks: indentation, spaces around parentheses in if condition, so:
```cpp
  if (get_state() == S_final) {
    return;
}
```

direct/src/interval/FunctionInterval.py
```diff
@@ -86,11 +86,18 @@ def makeUniqueName(func, suffix = ''):
         return name
 
     def privInstant(self):
+        """
+        In versions previous to 1.11, this function did not set the state of the Interval
+        In 1.11 this has been changed to be consistent with MetaInterval
+        """
```
rdb, line 92: IMO for the purposes of generating the API documentation it's not great to have a docstring on a private method, I'd be OK with omitting this or turning it into a regular comment

direct/src/interval/Interval.py
```diff
@@ -134,6 +134,12 @@ def pause(self):
         return self.getT()
 
     def resume(self, startT = None):
+        """
+        In versions previous to 1.11, this would resume an interval even if it had ended
+        As of 1.11 we now check if the interval has ended, and only resume it if startT has been passed
+        """
```
rdb, line 140: PEP8 nitpicks:
> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.

> Block comments generally consist of one or more paragraphs built out of complete sentences, with each sentence ending in a period.

rdb (changes requested):
I think I agree with the FunctionInterval change for consistency with the base implementation.

Thanks!  I just have some nitpicks.

Issue #1348, opened 2022-08-04
Feature Request: Add arm64 support for bdist_apps


DarthNihilus1:
## Description

At the moment, bdist_apps doesn't support arm64 for macOS. The change I want is simply put, arm64 support in platforms  e.g.: macosx_11_0_arm64
## Use Case

It would help when bundling apps to support Apple Silicon users.

rdb:
I put this in my setup.py:
```
'platforms': ['macosx_11_0_arm64'],
```
And it built... something.  I am unable to run it because I don't have access to macOS hardware at the moment, could someone let me know what happens then?

DarthNihilus1:
Ill do some testing with the newer commits since this issue  when I can

DarthNihilus1:
I'm still getting a problem with the freezetool when building for arm on my macbook m1 device
```
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/setuptools/__init__.py", line 153, in setup
    return distutils.core.setup(**attrs)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/direct/dist/commands.py", line 550, in run
    self.build_binaries(platform, build_dir, build_dir)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/direct/dist/commands.py", line 1036, in build_binaries
    create_runtime(platform, appname, scriptname, True)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/direct/dist/commands.py", line 1001, in create_runtime
    freezer.generateRuntimeFromStub(target_path, stub_file, use_console, {
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/direct/dist/FreezeTool.py", line 1882, in generateRuntimeFromStub
    load_commands = self._parse_macho_load_commands(stub_data)
[...]

Issue #538, opened 2019-01-16
deployment: Allow loading blob from an external final instead of embedded


Moguri:
This should make Android deployment easier.

PR #1441, opened 2023-01-14
Deploy-ng: Add custom blob support for FreezeTool

JaceMayer:
## Issue description
Apple's codesign rules for Arm64 makes it extremely difficult for us to provide a way to allow cross-platform compilation, as modifying the Mach-O binary in any way invalidates the binary signature. 

#538 - Discusses adding external blob support
#1348 - Discusses issues with building for Arm64 Macs.

## Solution description
Having discussed with rdb, we had 2 possible solutions. 
1) Ship deploy-stub with its signature stripped out, allowing FreezeTool to modify the binary. But this would require the user to sign the app bundle after FreezeTool. This however, means anyone building for Arm64 Macs won't be able to sign the binary. Which is required for running applications on ARM64 Macs.
2) Create a deploy-stub that requires an external blob to store the frozen code. The issue in this case, is that `deploy-stube` has to load unsigned code, but maintains compatibility for users who want to build for ARM64 Macs where signing is required and stricter. 

This PR implements the 2nd solution, as it allows Windows/Linux users to build for MacOS.

It adds a new parameter for build_apps called `external_blob` which determines the build behaviour we use for a MacOS app. 

When building with `external_blob=True`, we use `deploy-stube` and store the frozen code and the header in a .bin named after the target executable and store it in the Contents directory of the app bundle.

Initially I had tried to just have the Frozen code in the external blob, and having FreezeTool modify the blobInfo section of the binary, however this still invalidates the signature. 

## Checklist


rdb (changes requested):
I'm not at all a fan of duplicating the deploy-stub.c code out, that just creates a maintenance burden of keeping the two files in sync.

Maybe we can just add a flag to blobinfo indicating it should read from an external blob, and then set this flag by default in deploy-stub.c, and unset it when we're modifying the binary in FreezeTool?  That means we can just use the same binary, and keep the deploy-stub/deploy-stubw distinction.

I think we can just directly fread the blobinfo from the file in one go directly, I don't think we need to do it one field at a time (or via an intermediate buffer), we already have to determine the correct layout for the structure in memory in FreezeTool anyway.

I see you removed the mmap path - I don't think there's anything wrong with mmap *per se*, but it requires the offset to be a multiple of the page size, which means we have to offset the data by at least the page size since the header is at the beginning, wasting almost 4kb.  We *could* get around this by just putting the blobinfo at the beginning of the blob we mmap... so we first mmap and then memcpy over the blobinfo from the mmapped region the blobinfo struct, then shift the pointer by the offset in the blobinfo we read from the file to get the actual pointer to the data.  However, I don't know how advantageous this really is compared with just malloc/fread so I'm happy with that too if it otherwise makes the code too complicated.

rdb (changes requested):
Thanks for your work on this.

I'm struggling to imagine how the code works currently - you open the blob file, read the blobinfo header, then close it, then reopen the same blob file, and mmap it?  But we're not using an offset for the mmap, so I presume you've already offset the pointer values in the blob by the header offset in commands.py?  Then wouldn't it make more sense to get rid of map_header and instead just do the map_blob and memcpy over the blobinfo struct from the beginning of the mapped region?

DavidACrompton:
I've updated this PR to avoid having to close and re-open the blob. Fixed the get_blob_path function returning a pointer to temporary stack space. 

I've tested it in both a MacOS and Windows environment (even though external blobs are disabled in windows https://github.com/panda3d/panda3d/pull/1441/files#diff-ec9cca8d67396fec281b5d10d7d78423b814ce9be5771fc756fde792eba94321R1003).

@rdb Although this is only required on MacOS, perhaps we shouldn't be preventing other OS's from using this? It might cause an unexpected behaviour, when somebody sets `external_blob` to True only for commands.py to set it back to False?

I believe the only point not addressed it https://github.com/panda3d/panda3d/pull/1441#discussion_r1098773298, but I'm not sure how to go about this.

pandatool/src/deploy-stub/deploy-stub.c
```diff
@@ -553,53 +553,51 @@ int Py_FrozenMain(int argc, char **argv)
  * Maps the binary blob at the given memory address to memory, and returns the
  * pointer to the beginning of it.
  */
-static void *map_blob(off_t offset, size_t size) {
+static void *map_blob(off_t offset, size_t size, const char *path) {
```
Moguri, line 556: This new `path` argument appears to be unused?

pandatool/src/deploy-stub/deploy-stub.c
```diff
@@ -628,6 +626,62 @@ static void unmap_blob(void *blob) {
   }
 }
 
+/**
+ * Gets the file path to the games .bin file which stores the blob made by FreezeTool
+ */
+static void get_blob_path(const char *path, char buffer[]) {
```
Moguri, line 632: This seems to duplicate a fair bit of logic from `map_blob() `. Maybe create a `get_runtime_path()` to put some of that duplicate code into?

pandatool/src/deploy-stub/deploy-stub.c
```diff
@@ -628,6 +626,62 @@ static void unmap_blob(void *blob) {
...
+#endif
+}
+
+/*
+ * When we are dealing with an external file, we have to load
+ * the blobinfo manually. This is to resolve an codesigning issues
+ * on ARM64 MacOS
+ */
+static void *map_external_blob(const char *path) {
+    FILE *runtime;
```
Moguri, line 666: We should call this variable something else. We're not opening the runtime file, so this is confusing. Maybe blob_file?

pandatool/src/deploy-stub/deploy-stub.c
```diff
@@ -640,8 +694,21 @@ int main(int argc, char *argv[]) {
   ModuleDef *moddef;
   const char *log_filename;
   void *blob = NULL;
+#if defined(MS_WINDOWS)
```
Moguri, line 697: Do we really need an ifdef here? Maybe we already have a platform independent define somewhere for this?

pandatool/src/deploy-stub/deploy-stub.c
```diff
@@ -640,8 +694,21 @@ int main(int argc, char *argv[]) {
...
+#else
+  char blob_path[PATH_MAX];
+#endif
+  get_blob_path(argv[0], blob_path);
   log_filename = NULL;
-
+  // FreezeTool will overwrite blobinfo on non-MacOS to set the blob_offset 
+  // to non -1. If it is -1, then we need to load it from FileSystem
+  int load_external = 0;
+  if (blobinfo.blob_offset == -1) {
```
Moguri, line 707: Can we combine this with the later blob_offset check to avoid needing to define a new variable to block the later branch. Does this have to happen earlier for some reason?

pandatool/src/deploy-stub/deploy-stub.c
```diff
@@ -640,8 +694,21 @@ int main(int argc, char *argv[]) {
   ModuleDef *moddef;
   const char *log_filename;
   void *blob = NULL;
+#if defined(MS_WINDOWS)
+  char blob_path[MAX_PATH];
+#else
+  char blob_path[PATH_MAX];
+#endif
```
Moguri, line 701: We probably don't need to get the blob path unless we plan to load an external blob. Maybe just move this to map_external_blob() to match map_blob()?

Moguri (changes requested):

Moguri:
@rdb is there a reason to keep the append_blob option? It seems unlikely that we'll have a deploy-stub that doesn't match FreezeTool. I can see having some caution when first adding the feature, but we've probably had enough releases now, right?

PR #1443, opened 2023-01-15
Interval: Assertion error when given very long external intervals

### LINKED ISSUE #488: interval: Assertion error when given very long external intervals
In a slight rush so forgive my brevity:

- `MovieTypeRegistry` gives a naked `MovieAudio` if asked to load a file with an unrecognized extension
- `MovieAudioCursor` defaults to a length of 10 billion seconds (`_length(1.0E10),`)
- Creating a `SoundInterval` that would play this yields a 10-billion-second-long interval
- `CMetaInterval` does not use `double` for time; time is tracked using an `int` (the units are milliseconds by default)
- 10 billion seconds' worth of milliseconds overflows `int` to a negative number, so adding a 10-billion-second interval to a `CMetaInterval` (e.g. `Sequence`) in a way that forces the `CMetaInterval` to sort its events list will result in the end event getting sorted before the beginning event.
- Trying to play the interval causes an assertion error due to trying to end an interval that isn't active.

This one bit me pretty bad. I'm thinking we may need an assert that stops overly-long intervals from being added to `CMetaInterval` (or use a larger integer type?) and we should treat invalid sounds as 0-second empty clips, not 3 centuries long.

Thoughts?

<details><summary>Python code causing the error</summary>

    from panda3d.core import *
    from direct.interval.IntervalGlobal import *

    # Use the OpenAL AudioManager so we can be sure it plays MovieAudio
    load_prc_file_data('', 'audio-library-name p3openal_audio')
    amgr = AudioManager.create_AudioManager()

    # Load a "Load-Failure Stub" - this is what MovieTypeRegistry does if the file
    # can't be recognized
    sound = SoundInterval(amgr.get_sound(MovieAudio('Load-Failure Stub')))

    # 'sound' now has a duration of 10 billion seconds

    # Wrap it in a sequence
    seq = Sequence()

    # Add the SoundInterval *second* to force CMetaInterval to sort the events list
    seq.append(WaitInterval(1.0))
    seq.append(sound)

    # Try to run the sequence
    seq.start()
    ivalMgr.step()
</details>

JaceMayer:
## Issue description
As per issue #488 (Interval: Assertion error when given very long external intervals), this PR deals with two issues:
1. Creating an interval with a duration longer than `int`'s maximum value, triggers an assertion error in cMetaInterval. 
2. Creating an invalid MovieAudioCursor has a default length of 3 centuries

## Solution description
The PR aims to resolve this issue in 3 different steps in order to address the multiple issues at play:
1. Durations in cMetaInterval have been changed to a `long long` type as opposed to the `int` type that was used before
2. MovieAudioCursor default length has been changed to 0 seconds, this is to match the fact that it has 0 samples/frames
3. `loader.loadSound` now follows the behaviour of `loadModel` and raises an `IOError` if a `NullAudioSound` was returned

## Checklist

Fixes #488

### DIFF
--- a/direct/src/interval/cMetaInterval.I
+++ b/direct/src/interval/cMetaInterval.I
@@ -127,7 +127,7 @@ get_event_type() const {
 INLINE int CMetaInterval::
 double_to_int_time(double t) const {
   // Use floor() just in case there are negative values involved.
-  return (int)floor(t * _precision + 0.5);
+  return (long long)floor(t * _precision + 0.5);
 }
 
 /**
@@ -135,7 +135,7 @@ double_to_int_time(double t) const {
  * double time value or offset in seconds.
  */
 INLINE double CMetaInterval::
-int_to_double_time(int time) const {
+int_to_double_time(long long time) const {
   return (double)time / _precision;
 }
 
@@ -143,7 +143,7 @@ int_to_double_time(int time) const {
  *
  */
 INLINE CMetaInterval::PlaybackEvent::
-PlaybackEvent(int time, int n,
+PlaybackEvent(long long time, int n,
               CMetaInterval::PlaybackEventType type) :
   _time(time),
   _n(n),
@@ -164,7 +164,7 @@ operator < (const CMetaInterval::PlaybackEvent &other) const {
  *
  */
 INLINE CMetaInterval::EventQueueEntry::
-EventQueueEntry(int n, CInterval::EventType event_type, int time) :
+EventQueueEntry(int n, CInterval::EventType event_type, long long time) :
   _n(n),
   _event_type(event_type),
   _time(time)
--- a/direct/src/showbase/Loader.py
+++ b/direct/src/showbase/Loader.py
@@ -949,7 +949,7 @@ class Loader(DirectObject):
             return None
 
     def loadSound(self, manager, soundPath, positional = False,
-                  callback = None, okMissing=None, extraArgs = []):
+                  callback = None, extraArgs = [], okMissing=None):
 
         """Loads one or more sound files, specifying the particular
         AudioManager that should be used to load them.  The soundPath
@@ -976,7 +976,7 @@ class Loader(DirectObject):
                 sound = manager.getSound(soundPath, positional)
                 result.append(sound)
 
-                if str(sound).startswith('NullAudioSound') and not okMissing:
+                if manager.getNullSound() and not okMissing:
                     message = 'Could not load audio file: %s' % (soundPath)
                     raise IOError(message)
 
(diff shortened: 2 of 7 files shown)


direct/src/showbase/Loader.py
```diff
@@ -976,6 +976,10 @@ def loadSound(self, manager, soundPath, positional = False,
                 sound = manager.getSound(soundPath, positional)
                 result.append(sound)
 
+                if str(sound).startswith('NullAudioSound') and not okMissing:
```
rdb, line 979: This is not a clean way to check for this.  You can compare to `manager.getNullSound()` instead.

direct/src/showbase/Loader.py
```diff
@@ -949,7 +949,7 @@ def loadMusic(self, *args, **kw):
             return None
 
     def loadSound(self, manager, soundPath, positional = False,
-                  callback = None, extraArgs = []):
+                  callback = None, okMissing=None, extraArgs = []):
```
rdb, line 952: Style consistency nitpick
Also, this will break if people passed in the arguments positionally, rather than by keyword, so let's preserve the order

rdb (changes requested):
I think we need some unit tests for this.

DavidACrompton:
Thanks for the review @rdb, I'll get this sorted this afternoon!

direct/src/showbase/Loader.py
DavidACrompton: Fixed the argument order in https://github.com/panda3d/panda3d/pull/1443/commits/3ef954c80be2992f4a0eaa9b9a4e50393583be53

direct/src/interval/cMetaInterval.I
```diff
@@ -127,23 +127,23 @@ get_event_type() const {
 INLINE int CMetaInterval::
 double_to_int_time(double t) const {
   // Use floor() just in case there are negative values involved.
-  return (int)floor(t * _precision + 0.5);
+  return (long long)floor(t * _precision + 0.5);
```
rdb, line 130: Don't see what this does, other than add a compiler warning?

tests/interval/test_interval_length.py
```diff
@@ -0,0 +1,33 @@
+from panda3d.core import *
+from direct.interval.IntervalGlobal import *
+import pytest
+
+
+def test_invalid_sound_interval():
+    load_prc_file_data('', 'audio-library-name p3fmod_audio')
```
rdb, line 7: Why does this test require FMOD?

What happens if FMOD is not available?

What happens to tests running after this one?  If you do this it should at leat be in a try..finally where you unload the page in the `finally`, see `test_track_gui_items` in `tests/gui/test_DirectGuiBase.py` for an example

rdb (commented):
Thanks, some questions...

Issue #922, opened 2020-04-29
actor: Animations not working until actor defined as part of class


Benjamin8693:
Today I was doing some experimentation with Actors, when I ran into quite a weird issue. Originally, my code looked something like this:

```
model = Actor.Actor('my-mod.bam', {'spin': 'my-spin.bam'})
model.loop('spin')
```

For some reason, my actor was not doing his animations. I tried swapping out actors to make sure the one I was using didn't have something faulty with the model, but the issue kept happening. I asked a friend about this, and he told me he had this issue before. He recommended that I define my actor as part of the class, so I did. My code then looked like this:

```
self.model = Actor.Actor('my-mod.bam', {'spin': 'my-spin.bam'})
self.model.loop('spin')
```

Only after I did this was my actor animating correctly. I was a bit confused when it began to work, so I decided to post an issue here about it.

Thanks,
Benjamin

Moguri:
This is a known quirk of the `Actor` class. However, I was surprised to not find an existing issue for it.

When an Actor instance is destroyed, it will stop playing animations (I don't remember if the animation is explicitly stopped, or just no longer updating), and you need to keep a reference around to avoid having the instance garbage collected. This is why adding it as a variable to another instance solves the problem (the ref count is incremented and thus doesn't go to zero).

I agree this is strange behavior and not very newbie friendly. If we can fix this in a non-API breaking way I would be all for it. However, it would break games that were relying on this (i.e., expecting animations to stop when the instance went out of scope).

rdb:
Fundamentally, the problem is with Actor's design.  It is a Python class, therefore not living in the scene graph, extending a C++ class that does live in the scene graph.  So Panda continues to keep a reference to the C++ bit (the unanimated model) even when the Python bit (containing the animations) goes out of scope.

I think we probably want to replace Actor with a C++ interface.  That might involve improving the interface of Character to support all the things Actor does.  Or, creating a new C++-based interface to manage the AnimControls playing on a node.

PR #1444, opened 2023-01-15
Actor: Allow Actor to store a reference of instantiated Actors

DavidACrompton:
## Issue description
Feel free to delete/close if unhelpful/out-of-scope

An issue that comes up now and again, is that Actors become static as soon as it becomes out-of-scope. This is because as soon as the Actor becomes out of scope, Python's GC cleans up the Python class, which leaves the C++ NodePath still active. 

Related issue: #922 

## Solution description
My solution is to have Actor store a reference to instantiated Actors. This behaviour is toggled by a `ConfigVariableBool` as some developers may have leveraged this behaviour in their games. The config variable allows them to switch back to the old behaviour, where the model becomes static when the Actor becomes out-of-scope. 

This makes Actor easier to use and more intuitive for newbies, although at the cost of having to store a reference to each actor.  

## Checklist

### DIFF
--- a/direct/src/actor/Actor.py
+++ b/direct/src/actor/Actor.py
@@ -18,6 +18,8 @@ class Actor(DirectObject, NodePath):
     Actor class: Contains methods for creating, manipulating
     and playing animations on characters
     """
+    activeActors = []
+
     notify = DirectNotifyGlobal.directNotify.newCategory("Actor")
     partPrefix = "__Actor_"
 
@@ -31,6 +33,7 @@ class Actor(DirectObject, NodePath):
     validateSubparts = ConfigVariableBool('validate-subparts', True)
     mergeLODBundles = ConfigVariableBool('merge-lod-bundles', True)
     allowAsyncBind = ConfigVariableBool('allow-async-bind', True)
+    storeReference = ConfigVariableBool('store-actor-reference', True)
 
     class PartDef:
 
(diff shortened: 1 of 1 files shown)


rdb:
You could store it as a Python tag on the underlying node too.  But the main problem with either approach is that it feels unpythonic to have an object "live forever" by default rather than be able to be cleaned up automatically by the garbage collector.

Another approach that wouldn't preserve the Actor but would preserve the playing animations is to store a reference to the AnimControls that are currently playing.  This is the behaviour most people getting confused by after all.

I'm going to wait for others to weigh in on this.

DavidACrompton:
> You could store it as a Python tag on the underlying node too. But the main problem with either approach is that it feels unpythonic to have an object "live forever" by default rather than be able to be cleaned up automatically by the garbage collector.

Very true and I completely agree, I feel as though it comes down to the fact that Actor was seemingly designed to have a pointer retained for it. I think for this issue, any solution is going to be less than elegant.

Maxwell175:
> You could store it as a Python tag on the underlying node too. But the main problem with either approach is that it feels unpythonic to have an object "live forever" by default rather than be able to be cleaned up automatically by the garbage collector.

IMHO I think this is the way to go and would be the most "panda3d-ic" solution since it ties the object's lifetime to the node, which is probably most intuitive specifically for someone using panda3d as they would be used to Nodes living until they are removed from the scene graph

rdb:
Well, no, it doesn't tie the lifetime to the node.  It makes the actor *and* the node live forever, because it creates a reference cycle.

rdb:
Sorry, I'm going to close this as I don't agree with the solution - and I am loath to give users who would use this to fix an issue they don't understand a tool to shoot themselves in the foot with.

Issue #1395, opened 2022-11-18
Stop using distutils Python module


Moguri:
## Description
The `distutils` module  has been deprecated for a while and is slated for removal in Python 3.12. More information on this can be found in [PEP 632](https://peps.python.org/pep-0632/). Panda will need to stop making use of `distutils` in order to support Python 3.12.

I have identified the following that need to be addressed:
- [ ] `distutils.log` in `direct/src/dist` - replace with builtin `logging` module
- [x] `distutils.core.setup` in `contrib/src/panda3dtoolsgui` - replace with `setuptools.setup`
- [x] `distutils.sysconfig` in various locations - *should* be replace by `sysconfig`, but I am unsure if the issues from #1230 are resolved
- [x] `distutils.utils.get_platform()` in `makepanda.py` - replace with `platform` module

Moguri:
For `distutils.sysconfig`, I am thinking we make a `locations.py` utility script that re-exports the functions we need from `sysconfig` if the Python version is at least 3.12 and from `distutils.sysconfig` otherwise. If a user is actually using 3.12 on Debian, then hopefully this mess has been sorted out.

rdb:
One question on my mind, do we need to worry about this for 1.10? [Python 3.12 is due in October 2023](https://peps.python.org/pep-0693/), at which time I would really like to have already released Panda 1.11.  On the other hand, some users may stick with 1.10 for longer, but I'm not sure if they need Python 3.12 support.

Moguri:
The changes seem pretty minimal, so I was thinking about targeting 1.10 just to be safe. However, if we run into any roadblocks with that, I would be fine with targeting 1.11.

DavidACrompton:
I've had a look at #1230 which is related to this. On a fresh ubuntu install, I get the following. 
```
python3 -m site
sys.path = [
    '/home/parallels',
    '/usr/lib/python310.zip',
    '/usr/lib/python3.10',
    '/usr/lib/python3.10/lib-dynload',
    '/home/parallels/.local/lib/python3.10/site-packages',
    '/usr/local/lib/python3.10/dist-packages',
    '/usr/lib/python3/dist-packages',
]
USER_BASE: '/home/parallels/.local' (exists)
USER_SITE: '/home/parallels/.local/lib/python3.10/site-packages' (exists)
ENABLE_USER_SITE: True

python3 -m sysconfig
Platform: "linux-aarch64"
Python version: "3.10"
Current installation scheme: "posix_local"

[...]

Issue #1230, opened 2022-01-04
Python SDK is installed in wrong site


el-dee:
## Description

With recent master builds, the Python SDK is installed in the wrong site-packages on Linux Mint 20 (And I assume Ubuntu Focal). This problem occurs since 2fcacd1bab1226ea0571c5d925daf6028a862f96

Previously, the Python SDK was installed in `/usr/lib/python3/dist-packages` but now it is in `/usr/lib/python3.8/site-packages`
However this path is not in the configured Python 3.8 site :

<pre>
$ python3.8 -m site
sys.path = [
    '/usr/lib/python38.zip',
    '/usr/lib/python3.8',
    '/usr/lib/python3.8/lib-dynload',
    '/home/ld/.local/lib/python3.8/site-packages',
    '/usr/local/lib/python3.8/dist-packages',
    '/usr/lib/python3/dist-packages',
    '/usr/lib/python3.8/dist-packages',
]
USER_BASE: '/home/ld/.local' (exists)
USER_SITE: '/home/ld/.local/lib/python3.8/site-packages' (exists)
ENABLE_USER_SITE: True
</pre>

## Steps to Reproduce

    python3.8 makepanda/makepanda.py --everything --no-gles --no-gles2 --no-egl --use-sse2 --threads=4 --optimize=3 --wheel --installer

Disclaimer: I have another version of Python (3.7) also installed on this system, but until now it was not an issue :)

## Environment
 * Operating system: Linux Mint 20
 * System architecture: amd64
 * Panda3D version: latest master
 * Installation method: built from source
 * Python version (if using Python): Python 3.8.5
 * Compiler (if using C++): n/a

rdb:
Wow, that is annoying. I guess we have to go back to the deprecated module for now and revisit this when Python 3.12 comes around, and pray that they've fixed this bug by then.

el-dee:
After a bit of digging, it seems that Debian is patching `distutils.sysconfig` to use their own paths for site-packages, but not (yet) `sysconfig`.

rdb:
Reverted to `distutils.sysconfig`, with added comments not to "fix" the deprecated use.

PR #1445, opened 2023-01-15
Remove the usage of the deprecated distutils module

DavidACrompton:
## Issue description
As Python 3.12 will remove the distutils module per  [PEP 632](https://peps.python.org/pep-0632/), in order to support 3.12 Panda needs to remove distutils in favour of the setuptools module.

Related Issue: #1395 

## Solution description
I have solved this by:

1. Creating a locations.py file for FreezeTool to use, that on Python 3.10 and above uses the sysconfig module to get paths, and uses distutils on lower versions
2. Changed FreezeTool.py, commands.py, and installers.py to use the built in logging module, as opposed to using distutils.log
3. Changed makepanda back to RDB's commit https://github.com/panda3d/panda3d/commit/2fcacd1bab1226ea0571c5d925daf6028a862f96, where he removes distutils in favour of sysconfig

## Notes

Whilst working on this, I came across issue #1230, which was in response to RDB's commit removing distutils in favour of sysconfig. I have tested this on Ubuntu and Debian itself, and had no issues. The sysconfig module now includes the correct paths https://github.com/panda3d/panda3d/issues/1395#issuecomment-1383126253 

## Checklist

### DIFF
--- a/direct/src/dist/commands.py
+++ b/direct/src/dist/commands.py
@@ -19,7 +19,7 @@ import string
 import tempfile
 
 import setuptools
-import distutils.log
+import logging
 
 from . import FreezeTool
 from . import pefile
@@ -259,6 +259,7 @@ class build_apps(setuptools.Command):
     ]
     default_file_handlers = {
     }
+    log = logging.getLogger("build_apps")
 
     def initialize_options(self):
         self.build_base = os.path.join(os.getcwd(), 'build')
@@ -512,7 +513,8 @@ class build_apps(setuptools.Command):
             self.icon_objects[app] = iconobj
 
     def run(self):
-        self.announce('Building platforms: {0}'.format(','.join(self.platforms)), distutils.log.INFO)
+        build_apps.log.setLevel(logging.WARNING)
+        self.announce('Building platforms: {0}'.format(','.join(self.platforms)), logging.INFO)
 
         for platform in self.platforms:
             # Create the build directory, or ensure it is empty.
@@ -580,6 +582,12 @@ class build_apps(setuptools.Command):
             if self.macos_main_app and 'macosx' in platform:
                 self.bundle_macos_app(build_dir)
 
+    def announce(self, msg, level=logging.INFO):
+        """ Override the command's built-in announce function as older versions of setuptools does not
+        use logging style levels.
+        """
+        build_apps.log.log(level, msg)
+
     def download_wheels(self, platform):
         """ Downloads wheels for the given platform using pip. This includes panda3d
         wheels. These are special wheels that are expected to contain a deploy_libs
@@ -588,7 +596,7 @@ class build_apps(setuptools.Command):
 
         import pip
 
-        self.announce('Gathering wheels for platform: {}'.format(platform), distutils.log.INFO)
+        self.announce('Gathering wheels for platform: {}'.format(platform), logging.INFO)
 
         whlcache = os.path.join(self.build_base, '__whl_cache__')
 
@@ -674,7 +682,7 @@ class build_apps(setuptools.Command):
         fwdir = os.path.join(contentsdir, 'Frameworks')
         resdir = os.path.join(contentsdir, 'Resources')
 
-        self.announce('Bundling macOS app into {}'.format(appdir), distutils.log.INFO)
+        self.announce('Bundling macOS app into {}'.format(appdir), logging.INFO)
 
         # Create initial directory structure
         os.makedirs(macosdir)
@@ -819,7 +827,7 @@ class build_apps(setuptools.Command):
                 if not localtag.endswith('opt'):
                     self.announce(
                         'Could not find an optimized wheel (using index {}) for platform: {}'.format(self.optimized_wheel_index, platform),
-                        distutils.log.WARN
+                        logging.WARN
                     )
 
             for whl in wheelpaths:
@@ -837,7 +845,7 @@ class build_apps(setuptools.Command):
             path.insert(0, os.path.join(p3dwhlfn, 'deploy_libs'))
 
 
-        self.announce('Building runtime for platform: {}'.format(platform), distutils.log.INFO)
+        self.announce('Building runtime for platform: {}'.format(platform), logging.INFO)
 
         # Gather PRC data
         prcstring = ''
@@ -1170,7 +1178,7 @@ class build_apps(setuptools.Command):
             if module not in freezer_modules:
                 continue
 
-            self.announce('Copying data files for module: {}'.format(module), distutils.log.INFO)
+            self.announce('Copying data files for module: {}'.format(module), logging.INFO)
 
             # OK, find out in which .whl this occurs.
             for whl in wheelpaths:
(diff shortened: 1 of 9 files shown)


direct/src/dist/locations.py
```diff
@@ -0,0 +1,16 @@
+import sys
+
+if sys.version_info < (3, 10):
```
rdb, line 3: I am not sure this check should be for 3.10, unless we are sure that no Ubuntu/Debian versions exist with a broken version of Python 3.10.  I suggest we do some digging to find out where/when they fixed it.

makepanda/installpanda.py
```diff
@@ -13,8 +13,7 @@
 from optparse import OptionParser
 from makepandacore import *
 
-# DO NOT CHANGE TO sysconfig - see GitHub issue #1230
-from distutils.sysconfig import get_python_lib
+import sysconfig
```
rdb, line 16: There's a reason the comment is here.  This would break when building on older Ubuntu versions.

direct/src/dist/locations.py
```diff
@@ -0,0 +1,16 @@
...
+    PythonIPath = sysconf.get_python_inc()
+    PythonVersion = sysconf.get_config_var("LDVERSION") or sysconf.get_python_version()
+    Python = sysconf.PREFIX
+    lib_dir = sysconf.get_python_lib(plat_specific=1, standard_lib=1)
+    config_vars = sysconf.get_config_vars()
+else:
+    import sysconfig as sysconf
+    PythonIPath = sysconf.get_path("include")
+    PythonVersion = sysconf.get_config_var("LDVERSION") or sysconf.get_python_version()
+    Python = sysconf.get_config_var('prefix')
```
rdb, line 14: Not a fan of the naming of these variables

direct/src/dist/commands.py
```diff
@@ -580,6 +582,12 @@ def run(self):
             if self.macos_main_app and 'macosx' in platform:
                 self.bundle_macos_app(build_dir)
 
+    def announce(self, msg, level=logging.INFO):
+        """ Override the command's built-in announce function as older versions of setuptools does not
+        use logging style levels.
+        """
+        build_apps.log.log(level, msg)
```
rdb, line 589: Could you explain this?  Does this interact with the setuptools logging?  I would like not to be using two log systems at the same time.

rdb (changes requested):
Thanks for the PR.  Comments inline.

direct/src/dist/commands.py
DavidACrompton: Yes, essentially setuptools includes the disttools logger, but they're monkey-patch the some functions in the distutil  version (https://github.com/pypa/setuptools/blob/main/setuptools/logging.py#L27) to use the builtin logging module. In other places they use the logging module but with the log levels divided by 10. (so WARNING would become WARNING/10), I thought it best to override the `announce` function from setuptools, to use the logging module solely in `commands.py`. Happy to change this.

direct/src/dist/locations.py
DavidACrompton: @rdb, had a dig around for this on Debian Salsa. Looking at the change log it was Debian `python3.10 (3.10.2-3)` on Fri, 25 Feb 2022, when they re-introduced patching sysconfig. This does mean that there may be some people with a broken sysconfig on python3.10, but updating their python via APT should solve it. 
Ultimately, it's a judgement call for you to make. 
https://salsa.debian.org/cpython-team/python3/-/blob/master/debian/changelog#L202

Moguri: Instead of a version check, I would prefer that we just attempt to use `distutils` and fallback to `sysconfig` on an `ImportError`.

direct/src/dist/locations.py
Moguri: I don't either. I would prefer this file to just be a super thin wrapper for `distutils.sysconf`, even keeping the original names:

```python
try:
  from distutils.sysconfig import *

 # not sure if this works or if we need to setup explicit aliases
  __all__ = ['get_python_inc', 'get_config_var', ...] 
except ImportError:
  from sysconfig  import *

  __all__ = ['get_config_var', ...]

  def get_python_inc():
    return get_path('include')
```

makepanda/installpanda.py
Moguri: We might as well use `locations.py` here too.

makepanda/makepackage.py
```diff
@@ -8,6 +8,7 @@
 import subprocess
 from makepandacore import *
 from installpanda import *
+import sysconfig
```
Moguri: Prefer `locations.py` over `sysconfig`

makepanda/makepandacore.py
```diff
@@ -6,7 +6,7 @@
 ########################################################################
 
 import configparser
-from distutils import sysconfig # DO NOT CHANGE to sysconfig - see #1230
+import sysconfig
```
Moguri: Prefer `locations.py` over `sysconfig`

Moguri (changes requested):

direct/src/dist/locations.py
rdb: I'm happy with either an ImportError fallback, a version check for for Python 3.12+ or a version check for Python 3.10.3+.

Being certain that we don't break any build configuration weighs more heavily than possibly having a deprecation warning be shown in some configurations.

makepanda/installpanda.py
rdb: I think the difficulty is that we can't import direct modules in makepanda.  It's not immediately obvious to me what the neatest solution is to avoid duplicating that file.

direct/src/dist/commands.py
rdb: I see.  That's annoying.  I can't think of a cleaner solution at the moment.

DavidACrompton:
I've changed locations.py as per Moguri's recommendation to make it a wrapper around distutils.sysconfig, and created aliases for the functions that differ in the newer sysconfig. 

I've changed the version check to test for 3.12 or above, just in consideration that although sysconfig  now gets patched in mainline debian, theres still other distros that may lag behind the mainline. As rdb noted, a deprecation warning is preferable to breaking some build configurations. 

I also reverted the makepanda changes, if anyone any suggestions on how the best way to handle sysconfig there for 3.12+?

direct/src/dist/locations.py
```diff
@@ -0,0 +1,18 @@
+import sys
+
+if sys.version_info >= (3, 12):
+    from distutils.sysconfig import *
+    __all__ = ['get_python_inc', 'get_config_var', "get_python_version", "PREFIX", "get_python_lib", "get_config_vars"]
+else:
+    from sysconfig import *
+
+    __all__ = ['get_python_inc', 'get_config_var', "get_python_version", "PREFIX", "get_python_lib", "get_config_vars"]
```
rdb, line 9: This is the same in both cases, put this at the top

Also, use consistent quotation marks

direct/src/dist/commands.py
rdb: I think we have to verify that it's still possible to increase the output verbosity from the command-line with the setuptools verbosity flag

rdb (changes requested):
As for makepanda, I think we don't have much choice but to copy the locations.py file, or integrate the functions into makepandacore.py

Moguri:
> As for makepanda, I think we don't have much choice but to copy the locations.py file, or integrate the functions into makepandacore.py; since makepanda is on death row I don't mind not being super clean about that

Yeah, I am fine with just duplicating the file. It shouldn't need much, if any, updating, and, as you said, makepanda is on the way out anyways.

Moguri:
Any updates on this? Now that Python 3.12 is officially out, we should really get this sorted.

rdb:
Work continues in #1549.

PR #1446, opened 2023-01-20
cocoadisplay: Capture display before switching to fullscreen mode

el-dee:
## Issue description

On macOS, If you ask your app to enter a fullscreen mode with a lower resolution than the desktop then quit the app, when the display is switching back from the fullscreen display mode to the desktop display mode you can observe that all the other windows have been resized to match the size if the fullscreen mode and are resized back to their previous size.
If you have many windows opened (or heavy task running in the background), this resize back is slow and sometimes stops (either because there is a timeout or a bug in Cocoa).

The root cause is that in the current code, the display is captured by the application after switching the display mode. This mode switching triggers a resize event for all the opened windows, and the reverse-resize when the desktop mode is restored.

If the display is captured before switching mode, only the requesting application receives the resize notification, all the other ones stay as is.

For multi-display setup, this is not enough, though. You have to capture all the displays in order to avoid all the other windows to be resized.

Another issue that is solved by capturing the display before switching mode is that when leaving the fullscreen mode, the window is not repositioned at its former origin and instead stays at (0, 0) although the move event has been triggered by Cocoa. This results  with the window partly under the menubar of the Finder and there is an offset between the position of the mouse cursor and the position calculated by Panda. And to add to the craziness, this does not occur with a multidisplay setup

## Solution description

Once the root cause is known, the solution is trivial: the display is now captured before switching mode, and in case of multi-display, all the displays are captured.
This has also the added benefit that mode switching is much faster now.

## Checklist

See #1321 for additional informations.

### DIFF
--- a/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
+++ b/panda/src/cocoadisplay/cocoaGraphicsWindow.mm
@@ -1417,7 +1417,11 @@ do_switch_fullscreen(CFDictionaryRef mode) {
 #else
     CGDisplaySwitchToMode(_display, _windowed_mode);
 #endif
-    CGDisplayRelease(_display);
+    if (CGDisplayIsMain(_display)) {
+      CGReleaseAllDisplays();
+    } else {
+      CGDisplayRelease(_display);
+    }
     _windowed_mode = NULL;
     _context_needs_update = true;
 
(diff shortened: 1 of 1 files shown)


rdb (approved):
Thank you!

PR #1448, opened 2023-01-22
respect textures-header-only for embedded textures

lachbr:
## Issue description
The `textures-header-only` config option is not respected for textures that have been embedded in a bam file.  This is likely an oversight, as the common use case for textures-header-only is to reduce memory usage for processes that don't need the texture data, like servers.

## Solution description
Simply skip over the bytes containing the image data if textures-header-only is set.

## Checklist

### DIFF
--- a/panda/src/gobj/texture.cxx
+++ b/panda/src/gobj/texture.cxx
@@ -10693,10 +10693,15 @@ do_fillin_rawdata(CData *cdata, DatagramIterator &scan, BamReader *manager) {
       return;
     }
 
-    PTA_uchar image = PTA_uchar::empty_array(u_size, get_class_type());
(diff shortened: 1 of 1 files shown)


rdb:
Conflicted about this, because it is lossy.  Normally you don't lose information because you retain the filename to load from.  With this change you can no longer safely roundtrip bam files without losing data.

Can you elaborate further on your use case?

Maxwell175:
My use case for this PR was saving memory on the headless/server side of things, while still being able to reuse the same bam files as on the client.

rdb:
That sounds useful in principle, but does `textures-header-only` actually currently accomplish that?  It creates a RAM image filled with a solid blue colour, no?

Issue #1429, opened 2023-01-08
Implement Directional Sound for OpenAL


git2323:
## Description
I would like to use the directional sound feature of OpenAL as described in the [specification](http://openal.org/documentation/openal-1.1-specification.pdf) on page 38 (Direction and Cone). I tried to implement that on 1.10.x and it worked. Would you accept an PR for that?

## Use Case
This will allow developers to add attenuation based on the orientations of the 3D sound sources.

rdb:
This sounds useful.  What did you have in mind for the interface, ie. how will users set the direction and cone?

rdb:
Does FMOD support this feature, or would it be OpenAL only?

git2323:
FMOD also has this feature, although my current code affects OpenAL only.

rdb:
That all sounds fine.  As long as the parameterisation is defined such that someone will be able to go and implement the FMOD side later.

Note that the new arguments to set_3d_attributes should be optional.  The default could be the forward direction.

git2323:
Can you take a look at pull request #1451? Thanks.

git2323:
PR #1451 has been merged. This issue can be closed.

rdb:
For future reference, put "Fixes #1429" at the bottom the PR text and/or the commit description and the issue is closed automatically when the changes are merged.

PR #1451, opened 2023-01-27
Openal directional sound

git2323:
## Issue description
This change implements directional sound for OpenAL. The related issue is [1429](https://github.com/panda3d/panda3d/issues/1429)

## Solution description
audio_sound->set_3d_attributes() has been extended with three parameters dx,dy,dz for the direction of the sound source.
these new functions define the cone:
audio_sound->set_3d_cone_inner_angle(20.0f);
audio_sound->set_3d_cone_outer_angle(25.0f);
and to set the gain of the sound:
audio_sound->set_3d_cone_outer_gain(0.01f);
The getters are:
audio_sound->get_3d_cone_inner_angle()
audio_sound->get_3d_cone_outer_angle()
audio_sound->get_3d_cone_outer_gain()

## Checklist
* [ X] …I have familiarized myself with the CONTRIBUTING.md file
* [ X] …this change follows the coding style and design patterns of the codebase
* [ X] …I own the intellectual property rights to this code
* [ X] …the intent of this change is clearly explained
* [ X] …existing uses of the Panda3D API are not broken
* [ X] …the changed code is adequately covered by the test suite, where possible.


panda/src/audio/audioSound.cxx
```diff
@@ -35,12 +35,12 @@ AudioSound() {
...
 
 void AudioSound::
-set_3d_attributes(PN_stdfloat px, PN_stdfloat py, PN_stdfloat pz, PN_stdfloat vx, PN_stdfloat vy, PN_stdfloat vz) {
+set_3d_attributes(PN_stdfloat px, PN_stdfloat py, PN_stdfloat pz, PN_stdfloat vx, PN_stdfloat vy, PN_stdfloat vz, PN_stdfloat dx, PN_stdfloat dy, PN_stdfloat dz) {
   // Intentionally blank.
 }
 
 void AudioSound::
-get_3d_attributes(PN_stdfloat *px, PN_stdfloat *py, PN_stdfloat *pz, PN_stdfloat *vx, PN_stdfloat *vy, PN_stdfloat *vz) {
+get_3d_attributes(PN_stdfloat* px, PN_stdfloat* py, PN_stdfloat* pz, PN_stdfloat* vx, PN_stdfloat* vy, PN_stdfloat* vz, PN_stdfloat* dx, PN_stdfloat* dy, PN_stdfloat* dz) {
```
rdb, line 43: The asterisk location was correct before you changed it.  These qualifiers go around the identifier, not the type.  That's because C++ is parsed that way.

Some explanation of why it would be confusing to put it next to the type:
```cpp
int* a, b;
```
`a` is an `int*` but `b` is an `int`.  The following makes it much more obvious that b isn't a pointer:
```cpp
int *a, b;
```

That's why we put it around the identifier.

panda/src/audio/audioSound.cxx
```diff
@@ -66,6 +66,39 @@ get_3d_max_distance() const {
...
+
+void AudioSound::
+set_3d_cone_outer_gain(PN_stdfloat gain) {
+    // Intentionally blank.
+}
+
+PN_stdfloat AudioSound::
+get_3d_cone_outer_gain() const {
+    // Intentionally blank.
+    return 0.0f;
```
rdb, line 99: Indent is 2 spaces

panda/src/audio/audioSound.h
```diff
@@ -89,13 +90,15 @@ class EXPCL_PANDA_AUDIO AudioSound : public TypedReferenceCount {
...
+  // PER SECOND (default: meters). dx, dy and dz are the emitter's direction
   virtual void set_3d_attributes(PN_stdfloat px, PN_stdfloat py, PN_stdfloat pz,
-                                 PN_stdfloat vx, PN_stdfloat vy, PN_stdfloat vz);
-  virtual void get_3d_attributes(PN_stdfloat *px, PN_stdfloat *py, PN_stdfloat *pz,
-                                 PN_stdfloat *vx, PN_stdfloat *vy, PN_stdfloat *vz);
+      PN_stdfloat vx, PN_stdfloat vy, PN_stdfloat vz,
+      PN_stdfloat dx = LVector3::forward().get_x(), PN_stdfloat dy = LVector3::forward().get_y(), PN_stdfloat dz = LVector3::forward().get_z());
+  virtual void get_3d_attributes(PN_stdfloat* px, PN_stdfloat* py, PN_stdfloat* pz,
+      PN_stdfloat* vx, PN_stdfloat* vy, PN_stdfloat* vz,
+      PN_stdfloat* dx, PN_stdfloat* dy, PN_stdfloat* dz);
```
rdb, line 101: You changed the signature of a virtual method, but that requires changing the signature of all implementations of this method in derived classes.  As-is this breaks FMOD etc.

get_3d_attributes either needs an overload with fewer args for compatibility or you need to default these to nullptr with an extra check in the method to not write them if they're nullptr.

It may be less work to just make a separate set/get_3d_direction if you want.  You can also make it accept a direction vector if you do that rather than 3 separate args so you can handle the LVector3::forward default more cleanly.

panda/src/audiotraits/openalAudioSound.cxx
```diff
@@ -792,6 +809,81 @@ get_3d_drop_off_factor() const {
...
+set_3d_cone_outer_angle(PN_stdfloat angle) {
+    ReMutexHolder holder(OpenALAudioManager::_lock);
+    _cone_outer_angle = angle;
+
+    if (is_playing()) {
+        _manager->make_current();
+
+        alGetError(); // clear errors
+        alSourcef(_source, AL_CONE_OUTER_ANGLE, _cone_outer_angle);
+        al_audio_errcheck("alSourcefv(_source,AL_CONE_OUTER_ANGLE)");
```
rdb, line 850: We use 2 space indent

panda/src/audio/audioSound.h
```diff
@@ -19,6 +19,7 @@
 #include "typedReferenceCount.h"
 #include "pointerTo.h"
 #include "filterProperties.h"
+#include "nodePath.h"
```
rdb, line 22: Included but not used

rdb (changes requested):
Thanks.  Will OK the check run when you address the nitpicks.

rdb:
Also, please check your configured user.name and user.email in git.  Looks like the commits are showing up as "user1", and not linked to your GitHub account.

panda/src/audio/audioSound.h
git2323: I added nodePath.h so that I can use LVector3. Is there a better way?

rdb: Yes, include `lvector3.h`, or `luse.h` to pull in all linmath classes. :)

git2323:
Thanks for the review and the explanations. I adjusted the code.

rdb (changes requested):
Not all feedback was addressed.

The signature of a virtual method was changed without all of the implementations being changed to match.  As-is, this breaks 3D sound in FMOD.  Either move the direction setting to a separate method, or overload it, or fix the FMOD implementation to match the method signature.

Furthermore, while in principle I like using the vector type, I don't like that it's now inconsistent within the same method.  I would suggest just overloading the method to accept a point/vector for all three parameters, and leaving the existing method alone.

Please configure your editor to use 2 spaces of indentation for C++ sources.

git2323:
Thanks for the review. Now I left the old methods as they were, and added two new ones for the direction. I guess the initial intention to extend the set_3d_attributes method instead of adding another one was to reduce the number of function calls in the scene update function, but maybe that is negligible.

panda/src/audio/audioSound.h
```diff
@@ -88,15 +89,16 @@ class EXPCL_PANDA_AUDIO AudioSound : public TypedReferenceCount {
...
-  virtual void set_3d_attributes(PN_stdfloat px, PN_stdfloat py, PN_stdfloat pz,
-                                 PN_stdfloat vx, PN_stdfloat vy, PN_stdfloat vz);
-  virtual void get_3d_attributes(PN_stdfloat *px, PN_stdfloat *py, PN_stdfloat *pz,
-                                 PN_stdfloat *vx, PN_stdfloat *vy, PN_stdfloat *vz);
+  virtual void set_3d_attributes(PN_stdfloat px, PN_stdfloat py, PN_stdfloat pz, PN_stdfloat vx, PN_stdfloat vy, PN_stdfloat vz);
+  virtual void get_3d_attributes(PN_stdfloat *px, PN_stdfloat *py, PN_stdfloat *pz, PN_stdfloat *vx, PN_stdfloat *vy, PN_stdfloat *vz);
 
+  // Controls the direction of this sound emitter. Currently implemented only for OpenAL.
+  virtual void set_3d_direction(LVector3 d);
+  virtual void get_3d_direction(LVector3 *d);
```
rdb, line 101: I would just make this a return value, and mark it as `const`, ie. `virtual LVector3 get_3d_direction() const;`
Not sure why the other method was designed so strangely.

rdb (changes requested):
Looks good.  But there's no reason for get_3d_direction not to just return the vector rather than using an out argument (which is unpythonic), and there's still the issue of the indentation in audioSound.cxx.

Have you tested that it works?

git2323:
Intendation has been adjusted, and get_3d_direction() returns a LVector3 now. I have tested it and it worked.

rdb (approved):
Looks great!  Thanks a lot!

rdb:
Whoops!  I did not notice that this was targeting 1.10!

This really can't go into 1.10 due to ABI incompatibility, I am going to manually cherry-pick this to master

PR #1454, opened 2023-02-10
direct: Cut down on usage of `import *`

WMOkiishi:
## Issue description
Many of the modules in the `direct` package use `from <module> import *`. This is generally considered bad practice, as it hides which names are imported and can lead to issues when the module being imported from is updated. In addition to complicating debugging, it also makes analyzing the module more difficult for external tools: I use mypy's `stubtest` to check the accuracy of the Panda3D type stubs, and I have to tell it to ignore lots of names that it can't tell are imported.

## Solution description
The approach taken here is to simply replace importing `*` with the names the module actually needs to use. In some cases, the `*` imports were being used to import items through a module other than their source; in these cases, I've changed the import to instead pull directly from where the item is defined.

## Checklist


direct/src/directtools/DirectManipulation.py
```diff
@@ -1,10 +1,48 @@
...
+    SKIP_BACKFACE,
+    SKIP_CAMERA,
+    SKIP_HIDDEN,
+    SKIP_UNPICKABLE,
+    SKIP_WIDGET,
+    X_AXIS,
+    Y_AXIS,
+    Z_AXIS,
+    LE_showInOneCam,
+)
```
rdb, line 33: This is a bit awkward, can we just prefix it, like `from . import DirectGlobals as DG`?

direct/src/directbase/TestStart.py
```diff
@@ -1,8 +1,5 @@
 print('TestStart: Starting up test environment.')
 
-from panda3d.core import *
-
-from direct.showbase.PythonUtil import *
```
rdb, line 5: PythonUtil puts things in builtins, that's why they are there in these legacy scripts - no need to break that usage

direct/src/tkpanels/MopathRecorder.py
```diff
@@ -3,23 +3,60 @@
...
+    BooleanVar,
+    Button,
+    Checkbutton,
+    Entry,
+    Frame,
+    Label,
+    Radiobutton,
+    Scale,
+    StringVar,
+)
```
rdb, line 59: Let's just use `import tkinter` here and `tkinter.X` for at least the constants, possibly for everything

rdb (changes requested):
Thanks for doing this work!

direct/src/directbase/TestStart.py
rdb: I guess we can just turn this into `import direct.showbase.PythonUtil`

rdb (approved):
Great, thanks

We don't have unit tests for many of these modules but I trust that people will send us feedback if this breaks anything for whatever odd reason

PR #1457, opened 2023-02-12
python: Use public PyErr_Occurred API instead of private macro

darktohka:
## Issue description

The private `_PyErr_OCCURRED` macro was removed in Python 3.10.0 alpha 5: https://bugs.python.org/issue43270

While building Panda3D for Python 3.12.0 alpha 5, I noticed that Panda3D's `py_compat.h` header was erroneously defining `_PyErr_OCCURRED` as:

```cpp
PyThreadState_GET()->curexc_type
```

However, Python 3.12.0 alpha 5 no longer has a  `curexc_type` field as part of the `PyThreadState` struct. This causes the build to fail.

Problem 1: `py_compat.h` should not be redefining the `_PyErr_OCCURRED` macro for all Python versions. The invalid macro definition only exists prior to Python 3.3.3. It comes as a surprise that it is being redefined in all subsequent Python versions also. This can lead to issues if newer Python versions have a different implementation of the macro.

Problem 2: `_PyErr_OCCURRED` no longer exists in versions after Python 3.10.0 alpha 5. Therefore, I believe that we should switch to using the public `PyErr_Occurred` method instead, which does not require any compatibility headers as it is available in all Python 3 versions.

Problem 3: `_PyErr_OCCURRED` is being used to compare exception types in multiple locations. This may not work in all cases according to the Python manual:

> Do not compare the return value to a specific exception; use [PyErr_ExceptionMatches()](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_ExceptionMatches) instead, shown below. (The comparison could easily fail since the exception may be an instance instead of a class, in the case of a class exception, or it may be a subclass of the expected exception.)

## Solution description
Problem 1 is solved by removing the compatibility headers, which are no longer required.

Problem 2 is solved by switching all `_PyErr_OCCURRED` references to `PyErr_Occurred` calls.

Problem 3 is solved by replacing the relevant `PyErr_Occurred` calls to `PyErr_ExceptionMatches` calls.

## Checklist
* [X]  …the intent of this change is clearly explained

### DIFF
--- a/dtool/src/interrogate/interfaceMakerPythonNative.cxx
+++ b/dtool/src/interrogate/interfaceMakerPythonNative.cxx
@@ -1917,7 +1917,7 @@ write_module_class(ostream &out, Object *obj) {
           write_function_forset(out, def._remaps, 0, 0, expected_params, 2, true, true,
                                 AT_no_args, return_flags, false);
 
-          out << "  if (!_PyErr_OCCURRED()) {\n";
+          out << "  if (!PyErr_Occurred()) {\n";
           out << "    return Dtool_Raise_BadArgumentsError(\n";
           output_quoted(out, 6, expected_params);
           out << ");\n";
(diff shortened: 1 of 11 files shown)


rdb (approved):
OK.  It's a shame that we end up having to make an extra call to the Python C runtime, because we check the Python exception state a *lot* (IIRC in almost every wrapper function).  But it's not possible to avoid at at this point as even checking the new `current_exception` field requires getting the thread state, which is nowadays locked behind a function call as well.

I don't think we need to worry about using PyErr_ExceptionMatches.  Where we're comparing to an exception type we're using it to check specific exceptions generated by the Python's C runtime, not by Python code.  These won't be generating heap-type exception instances or chained exceptions, and where we use it we don't care about sub-classed exceptions.  But I don't think it matters at this point as we can't avoid calling into the runtime to check the exception state anyway - and I'm hoping the fast-path for PyErr_ExceptionMatches is inlined anyway.

PR #1473, opened 2023-02-28
dcparser: Updated out-of-panda dc_sort_inheritance_by_file variable to match PRC default

ksmit799:
## Issue description
Using dcparser out-of-panda with multiple .DC files results in a compatibility error as Panda sorts inheritance by file by default. In particular, field numbers do not match.

## Solution description
See dcClass.cxx PRC variable for dc_sort_inheritance_by_file (defaults to true)

## Checklist

### DIFF
--- a/direct/src/dcparser/dcClass.h
+++ b/direct/src/dcparser/dcClass.h
@@ -32,7 +32,7 @@ extern ConfigVariableBool dc_sort_inheritance_by_file;
 
 static const bool dc_multiple_inheritance = true;
 static const bool dc_virtual_inheritance = true;
-static const bool dc_sort_inheritance_by_file = false;
+static const bool dc_sort_inheritance_by_file = true;
 
 #endif  // WITHIN_PANDA
 


rdb:
The description of the variable says this is a "temporary hack", and the past tense "sorted" suggests that the intended new behaviour is for it to be disabled, not enabled?  See 14374eef99583ce9f504c78ebc40c07a472f7c8f for the relevant change.  A cursory look at this change suggests that *not* sorting is the "new" behaviour.

> This is a temporary hack.  This should be true if you are using version 1.42 of the otp_server.exe binary, which sorted inherited fields based on the order of the classes within the DC file, rather than based on the order in which the references are made within the class.

Perhaps @drwrose might remember?

darktohka:
This is a tricky issue. In my opinion, we should get rid of `dc_sort_inheritance_by_file` in the first place. Unfortunately, there is a mismatch between the behavior of `dc_sort_inheritance_by_file` and the most commonly used OTP implementation's, Astron's behavior.

Here's a collection of "cases" that the setting has an effect on:

1. https://github.com/panda3d/panda3d/blob/master/direct/src/dcparser/dcClass.cxx#L505

In Panda3D, unnamed fields are only inherited when `dc_sort_inheritance_by_file` is set to false.

In Astron, unnamed fields are always inherited (they aren't even checked): https://github.com/Astron/Astron/blob/master/src/dclass/dc/Class.cpp#L46

2. In both Panda3D and Astron, inherited fields are sorted: https://github.com/panda3d/panda3d/blob/master/direct/src/dcparser/dcClass.cxx#L545 and https://github.com/Astron/Astron/blob/master/src/dclass/dc/Class.cpp#L173. In Astron, this sorting behaviour is achieved by always inserting the inserted field into the correct position, rather than sorting it at the end.

3. In Panda3D, if `dc_sort_inheritance_by_file` is set to `false`, field indices are not even set properly... what?

https://github.com/panda3d/panda3d/blob/master/direct/src/dcparser/dcClass.cxx#L608

I believe this is erroneous behavior. If `dc_sort_inheritance_by_file` is not set, then... fields don't even have an index set. 

4. In Panda3D, `dc_sort_inheritance_by_file` influences the DC hash. If `dc_sort_inheritance_by_file` is set to False, it generates a different hash than Astron does, which always hashes as if `dc_sort_inheritance_by_file` was set:

https://github.com/panda3d/panda3d/blob/master/direct/src/dcparser/dcFile.cxx#L432
https://github.com/Astron/Astron/blob/master/src/dclass/file/hash_legacy.cpp#L62

In my opinion, we should remove `dc_sort_inheritance_by_file`. `otp_server.exe` is definitely no longer being used, it's not available anywhere on the internet, and if it is, nobody has complained about it not working due to the PRC default anymore.

Case 1: I believe unnamed fields should always be inherited. In fact, we shouldn't even check if it's unnamed or not. This is consistent with Astron.

Case 2:  I think it's fine to keep the fields sorted. If we remove this functionality, we would also have to update Astron.

Case 3: I think we should fix this and this should definitely not rely on the variable.

Case 4: DC hash should be always generated as if `dc_sort_inheritance_by_file` is set (i.e. hash with 1 rather than 2)

rdb:
I'll trust your judgement on this, feel free to file PRs or open issues with the suggested changes.

rdb:
Okay, sorry for the long wait on this one. I've been sitting on the decision because it feels like standardizing a hack that kept old behaviour for compatibility, rather than adopting David's intended design.

But in fact, the hack is already a de facto standard and David's new behaviour is broken as @darktohka pointed out, so I think I will merge it for now.

That said, I would love for someone to revisit this at some point and tackle @darktohka's suggestions above.

PR #1476, opened 2023-03-13
Get direct to pass a mypy check

WMOkiishi:
## Issue description
[Mypyc](https://mypyc.readthedocs.io/) is a powerful optimization tool for Python. The first step toward utilizing it for Panda3D is to have `direct` pass a mypy check. This will pave the way toward future optimizations through type annotations and type stubs.

## Solution description
As mypy doesn't check the bodies of unannotated functions, getting `direct` to pass only requires a few changes here and there. This PR also adds a mypy check to the CI workflow, ensuring future changes maintain type safety (and thus the ability to be compiled with mypyc). This is done on a temporary copy of the directory because mypy doesn't work with `direct/src` layout used in the repository.

Also of note is the fact that the annotations use the latest typing syntax. This doesn't break using `direct` on older version of Python because `from __future__ import annotations` is used in all the files where the type syntax could be problematic. While attempting to _evaluate_ these annotations on older versions of Python still won't work, this isn't a backward compatibility concern because there previously weren't any annotations to evaluate.

## Checklist


direct/src/distributed/AsyncRequest.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
rdb, line 1: Not OK, requires Python 3.7 - would rather have a different solution even if we have to just use Any for this particular annotation

direct/src/distributed/CRDataCache.py
```diff
@@ -81,11 +81,11 @@ def flush(self):
     dc._startMemLeakCheck()
 
     cd = CachedDOData()
-    cd.foo = 34
+    cd.foo = 34  # type: ignore[attr-defined]
```
rdb, line 84: Is there no way to mark CachedDOData as a class that stores arbitrary attributes?  I assume this will fail with mypyc.

mypy.ini
```diff
@@ -0,0 +1,13 @@
+[mypy]
+mypy_path = built
```
rdb, line 2: I don't like having a configurable build directory name hardcoded in the mypy config file

direct/src/tkpanels/ParticlePanel.py
```diff
@@ -2920,14 +2920,14 @@ def setRadius(radius, f = force):
 # Create demo in root window for testing.
 if __name__ == '__main__':
     try:
-        base
+        base  # type: ignore[has-type, used-before-def]
```
rdb, line 2923: Maybe we can just clean this up by importing ShowBaseGlobal and checking if it has a base attribute

direct/src/directdevices/DirectJoybox.py
```diff
@@ -3,6 +3,7 @@
 from .DirectDeviceManager import ANALOG_DEADBAND, ANALOG_MAX, ANALOG_MIN, DirectDeviceManager
 from direct.directtools.DirectUtil import CLAMP
 from direct.gui import OnscreenText
+from direct.showbase.ShowBaseGlobal import base
```
rdb, line 6: This will fail, because ShowBaseGlobal doesn't have a `base` member until ShowBase is started.  Better to import ShowBaseGlobal and then in the code use `ShowBaseGlobal.base`.

direct/src/distributed/ClientRepository.py
```diff
@@ -3,7 +3,7 @@
 from .ClientRepositoryBase import ClientRepositoryBase
 from direct.directnotify import DirectNotifyGlobal
 from direct.showbase.MessengerGlobal import messenger
-from .MsgTypesCMU import (
+from .MsgTypesCMU import (  # type: ignore[attr-defined]
     CLIENT_DISCONNECT_CMU,
     CLIENT_HEARTBEAT_CMU,
     CLIENT_OBJECT_UPDATE_FIELD_TARGETED_CMU,
```
rdb, line 9: ugh - maybe we ought to just refactor MsgTypesCMU not to play games with the globals?

direct/src/showbase/TkGlobal.py
```diff
@@ -31,7 +31,7 @@ def bordercolors(root, colorName):
         '#%04x%04x%04x' % (darkRGB[0], darkRGB[1], darkRGB[2])
     )
 
-Pmw.Color.bordercolors = bordercolors
+Pmw.Color.bordercolors = bordercolors  # type: ignore[attr-defined]
```
rdb, line 34: What's with all the ignores for Pmw - can't we get mypy to recognise Pmw members somehow?  Or just blanket ignore all of Pmw in a config file?

rdb (changes requested):
Unfortunately we can't use `from __future__ import annotations` and maintain Python 3.6 support

direct/src/showbase/TkGlobal.py
WMOkiishi: I did that, but for some reason the errors still pop up on macOS. I have no idea what's going on with that; it's probably worth digging into.

direct/src/distributed/ClientRepository.py
WMOkiishi: Probably. I can include that in this PR or make a new one.

direct/src/directdevices/DirectJoybox.py
WMOkiishi: It actually fails already—`DirectJoybox` can't be imported without creating a `ShowBase`. I can still make the changes, though.

direct/src/distributed/CRDataCache.py
WMOkiishi: Ah, I hadn't realized that was the entire point of `CachedDOData`. Yes, that should be possible.

WMOkiishi:
> Unfortunately we can't use `from __future__ import annotations` and maintain Python 3.6 support

Hmm... okay, we can work around that. I was mostly using it to avoid having to use `typing.Union` and `typing.Dict` and the like. I'll refactor those cases.

ksmit799:
> Unfortunately we can't use `from __future__ import annotations` and maintain Python 3.6 support

Python 3.6 is EOL, is it still going to be supported for future releases? I.e. 1.11

BMaxV:
We had a discussion six months ago where we talked about doing type hints and type checking in panda3d, where I was against and you were for it.

I don't think that was "resolved".

You already put a lot of work into this PR, so maybe we can continue that discussion and you can explain what this actually improves, with this hands-on example?

WMOkiishi:
@BMaxV That is correct. Recently, though, @rdb approached me with the prospect of upstreaming the Panda3D type stubs so that Panda apps could be compiled with mypyc. He'd probably be able to explain the rationale better, but I'd imagine "it's fast" is the main benefit 🙂

rdb:
The prospect of using Panda3D with mypyc (especially in the context of compiling for web) changed my mind to be inclined towards accepting non-invasive type hints.  Non-invasive means: no extra dependencies, change of behaviour, clunky hacks, messy code, or imports of modules that would not have been made otherwise (I suppose `typing` will have to be an exception to that rule).  And the type hints need to be validated by CI.

I have also changed my mind on needing to support Python 3.6, as it is officially EOL.  You can even use Python 3.8+-only features if you so wish, but in that case the PR will not be merged until Python 3.7 has gone EOL (June 28th, less than two months from now).

direct/src/distributed/MsgTypesCMU.py
```diff
@@ -5,25 +5,22 @@
...
+CLIENT_SET_INTEREST_CMU                 = 9009
+OBJECT_SET_ZONE_CMU                     = 9010
+CLIENT_HEARTBEAT_CMU                    = 9011
+CLIENT_OBJECT_UPDATE_FIELD_TARGETED_CMU  = 9011
 
-    'CLIENT_OBJECT_UPDATE_FIELD' : 120,  # Matches MsgTypes.CLIENT_OBJECT_SET_FIELD
-}
+CLIENT_OBJECT_UPDATE_FIELD = 120  # Matches MsgTypes.CLIENT_OBJECT_SET_FIELD
+
+MsgName2Id = {name: value for name, value in globals().items() if isinstance(value, int)}
```
rdb, line 23: Good solution, I like :+1:

WMOkiishi:
> I have also changed my mind on needing to support Python 3.6, as it is officially EOL.

Excellent. I'll update the PR.

> You can even use Python 3.8+-only features if you so wish, but in that case the PR will not be merged until Python 3.7 has gone EOL


I don't think any 3.8+ features are needed here, but I'll keep that in mind for future typing PRs.

rdb (changes requested):
Needs new rebase/merge.

PR #1480, opened 2023-03-21
direct: Extract test code to test modules

WMOkiishi:
## Issue description
The `direct` package doesn't have very good test coverage. Some of its modules, though, do have some test code buried within them.

## Solution description
These changes simply move the test code scattered across `direct` into proper test modules. While some of them might not be very good tests, they're better than nothing. At the very least, they might make good starting points for proper unit tests in the future.

## Checklist


rdb (approved):
Great!

Issue #1481, opened 2023-03-21
tinydisplay not built by CMake


rdb:
When building with CMake, the tinydisplay module is not built.

A CMakeLists.txt file needs to be created for the tinydisplay directory.

rdb:
Please do @toadkarter!  Feel free to join #engine-dev on our [Discord server](https://discord.gg/UyepRMm) if you have any questions.

pmp-p:
i think static build does not pass in all cmake cases
narrowed it down to
```
cmake  -DHAVE_EGL=NO -DHAVE_GL=NO -DHAVE_GLX=NO -DHAVE_X11=YES -DHAVE_GLES1=NO -DHAVE_GLES2=NO -DHAVE_TINYDISPLAY=1
```
where the default (and only) renderer whould be offscreen tinydisplay

PR #1487, opened 2023-04-02
Issue #1481: Add CMake for TinyDisplay module

toadkarter:
## Issue description
This change is intended to fix Issue #1481 by adding a CMake file for TinyDisplay module. The intent is that this module is built when a user compiles using CMake.

## Solution description
A new CMakeLists.txt file is created in the TinyDisplay directory in the same style as the other modules. The core CMakeLists.txt file is also updated to read the TinyDisplay subdirectory.

I understand that not all headers from TinyDisplay are necessary for inclusion. I have modelled the list of headers and sources to be included from the equivalent `premake` file: https://github.com/panda3d/panda3d/blob/release/1.9.x/panda/src/tinydisplay/Sources.pp

## Checklist

### DIFF
--- a/panda/src/tinydisplay/CMakeLists.txt
+++ b/panda/src/tinydisplay/CMakeLists.txt
@@ -1,71 +1,60 @@
 set(P3TINYDISPLAY_HEADERS
-    config_tinydisplay.h
-    srgb_tables.h
-    store_pixel.h
-    store_pixel_code.h
-    store_pixel_table.h
-    tinyCocoaGraphicsPipe.I tinyCocoaGraphicsPipe.h
-    tinyCocoaGraphicsWindow.I tinyCocoaGraphicsWindow.h
-    tinyGeomMunger.I tinyGeomMunger.h
-    tinyGraphicsBuffer.I tinyGraphicsBuffer.h
-    tinyGraphicsStateGuardian.I tinyGraphicsStateGuardian.h
-    tinyOffscreenGraphicsPipe.I tinyOffscreenGraphicsPipe.h
-    tinySDLGraphicsPipe.I tinySDLGraphicsPipe.h
-    tinySDLGraphicsWindow.I tinySDLGraphicsWindow.h
-    tinyTextureContext.I tinyTextureContext.h
-    tinyWinGraphicsPipe.I tinyWinGraphicsPipe.h
-    tinyWinGraphicsWindow.I tinyWinGraphicsWindow.h
-    tinyXGraphicsWindow.I tinyXGraphicsWindow.h
-    zbuffer.h
-    zfeatures.h
-    zgl.h
-    zline.h
-    zmath.h
-    ztriangle.h
-    ztriangle_code_1.h
-    ztriangle_code_2.h
-    ztriangle_code_3.h
-    ztriangle_code_4.h
-    ztriangle_table.h
-    ztriangle_two.h
+  config_tinydisplay.h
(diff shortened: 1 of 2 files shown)


Moguri:
Merging in master to ru-run stuck CI jobs. Feel free to rebase if you'd prefer.

Moguri:
Looks like there are still some build errors that need to be resolved (see failing CI jobs). If you have questions -- or would otherwise like help addressing the failures -- please reach out (either here or on Discord).

toadkarter:
> Looks like there are still some build errors that need to be resolved (see failing CI jobs). If you have questions -- or would otherwise like help addressing the failures -- please reach out (either here or on Discord).

Apologies, I had completely missed the original notification showing the the original CI jobs were being re-run! 

I just returned to this task here - it seems to be compiling here for me locally, but I think what is happening is that the CI environment might not have X11 installed (which I do) which is causing the issue. I will add some logic to make sure that we're not linking to libraries that aren't available depending on what the user has installed.

panda/src/tinydisplay/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+        tinyTextureContext.cxx
+        tinyWinGraphicsPipe.cxx
+        tinyWinGraphicsWindow.cxx
+        tinyXGraphicsPipe.cxx
+        tinyXGraphicsWindow.cxx
+        vertex.cxx
+        zbuffer.cxx
+        zdither.cxx
+        zline.cxx
+        zmath.cxx
```
pmp-p, line 57: maybe missing 
  ztriangle_table.cxx
  ztriangle_1.cxx
  ztriangle_2.cxx
  ztriangle_3.cxx
  ztriangle_4.cxx
  srgb_tables.cxx

some these symbols defined in there are not found when linking static

panda/src/tinydisplay/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+)
+
+composite_sources(p3tinydisplay P3TINYDISPLAY_SOURCES)
+add_component_library(p3tinydisplay SYMBOL BUILDING_TINYDISPLAY ${P3TINYDISPLAY_HEADERS} ${P3TINYDISPLAY_SOURCES})
+target_link_libraries(p3tinydisplay panda p3windisplay p3x11display p3cocoadisplay)
```
pmp-p, line 62: why should a display be bound to other display ?
is not the point of tinydisplay is to be able to have both headless(eg offscreen) gpuless(eg wasm) capacity ?

pmp-p (commented):
i tried a wasi-sdk cmake build with that, and some points bother me.

panda/src/tinydisplay/CMakeLists.txt
rdb: p3windisplay, p3x11display, p3cocoadisplay are used to supply the OS-specific window creation code.  Doesn't preclude use with headless/wasm.

panda/src/tinydisplay/CMakeLists.txt
rdb: Yes, they are missing, and they should also be marked as not participating in the unity build system due to their size.

rdb (changes requested):
Is this tested to run?  I don't understand how it could without the ztriangle code.

toadkarter:
Hey folks, my apologies but unfortunately work IRL has been very busy so I don't think I will be able to work on this. I am going to close down this pull request, apologies again.

Issue #1491, opened 2023-04-22
reducing complexity, then test coverage via cyclomatic complexity


BMaxV:
## Description
So, I found this tool that I like

https://pypi.org/project/radon/

It measures complexity. It can help us identify and prioritize pain points prioritize them.

It counts the "paths" through a function, and every "if" increases the number of paths by 1. 

I wrote a little crawler that looks at a project/repo/tree and makes it convenient to look at the thing as a whole.

The idea would be to encapsulate small parts of problematic functions. We could then write tests for those now small and easy to understand functions.

## Use Case

# how to

A being good, F being the worst and very bad, all of panda3d currently looks like this:

![piechart](https://user-images.githubusercontent.com/5713211/233784796-0215e972-3bc3-4296-9335-338bc0bf7791.svg)

And it helps navigating to e.g. `/direct/src/actor/Actor.py` which looks like this:

![radonpieActor](https://user-images.githubusercontent.com/5713211/233784894-e5286602-a2bd-4a98-8290-d0b48d8dc4e8.svg)

```
Actor.py
    M 1773:4 Actor.getAnimControls - E
    M 130:4 Actor.__init__ - D
    M 1899:4 Actor.loadModel - C
    M 2144:4 Actor.loadAnims - C
```

getAnimControls

https://github.com/panda3d/panda3d/blob/master/direct/src/actor/Actor.py#L1773

This block could be wrapped into a function

`actor.getAnimControlDictItems(lodName) -> partDict, animControlDictItems`

```
if lodName is None or self.mergeLODBundles:
    # Get all LOD's
    animControlDictItems = self.__animControlDict.items()
else:
    partDict = self.__animControlDict.get(lodName)
    if partDict is None:
        Actor.notify.warning("couldn't find lod: %s" % (lodName))
        animControlDictItems = []
    else:
        animControlDictItems = [(lodName, partDict)]
```

I think this would:
 * create small, easy to review pull requests. 
 * good first issues, since it's not required to understand how panda words, all that's necessary is making sure all required variables are passed to the function and all assigned variables are returned.
 * writing tests should be easier if the function that needs to be tested is small

I would really like to get in the habit of contributing small improvements like this. So it would be cool if we could figure something out to make that happen @Moguri @rdb

rdb:
[...]

PR #1498, opened 2023-05-07
Cyclomatic complexity improvement and tests for actor

BMaxV:
## Issue description

This is related to #1491

It's up with the goal of having smaller functions that are easier to test.

## Solution description

I simply took some blocks and put them into their own functions.

## Checklist

The code is not tested, I think, but I can write tests for new functions. At least for those that I can guess at what they do, I haven't really used LODs in animations or rigs yet.

### DIFF
--- a/direct/src/actor/Actor.py
+++ b/direct/src/actor/Actor.py
@@ -1878,7 +1878,7 @@ class Actor(DirectObject, NodePath):
             animDictItems = self.getAnimDictItems(partName,lodName,partDict)
         
             if animName is None:
-                controls = self.getPlayingAnimationControls(animDictItems)
+                controls += self.getPlayingAnimationControls(animDictItems)
             
             else:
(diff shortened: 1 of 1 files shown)


rdb:
I appreciate the contribution, but I'm inclined to reject this for the reasons described in #1491. As-is there is no unit test, it increases the surface area of the API (especially since the extra methods are public), the code doesn't seem meaningfully more readable or maintainable to me with the change, and I simply don't think the perceived benefits outweigh the risks of touching the code.

BMaxV:
As I have previously said, the animation system is "broken" in the sense that I can't reproduce the behavior the manual says should work.

Please consider how we could get to the point that the animation system works as document or to document how the current code works. I don't think it can be done without rewriting the code and/or adding tests.

The purpose of writing smaller functions is that they are smaller pieces of functionality that are easier to test.

PR #1503, opened 2023-05-30
Add wrapper around connect error codes when using LibreSSL on Windows

Hawkheart:
## Issue description

When using the native WinSock2 sockets with LibreSSL's implementation of the BIO library, it's necessary to wrap the returned error codes so that LibreSSL knows how to handle it. In particular, I observed the `connect` function's "try again later" error code returned when using non-blocking IO gets interpreted as a fatal error if this is not done.

## Solution description

I've borrowed the approach used by the LibreSSL example programs and just set `errno` to an appropriate value depending on the socket error. In my testing, this is good enough for my game client to successfully connect to its server.

## Checklist


rdb:
I don't quite understand this.  If an error occurs then Panda uses `format_error()`, which doesn't use `errno`, but calls `WSAGetLastError()`.  It therefore shouldn't be necessary to translate any error number.

I first want to understand what is going wrong.  Is something outside of `BioPtr::connect()` reading `errno` and acting on that?  Or is LibreSSL's `BIO_sock_should_retry` buggy?

Hawkheart:
LibreSSL's `BIO_sock_should_retry`  (and more specifically `BIO_sock_non_fatal_error`) treats the Windows `WSAEWOULDBLOCK` error code as a fatal error. They've stripped out any platform specific code in that function in favor of hardcoding the error codes OpenBSD would return.

rdb:
I'd rather just replace the call to `BIO_sock_should_retry` with something that checks the WSAGetLastError code directly against the codes that `BIO_sock_should_retry` is known to return 1 with in official OpenSSL.

Only in the LibreSSL case, of course.

Hawkheart:
That makes sense - I'll rewrite this to do something like that. I'd copied the approach from the LibreSSL sample programs, but in hindsight they do it like this to avoid having to make any modifications to their app source code in the portable distribution.

rdb (changes requested):
see comments

Hawkheart:
Pushed a version with the modified approach; I don't have access to my test environments to check if this works at the moment so it's not ready for a merge.

Hawkheart:
Alright, I've double checked and this patch works properly in my engine build (across Windows/Mac/Linux, all with LibreSSL)

Moguri:
This looks good to me, and I believe the requested changes have been made.

rdb (approved):
This is mostly good other than some indent issues, the use of `and` instead of `&&`, and the fact that this function should be `static`, but I've fixed those when I committed it.

Thanks!

PR #1514, opened 2023-06-29
Mark more constructors as `explicit`

### LINKED ISSUE #1431: Marking more constructors as `explicit`
C++ constructors not marked `explicit` that can be called with a single parameter allow for implicit conversion (aka coercion) of the parameter type to the type to which the constructor belongs. Panda3D extends this behavior to its Python bindings. This feature is sometimes useful, but sometimes masks bugs. I have compiled a list of coercions supported by these "converting constructors" whose value seems dubious:

- Coercing a matrix (of either size) to an `LRotation` or `LOrientation`, both of which are `LVecBase4` subclasses
- Coercing a string to an `HTTPDate`
- Coercing a `TypeHandle` to a `PointerToArray`
- Coercing a string to a `MovieVideo`
- Coercing a `Thread` to a `GeomVertexReader`, `GeomVertexWriter`, or `GeomVertexRewriter`
- Coercing a string to a `Spotlight`
- Coercing an integer to an `PNMImageHeader::PixelSpec`
- Coercing a float to a `BulletTickCallbackData`
- Coercing an `OdeWorld` to various ODE classes
- Coercing a string to a `ConfigVariableColor`, `ConfigVariableList`, or `ConfigVariableString` (These seem to have been missed when some other `ConfigVariable` constructors were marked `explicit`.)
- Coercing a 3-vector to a 4-vector (These might have value, but I've decided to list it for completeness.)
- Coercing a single to number to a vector of any size

I would suggest considering marking the constructors supporting these coercions as `explicit`. There are almost certainly more, but these were the ones that stood out to me from the list I generated with `interrogatedb`.
Additionally, constructors not marked `explicit` that require multiple parameters allow for implicit conversion from a sequence of values. This is extended to Python as coercion from a tuple. I suspect that the only constructors requiring multiple parameters that _shouldn't_ be marked explicit are those converting a sequence of numbers to a vector or a sequence of vectors to a matrix.

Of course, there would be some backwards-compatibility concerns. I suspect that, for most of these coercions, most use "in the wild" is accidental. If any of them are being used deliberately, the workaround is simply converting the value explicitly before passing it to a function.

WMOkiishi:
## Issue description
Resolves #1431. See that issue for details.

## Solution description
This marks the constructors responsible for the coercions listed in the above issue as `explicit`, preventing implicit conversion. It also marks a number of constructors that take multiple arguments as `explicit`, preventing coercion from a sequence of arguments.

## Checklist


WMOkiishi:
It looks like the failing tests are due to something external. Is it possible to run them again?

rdb (approved):
All seems good, thanks.

Seems that there are issues with the clang-tidy reviewer since there are no compilation rules for some headers that are included indirectly. Not sure how to fix that right now.

PR #1526, opened 2023-08-05
event: Don't use _PyGen_FetchStopIterationValue in Python 3.13+

darktohka:
## Issue description
Python 3.13 has unpublished the _PyGen_FetchStopIterationValue private C API call. It is now part of the internal API: https://github.com/python/cpython/issues/106320

## Solution description
This pull request replaces the private _PyGen_FetchStopIterationValue API call with the new PyErr_GetRaisedException API call, added in Python 3.12.

Because the versioning is so complicated (`PyErr_GetRaisedException` being added in Python 3.12, `_PyGen_FetchStopIterationValue` not working in Python 3.13, older versions need to be supported still) I have decided to do the following:
- Keep the old behaviour on versions older than Python 3.13, as it works perfectly fine
- Use `PyErr_GetRaisedException` on Python 3.13+

Right now this code is not covered by the test suite, given that Python 3.13 is still in alpha.

## Checklist

### DIFF
--- a/panda/src/event/pythonTask.cxx
+++ b/panda/src/event/pythonTask.cxx
@@ -617,9 +617,12 @@ do_python_task() {
       Py_DECREF(_generator);
       _generator = nullptr;
 
+#if PY_VERSION_HEX >= 0x030D0000 // Python 3.13
+      // Python 3.13 does not support _PyGen_FetchStopIterationValue anymore.
+      if (PyErr_ExceptionMatches(PyExc_StopIteration)) {
+#else
       if (_PyGen_FetchStopIterationValue(&result) == 0) {
-        PyErr_Clear();
-
+#endif
         if (_must_cancel) {
           // Task was cancelled right before finishing.  Make sure it is not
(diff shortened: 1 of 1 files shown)


rdb (commented):
Thanks.

Should we check for whether the stored value may be nullptr?  Right now that would result in a crash.

I agree with only enabling this in Python 3.13.  Have you run the test suite with the new code enabled on an older version just to make sure it works, though?

rdb:
Looking at this in more detail, I don't think this fix is right.  `result` also needs to be populated if `_function != nullptr`, used further down in the function, otherwise there won't be a valid DoneStatus extracted for coroutine tasks:

```python
from direct.task.Task import Task
import sys

cont = Task.cont

async def thingy(task):
    print(sys.getrefcount(cont))
    return cont

taskMgr.add(thingy)

base.run()
```

I'll check something in for 1.10.

darktohka:
Thank you! I'll let you take things from here then.

PR #1527, opened 2023-08-08
directnotify: annotate types

WMOkiishi:
## Change description
These changes add type annotations to the modules within `directnotify` to help with static typing, code analysis, and mypyc compilation.

## Checklist


direct/src/directnotify/Notifier.py
```diff
@@ -42,12 +47,12 @@ def __init__(self, name, logger=None):
...
-        self.__warning = 1
-        self.__debug = 0
-        self.__logging = 0
+        self.__info = True
+        self.__warning = True
+        self.__debug = False
+        self.__logging = False
 
-    def setServerDelta(self, delta, timezone):
+    def setServerDelta(self, delta: SupportsRound[SupportsInt], timezone: int) -> None:
```
rdb, line 55: Can we just make this a float? That's probably what's intended to be passed here. Would that prevent passing an int?

direct/src/directnotify/RotatingLog.py
```diff
@@ -83,17 +92,18 @@ def rotate(self):
             self.closed = self.file.closed
             self.mode = self.file.mode
             self.name = self.file.name
-            self.softspace = self.file.softspace
+            self.softspace = self.file.softspace  # type: ignore[attr-defined]
```
rdb, line 95: I'm going to remove this, `softspace` was only necessary for Python 2.

rdb (commented):
How can we be sure that all of this is correct?  Do we currently do static analysis or is it just runtime checks on the test suite?

Would it be reasonable to require unit tests for methods we add annotations for - at least those that aren't called indirectly in other parts of the code base?

direct/src/directnotify/Notifier.py
WMOkiishi: Annotating a parameter as a float is special-cased by PEP 484 to also accept ints, so that would work. I doubt anyone will be passing custom numeric types to this function anyway.

WMOkiishi:
Currently, annotations are checked for correctness by mypy, which checks top-level code and the bodies of all annotated functions. Outside of limitations of the Python typing system, this should mean that none of these functions run into undefined attributes or type errors. That being said, unit tests aren't a bad idea, and I wouldn't mind adding them if they're deemed necessary.

direct/src/directnotify/Notifier.py
rdb: Yes, let's just do float.

rdb:
Would it be reasonable to suggest that every annotated function needs to have a point of use, either elsewhere in the codebase (that is checked by mypy) or otherwise in a unit test?

WMOkiishi:
I believe that would make sense. Should I add unit tests for the functions annotated here that aren't used in other annotated functions?

rdb:
If you'd like to do that, that would be great!

WMOkiishi:
Okay, I believe I've done that. I'm no expert in writing tests, so please let me know if anything needs to be changed. The only methods I left annotated without tests are dunders such as `__str__` or `__del__` whose signatures are always the same.

direct/src/directnotify/DirectNotify.py
```diff
@@ -77,40 +83,42 @@ def setDconfigLevel(self, categoryName):
...
             category = self.getCategory(categoryName)
-            category.setWarning(1)
-            category.setInfo(1)
-            category.setDebug(1)
+            assert category is not None
+            category.setWarning(True)
+            category.setInfo(True)
+            category.setDebug(True)
 
     def popupControls(self, tl = None):
```
Moguri, line 123: Should this be annotated too?

direct/src/directnotify/DirectNotify.py
```diff
@@ -77,40 +83,42 @@ def setDconfigLevel(self, categoryName):
             level = 'error'
 
         category = self.getCategory(categoryName)
+        assert category is not None
```
Moguri, line 86: I see no guarantees here that `category` will not be `None`. I believe it would be better to raise an Error instead. The previous code would have already failed with an attribute error of some kind (None type object has no attribute...). Neither that now an assertion error are particularly friendly. Raising an error and telling the user that the supplied category name did not exist is much more informative, and it will still provide the necessary type-narrowing that this assert is doing.

direct/src/directnotify/DirectNotify.py
```diff
@@ -12,49 +16,51 @@ class DirectNotify:
...
         """newCategory(self, string)
         Make a new notify category named categoryName. Return new category
         if no such category exists, else return existing category
         """
         if categoryName not in self.__categories:
             self.__categories[categoryName] = Notifier.Notifier(categoryName, logger)
             self.setDconfigLevel(categoryName)
-        return self.getCategory(categoryName)
+        notifier = self.getCategory(categoryName)
+        assert notifier is not None
```
Moguri, line 60: This is funky, but I understand why it is here: we can see (via inspection of `getCategory()`) that checking `self.__categories` is sufficient to guarantee `getCategory()` will not return `None` in this case. However, the danger is that the logic of `getCategory()` *could* change and break this assumption. That said, this is pretty unlikely to happen and the assert should make the broken assumption more obvious, so I am fine with the simpler fix.

direct/src/directnotify/RotatingLog.py
```diff
@@ -8,7 +11,12 @@ class RotatingLog:
     to a new file if the prior file is too large or after a time interval.
     """
 
-    def __init__(self, path="./log_file", hourInterval=24, megabyteLimit=1024):
+    def __init__(
+        self,
+        path: str = "./log_file",
+        hourInterval: int | None = 24,
+        megabyteLimit: int | None = 1024,
+    ) -> None:
```
Moguri, line 19: Should we also add types to `self.timeInterval`, `self.timeLimit`, and `self.sizeLimit`?

Moguri (changes requested):
Leaving a partial review. I still need to go over the tests.

direct/src/directnotify/DirectNotify.py
WMOkiishi: Yeah, that makes sense. What kind of error would you suggest?

direct/src/directnotify/DirectNotify.py
WMOkiishi: It could also be changed to just `return self.__categories[categoryName]`. Would that be better?

direct/src/directnotify/DirectNotify.py
WMOkiishi: Maybe. It would require a `tkinter` import, and I'm not really sure how to test it, either.

direct/src/directnotify/RotatingLog.py
WMOkiishi: It looks like mypy and pyright are both able to correctly infer the types, as `__init__` assigns as both numeric values and `None`. So we could, but we don't need to.

direct/src/directnotify/DirectNotify.py
Moguri: Okay, we'll leave it out for now then.

direct/src/directnotify/DirectNotify.py
Moguri: Okay, we'll leave it out for now.

direct/src/directnotify/DirectNotify.py
Moguri: Even just adding an explanation to the assert would probably do: `assert category is not None, f'failed to find category: {categoryName}'`

direct/src/directnotify/DirectNotify.py
Moguri: No, we might as well continue to use self.getCategory(). I am okay with the code changes here; I was just leaving a comment. I will mark this thread as resolved.

direct/src/directnotify/RotatingLog.py
Moguri: Awesome, great to see type inference working here.

direct/src/directnotify/Notifier.py
```diff
@@ -285,7 +290,7 @@ def debugStateCall(self, obj=None, fsmMemberName='fsm',
             self.__print(string)
         return 1 # to allow assert self.notify.debugStateCall(self)
 
-    def debugCall(self, debugString=''):
+    def debugCall(self, debugString: object = '') -> int:
```
rdb, line 293: Just for my curiosity, how does annotating something with `object` differ from not annotating it?

direct/src/directnotify/RotatingLog.py
```diff
@@ -28,33 +36,34 @@ def __init__(self, path="./log_file", hourInterval=24, megabyteLimit=1024):
...
+            return True
         if self.sizeLimit is not None and self.file.tell() > self.sizeLimit:
-            return 1
-        return 0
+            return True
+        return False
 
-    def filePath(self):
+    def filePath(self) -> str:
+        assert self.sizeLimit is not None
```
rdb, line 66: You've identified an oversight in the code - the assert is improper because this case ought to be handled.  I'll check in a commit.

direct/src/directnotify/Notifier.py
Wizzerinus: The difference is that annotating as an object means none of that object's attributes can be used in the function. A more significant difference happens when a constant/variable is annotated with an object, than when the argument function is.

The correct annotation in this case though would likely be Any, which is equivalent to not having the annotation.

Wizzerinus: https://docs.python.org/3/library/typing.html#the-any-type

direct/src/directnotify/RotatingLog.py
rdb: Done in c63a92bdec8043ec883ae8afc57ecf1670325d70 - you can remove the assert.

direct/src/directnotify/Notifier.py
WMOkiishi: @Wizzerinus is correct that leaving a parameter unannotated is equivalent to annotating it with `Any`. Annotating a parameter with `object`, on the other hand, means that literally any object can be passed to the function. I've done that on purpose here, as all that's done with the argument is passing it to `str`, which can accept any object and return a string.

rdb: Ahh, cool.  Thanks for the explanation.  Not sure the annotation has added value, but setting it to `object` seems fine, at least it makes it explicit that we accept non-strings here.

rdb (commented):
Tests seem okay.  If you address the rest it's good to go.  Thanks!

rdb (approved):
Thanks!

PR #1542, opened 2023-10-08
(draft) allow to run offscreen C++ test with headless pview

pmp-p:
in case of a wasi/emsdk CI with tinydisplay

### DIFF
--- a/panda/src/testbed/pview.cxx
+++ b/panda/src/testbed/pview.cxx
@@ -28,6 +28,7 @@
 #include "asyncTaskManager.h"
 #include "asyncTask.h"
 #include "boundingSphere.h"
+#include "load_prc_file.h"
 
 using std::cerr;
 using std::endl;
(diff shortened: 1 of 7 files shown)


pmp-p:
That's not valid yet because the framebuffer is contained in the GraphicsWindow object.

to work fully it would need
1 ) select OffscreenPipe
2 ) create a OffscreenGraphicsWindow which hold the Zbuffer of requested dimensions.

rdb:
window-type is (to my knowledge) only used by ShowBase.

On the C++ end I think we need to disable the "require_window" flag. I think we can just do this if a screenshot is requested, probably no need for a separate flag.

pmp-p:
but still need the window properties to set the requested size like https://github.com/panda3d/panda3d/blob/cb4597ebb84fe7228c7039e2f54c545b63d60046/panda/src/tinydisplay/tinySDLGraphicsWindow.cxx#L347 . so why not an offscreen window ?

rdb:
I don't think we need to create a new "offscreen window" class.  Is there a reason we can't just use TinyGraphicsBuffer?  Does WindowFramework expect something that inherits from GraphicsWindow?

All we need to do is call the version of PandaFramework that allows us to specify flags like so:
```cpp
  WindowProperties props;
  get_default_window_props(props);

  // this is where you can set the buffer size if you need - maybe with a cmdline option?
  props.set_size(...);

  int flags = 0;
  if (/*screenshot flag is false*/) {
    flags = GraphicsPipe::BF_require_window;
  }

  window = panda_framework.open_window(props, flags, pipe, gsg);
```

rdb:
Obviated by latest changes

PR #1546, opened 2023-10-10
Added missing control message types

ksmit799:
## Issue description
Distributed MsgTypes is missing certain control (MD) messages, mostly used for telemetry and debugging.

## Solution description
Added the missing message types. See [Astron](https://github.com/Astron/Astron/blob/master/src/core/msgtypes.h#L39) and [Ardos](https://github.com/ksmit799/Ardos/blob/main/src/net/message_types.h#L41) for reference.

## Checklist


rdb:
I don't mind adding a harmless change like this, but does it solve a problem?  None of these symbols are referenced in the Panda source.

ksmit799:
> I don't mind adding a harmless change like this, but does it solve a problem? None of these symbols are referenced in the Panda source.

FWIW neither are many of the other messages, but they're used in (server-side) applications using Panda3D networking. This change prevents otherwise redundant re-definitions on the application side.

PR #1547, opened 2023-10-10
draft: reboot of #1481 but wasm/ci static as first goal

pmp-p:
first steps for building with both wasi-sdk and emsdk the C++ part of panda3D, and be able to use wasm pview for screenshot testing/compare in the CI.

makepanda is ok for emsdk user build, but not so great for automated build/testing

Also it is a good opportunity to explain later to cmake how to locate cpython correctly once and for all ( that will profit android/wasi/emsdk at once )


kamgha:
Related PR: #1487

panda/metalibs/panda/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ set(PANDA_LINK_TARGETS
   p3pnmimage p3pnmimagetypes p3pstatclient p3putil p3recorder p3text p3tform
 )
 
+if(HAVE_TINYDISPLAY)
+  list(APPEND PANDA_LINK_TARGETS p3tinydisplay)
+endif()
```
rdb, line 10: I don't think this should be linked into libpanda.  tinydisplay doesn't have to be built as a metalib.

panda/src/tinydisplay/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+  ztriangle_3.cxx
+  ztriangle_4.cxx
+  srgb_tables.cxx
+  zbuffer.cxx
+  zdither.cxx
+  zline.cxx
+  zmath.cxx
+)
+
+add_compile_options(-Wno-unused-but-set-variable)
```
rdb, line 71: `if(NOT MSVC)`

panda/metalibs/panda/CMakeLists.txt
pmp-p: but right now the current cmake changes in master for tinydisplay don't work
for static build 
```
CMake Error: install(EXPORT "Framework" ...) includes target "p3framework" which requires target "p3tinydisplay" that is not in any export set.
CMake Error in panda/src/framework/CMakeLists.txt:
  export called with target "p3framework" which requires target
  "p3tinydisplay" that is not in any export set.

```

rdb: If you need me to investigate this please open an issue for it describing how I can repro it

PR #1551, opened 2023-10-19
tests: Fix mypy test

WMOkiishi:
## Issue description
The mypy test (`tests/run_mypy.py`) wasn't actually returning error codes, so the GitHub action passed regardless of the output.

## Solution description
These changes fix the test script as well as some issues that were overlooked because the test was broken. These include some issues with circular imports that arose because of an import that needed to be added to define an annotation.

## Checklist

### DIFF
--- a/direct/src/showbase/ShowBaseGlobal.py
+++ b/direct/src/showbase/ShowBaseGlobal.py
@@ -17,7 +17,6 @@ __all__ = ()
 
 from .ShowBase import ShowBase, WindowControls # pylint: disable=unused-import
 from direct.directnotify.DirectNotifyGlobal import directNotify, giveNotify # pylint: disable=unused-import
-from direct.directtools import DirectSession
 from panda3d.core import VirtualFileSystem, Notify, ClockObject, PandaSystem
 from panda3d.core import ConfigPageManager, ConfigVariableManager, ConfigVariableBool
 from panda3d.core import NodePath, PGTop
@@ -66,8 +65,6 @@ aspect2d = render2d.attachNewNode(PGTop("aspect2d"))
 hidden = NodePath("hidden")
 
 loader: Loader
-
-direct: DirectSession.DirectSession
 
 # Set direct notify categories now that we have config
 directNotify.setDconfigLevels()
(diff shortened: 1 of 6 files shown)


direct/src/showbase/ShowBaseGlobal.py
```diff
@@ -11,10 +11,13 @@
 since ShowBase may not have been created yet; instead, ShowBase dynamically
 adds itself to this module's scope when instantiated."""
 
+from __future__ import annotations
+
 __all__ = ()
 
 from .ShowBase import ShowBase, WindowControls # pylint: disable=unused-import
 from direct.directnotify.DirectNotifyGlobal import directNotify, giveNotify # pylint: disable=unused-import
+from direct.directtools import DirectSession
```
rdb, line 20: Not acceptable for ShowBaseGlobal to pull in the DIRECT tools.

rdb (changes requested):

direct/src/showbase/ShowBaseGlobal.py
WMOkiishi: Hm. Would it be acceptable to get rid of that annotation, then? Or perhaps replace it with `Any`?

Alternatively, we could put the import under an `if typing.TYPE_CHECKING` guard.

rdb: Let's not use a guard like that.  If there's no way to do it well without an import (or losing the typing benefit), I think we have to just move it out of ShowBaseGlobal into a new `directtools/DirectSessionGlobal.py` module.

Moguri: Does a string annotation not work in this case? If the situation is too funky, we may be better off leaving it un-annotated. I am not too concerned above covering every corner of direct, especially since it has some bad coding practices that can make annotations difficult.

rdb: A string annotation is what I had before this change.  I assume there was a problem with that?

It doesn't really need to be in ShowBaseGlobal, and nobody relies on that since I added that only recently.  I think moving it into a new DirectSessionGlobal is possibly best if we can't do the annotation cleanly otherwise.

WMOkiishi: Yeah, the string annotation didn't work because mypy had no way of knowing what `DirectSession` was. I'll just remove it.

rdb: Okay, but then most of the code in DirectSession won't benefit from type checking.

WMOkiishi: A DirectSessionGlobal module sounds like a good idea, but I think that's probably outside the scope of this PR.

rdb: Fair.

PR #1552, opened 2023-10-20
pgui: Add more property-style accessors

WMOkiishi:
## Change description
These changes just allow for property-style access from Python to data in various `pgui` classes.

## Checklist

### DIFF
--- a/panda/src/pgui/pgScrollFrame.h
+++ b/panda/src/pgui/pgScrollFrame.h
@@ -56,20 +56,27 @@ PUBLISHED:
   INLINE const LVecBase4 &get_virtual_frame() const;
   INLINE bool has_virtual_frame() const;
   INLINE void clear_virtual_frame();
+  MAKE_PROPERTY2(virtual_frame,
+                 has_virtual_frame, get_virtual_frame,
+                 set_virtual_frame, clear_virtual_frame);
 
   INLINE void set_manage_pieces(bool manage_pieces);
   INLINE bool get_manage_pieces() const;
+  MAKE_PROPERTY(manage_pieces, get_manage_pieces, set_manage_pieces);
 
   INLINE void set_auto_hide(bool auto_hide);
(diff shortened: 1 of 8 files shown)


panda/src/pgui/pgScrollFrame.h
```diff
@@ -56,20 +56,27 @@ class EXPCL_PANDA_PGUI PGScrollFrame : public PGVirtualFrame, public PGSliderBar
...
+  MAKE_PROPERTY(manage_pieces, get_manage_pieces, set_manage_pieces);
 
   INLINE void set_auto_hide(bool auto_hide);
   INLINE bool get_auto_hide() const;
+  MAKE_PROPERTY(auto_hide, get_auto_hide, set_auto_hide);
 
   INLINE void set_horizontal_slider(PGSliderBar *horizontal_slider);
   INLINE void clear_horizontal_slider();
   INLINE PGSliderBar *get_horizontal_slider() const;
+  MAKE_PROPERTY(horizontal_slider, get_horizontal_slider, set_horizontal_slider);
```
rdb, line 74: I think it's not possible to clear the horizontal slider now via the property, right?  Maybe this will require adding a has_horizontal_slider.

WMOkiishi: I believe that's correct. As far as I understand, yeah, we'd have to add a `has_horizontal_slider` method to be able to use `MAKE_PROPERTY2` and allow for clearing using the property. There were a couple of other cases where attributes has a clear method but not a has method. I would be willing to add those to this PR if you'd like.

rdb: Actually, I think we can also just pass get_horizontal_slider as the "has" function to MAKE_PROPERTY2 since a nullptr also evaluates to false.

Please make that change, otherwise this property is not complete.

rdb (approved):

rdb:
Thanks!

PR #1555, opened 2023-10-23
some fixes for wasi build

pmp-p:
from wasi-sdk stdlib.h
```
#if defined(_XOPEN_SOURCE) || defined(_GNU_SOURCE) \
 || defined(_BSD_SOURCE)
#ifdef __wasilibc_unmodified_upstream /* WASI has no absolute paths */
char *realpath (const char *__restrict, char *__restrict);
#endif
```


pmp-p:
replaced by https://github.com/panda3d/panda3d/pull/1557

PR #1557, opened 2023-10-23
webgl-port: wasi support

pmp-p:
minimal changes to be able to build the C++ part with wasi-sdk 20 and run a wasm pview triangle test with wasmtime in a github CI

[successfully tested](https://github.com/pmp-p/panda3d/actions/runs/6648637919)  ~ 15 minutes build time with cmake 3.27.7 and wasmtime-14.0.0 as a wasi runner ( both installed with pip ) 
with a md5sum test on pview bmp output

### DIFF
--- a/dtool/Package.cmake
+++ b/dtool/Package.cmake
@@ -68,6 +68,9 @@ if(THIRDPARTY_DIRECTORY)
   elseif(CMAKE_SYSTEM_NAME STREQUAL "Android")
     set(_package_dir ${THIRDPARTY_DIRECTORY}/android-libs-${CMAKE_ANDROID_ARCH})
 
+  elseif(CMAKE_SYSTEM_NAME STREQUAL "WASI")
+    set(_package_dir ${THIRDPARTY_DIRECTORY})
+
   else()
     message(FATAL_ERROR
       "You can't use THIRDPARTY_DIRECTORY on this platform. Unset it to continue.")
--- a/dtool/src/dtoolbase/dtool_platform.h
+++ b/dtool/src/dtoolbase/dtool_platform.h
@@ -65,6 +65,9 @@
 #define DTOOL_PLATFORM "android_i386"
 #endif
 
+#elif defined(__wasi__)
+#define DTOOL_PLATFORM "wasi"
+
 #elif defined(__EMSCRIPTEN__)
 #if defined(__wasm64__)
 #define DTOOL_PLATFORM "emscripten_wasm64"
--- a/dtool/src/dtoolbase/dtool_platform.h
+++ b/dtool/src/dtoolbase/dtool_platform.h
@@ -65,8 +65,15 @@
 #define DTOOL_PLATFORM "android_i386"
 #endif
 
+#elif defined(__wasm32__)
+#define DTOOL_PLATFORM "wasi_wasm32"
+
+#elif defined(__wasm64__)
+#define DTOOL_PLATFORM "wasi_wasm64"
+
+// fallback to wasm32 with older sdk
 #elif defined(__wasi__)
-#define DTOOL_PLATFORM "wasi"
+#define DTOOL_PLATFORM "wasi_wasm32"
 
 #elif defined(__EMSCRIPTEN__)
 #if defined(__wasm64__)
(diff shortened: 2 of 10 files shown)


dtool/src/dtoolbase/dtoolbase.h
```diff
@@ -152,11 +152,11 @@ typedef struct _typeobject PyTypeObject;
 #ifdef PHAVE_TYPES_H
 #include <types.h>
 #endif // PHAVE_TYPES_H
-
+#if !defined(__wasi__)
 #ifdef PHAVE_SYS_TYPES_H
 #include <sys/types.h>
 #endif // PHAVE_SYS_TYPES_H
```
rdb, line 158: Build system should simply not set PHAVE_SYS_TYPES_H for wasi if it doesn't have it, that's what these defines are for

dtool/src/dtoolutil/executionEnvironment.cxx
```diff
@@ -62,7 +62,7 @@ extern char **environ;
 #include <sys/sysctl.h>
 #endif
 
-#if defined(IS_LINUX) || defined(IS_FREEBSD)
+#if (defined(IS_LINUX) || defined(IS_FREEBSD)) && !defined(__wasi__)
```
rdb, line 65: Neither IS_LINUX nor IS_FREEBSD should be set for WASI to begin with

dtool/src/dtoolbase/dtool_platform.h
```diff
@@ -65,6 +65,9 @@
 #define DTOOL_PLATFORM "android_i386"
 #endif
 
+#elif defined(__wasi__)
+#define DTOOL_PLATFORM "wasi"
+
```
rdb, line 70: [wasi has arch tags](https://github.com/WebAssembly/wasi-libc/blob/be1704a9568e98ae073f074f7271cae68c23d161/libc-top-half/musl/src/misc/uname.c#L18-L21) just like emscripten does, so distinguish wasi_wasm64/wasi_wasm32 for future extensibility

dtool/Package.cmake
```diff
@@ -68,6 +68,9 @@ if(THIRDPARTY_DIRECTORY)
   elseif(CMAKE_SYSTEM_NAME STREQUAL "Android")
     set(_package_dir ${THIRDPARTY_DIRECTORY}/android-libs-${CMAKE_ANDROID_ARCH})
 
+  elseif(CMAKE_SYSTEM_NAME STREQUAL "WASI")
+    set(_package_dir ${THIRDPARTY_DIRECTORY})
```
rdb, line 72: Set this to a wasi-specific subdirectory or don't support THIRDPARTY_DIRECTORY for wasi at all

rdb (changes requested):

dtool/Package.cmake
pmp-p: maybe THIRDPARTY_DIRECTORY should be CMAKE_INSTALL_PREFIX as default

dtool/src/dtoolbase/dtoolbase.h
pmp-p: indeed, got that one fixed my cross toolchain file was not robust enough and was peeking into host /usr

dtool/src/dtoolutil/executionEnvironment.cxx
pmp-p: That's just temporary later it should be
defined(IS_LINUX) || defined(IS_FREEBSD) || defined(__wasi__)
because dlfcn is on the way ( though it's via static linking for now )

dtool/Package.cmake
pmp-p: documenting from discord

https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html  could replace THIRDPARTY_DIRECTORY, and maybe default to CMAKE_INSTALL_PREFIX if not set at all.

Todo: find how cmake android-studio ( recent ndk) set/get/detect the prebuilt folder which maybe contains the per arch prebuilts from third parties.

rdb (approved):

PR #1558, opened 2023-10-27
showbase: Annotate basic `ShowBase` setup/shutdown methods

WMOkiishi:
## Change description
These changes add type annotations to some basic `ShowBase` methods used on setup and shutdown. This mostly comprises adding annotations to attributes assigned in `ShowBase.__init__`. See also #1217, which aimed to do mostly the same thing, but seems to have been abandoned.

## Checklist

### DIFF
--- a/direct/src/showbase/ShowBase.py
+++ b/direct/src/showbase/ShowBase.py
@@ -32,6 +32,8 @@ built-in scope.
 
 """
 
+from __future__ import annotations
+
 __all__ = ['ShowBase', 'WindowControls']
 
 # This module redefines the builtin import function with one
@@ -71,9 +73,11 @@ from panda3d.core import (
     GraphicsPipe,
     GraphicsPipeSelection,
     GraphicsWindow,
+    InputDevice,
     InputDeviceManager,
     InputDeviceNode,
     KeyboardButton,
+    Lens,
     LensNode,
     Mat4,
     ModelNode,
@@ -98,6 +102,7 @@ from panda3d.core import (
     RescaleNormalAttrib,
     SceneGraphAnalyzerMeter,
     TexGenAttrib,
+    TexMemWatcher,
     Texture,
     TextureStage,
     Thread,
@@ -141,7 +146,7 @@ import importlib
 from direct.showbase import ExceptionVarDump
 from . import DirectObject
 from . import SfxPlayer
-from typing import ClassVar, Optional
+from typing import Any, Callable, ClassVar, Literal
 if __debug__:
     from direct.showbase import GarbageReport
     from direct.directutil import DeltaProfiler
@@ -170,7 +175,7 @@ class ShowBase(DirectObject.DirectObject):
     aspect2d: NodePath
     pixel2d: NodePath
 
-    def __init__(self, fStartDirect=True, windowType=None):
+    def __init__(self, fStartDirect: bool = True, windowType: str | None = None) -> None:
         """Opens a window, sets up a 3-D and several 2-D scene graphs, and
         everything else needed to render the scene graph to the window.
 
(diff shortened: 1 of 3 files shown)


direct/src/showbase/ShowBase.py
```diff
@@ -371,19 +376,19 @@ def __init__(self, fStartDirect=True, windowType=None):
...
         #: need to call :meth:`~panda3d.core.CollisionTraverser.traverse()`
         #: yourself every frame.
-        self.cTrav = 0
-        self.shadowTrav = 0
+        self.cTrav: CollisionTraverser | Literal[0] = 0
+        self.shadowTrav: CollisionTraverser | Literal[0] = 0
         self.cTravStack = Stack()
         # Ditto for an AppTraverser.
-        self.appTrav = 0
+        self.appTrav: CollisionTraverser | Literal[0] = 0
```
rdb, line 383: This is not meant to be a CollisionTraverser, better make it Any (though we really just ought to remove this field)

direct/src/showbase/ShowBase.py
```diff
@@ -279,68 +284,68 @@ def __init__(self, fStartDirect=True, windowType=None):
...
+        self.camLens: Lens | None = None
+        self.camFrustumVis: NodePath | None = None
         self.direct = None
         #: This is used to store the wx.Application object used when want-wx is
         #: set or `startWx()` is called.
-        self.wxApp = None
+        self.wxApp: Any | None = None
         self.wxAppCreated = False
-        self.tkRoot = None
+        self.tkRoot: Any | None = None
```
rdb, line 348: Is `Any | None` different than just `Any`?

direct/src/showbase/ShowBase.py
```diff
@@ -512,41 +517,41 @@ def __init__(self, fStartDirect=True, windowType=None):
...
+        builtins.pixel2d = self.pixel2d  # type: ignore[attr-defined]
+        builtins.render = self.render  # type: ignore[attr-defined]
+        builtins.hidden = self.hidden  # type: ignore[attr-defined]
+        builtins.camera = self.camera  # type: ignore[attr-defined]
+        builtins.loader = self.loader  # type: ignore[attr-defined]
+        builtins.taskMgr = self.taskMgr  # type: ignore[attr-defined]
+        builtins.jobMgr = self.jobMgr  # type: ignore[attr-defined]
+        builtins.eventMgr = self.eventMgr  # type: ignore[attr-defined]
+        builtins.messenger = self.messenger  # type: ignore[attr-defined]
+        builtins.bboard = self.bboard  # type: ignore[attr-defined]
```
rdb, line 532: Hmm, this is awkward, but I guess there's no other way... I'm told that we'd need a custom version of typeshed otherwise to be able to define custom builtins.  Do these ignores work with mypyc?

rdb (changes requested):

direct/src/showbase/ShowBase.py
WMOkiishi: Yes, it means that, when accessing the field, a type checker will make sure the potential case that the value is `None` is handled.

direct/src/showbase/ShowBase.py
WMOkiishi: It looks like, in general ,mypyc can still compile code like this, but it crashes if any of the values are accessed without the `builtins` prefix. If we want to have less `type: ignore` comments, we could add a function that sets attributes on `builtins`, so the `type: ignore` would only need to be in one place. Configuration could also switch the function to a no-op.

There might also be a way to set things up such that, when compiling with mypyc, some other object that does allow these attributes to be set is instead imported as `builtins`.

direct/src/showbase/ShowBase.py
Souzooka: I should note that this usages of this syntax (contrast to `Union[CollisionTraverser, Literal[0]]`) only works on Python 3.10 and below.

WMOkiishi: Unless `from __future__ import annotations` is used (which it is). It will just be a string at runtime, and type-checkers can understand it even if they're run on 3.8 or 3.9.

direct/src/showbase/ShowBase.py
rdb: I think if mypyc crashes when these are accessed, we shouldn't just ignore these... maybe we ought to just suppress attr-defined warnings?

WMOkiishi:
Alright, I think I've found a good way to make mypy not complain about setting custom attributes on `builtins`. Let me know if it's acceptable.

Issue #1587, opened 2024-01-06
Modifying mutable option in DirectGui modifyes the default value


Augustifolia:
## Description

It appears that when DirectGui saves the default value for some option, it saves the same object that you can access with `gui["option"]`. This means that modifying the option (if it is mutable, like a list) also modifies the default value.

Fixing this should not be that difficult. Just check if the object is mutable before the default value is saved. If it is mutable, create a copy to save.

## Steps to Reproduce

```
from direct.showbase.ShowBase import ShowBase
from direct.gui.DirectGui import DirectButton

ShowBase()
gui = DirectButton()
print(gui.configure("extraArgs"))  # the default value is "[]" and the current value is "[]"
gui["extraArgs"].append(1)  # this not only modifies the current value, but also the default value
print(gui.configure("extraArgs"))  # this will now give us the default value "[1]" 

base.run()
```

## Environment
 * Operating system: Linux Mint
 * System architecture: x86_64
 * Panda3D version: 1.10.13
 * Installation method: pip
 * Python version (if using Python): 3.11
 * Compiler (if using C++):

BMaxV:
I think this is fully intended behavior of DirectGUI ?

Setting other kinds of values works the same way and if I recall correctly, accessing these variables is required for rebuilding / reconstruction of the GUI.

At least in this case you're modifying the button and it's values. It should be expected behavior that you get the value back that you just set, for this object.

Augustifolia:
I'm sorry if my original post was a bit unclear. But DirectGui actually keeps track of both the default value and the current value for each option. `gui.configure("extraArgs")` returns a tuple of the option name, default value and current value. So the expected behavior would be that the default value is still an empty list, while the current value becomes `[1]`.

I included an example that produces the expected behavior. 

```
from direct.showbase.ShowBase import ShowBase
from direct.gui.DirectGui import DirectButton

ShowBase()
gui = DirectButton()
print(gui.configure("extraArgs"))  # this returns ("extraArgs", [], [])
gui["extraArgs"] = [1]
print(gui.configure("extraArgs"))  # this returns ("extraArgs", [], [1]), as expected
base.run()
```

I hope this clears up any confusion.

rdb:
Might as well fix this, though I don't think it's really a problem as the issue is not "contagious" across different DirectButton objects

PR #1593, opened 2024-01-15
Fixed bug with default values in DirectGui

Augustifolia:
## Issue description

This change is intended to fix issue #1587. Where mutable options for DirectGui items could be changed after creation time by modifying the value of the option with `gui["option"]`. 

## Solution description

I have simple added a check to see it the option is mutable, and if it is save a copy of the object as the default value instead of the original object. This makes sure that modifying the current value does not change the default.

## Checklist


rdb:
Thanks!  This is the right fix, but I can't apply it in the current form to 1.10.x for compatibility reasons.  I'm going to close the PR and submit a modified fix myself.

Issue #1596, opened 2024-01-18
CMake: No ANSI colored output when using CMake + Ninja


maxrdz:
## Description
Linux users are recommended to run `cmake -G Ninja ..` when generating the build files for Panda through CMake. When using Ninja, no color output is shown on the terminal making it hard to read / debug compiler messages. Ninja buffers output, so tools are writing to a pipe, not directly to the terminal, so auto-detection may turn color diagnostics off conservatively. 

## Steps to Reproduce
Use Ninja as your CMake generator.

## Environment
 * Operating system: Arch Linux
 * System architecture: x86_64
 * Panda3D version: 1.10.14
 * Installation method: source
 * Python version (if using Python): N/A
 * Compiler (if using C++): GCC

PR #1597, opened 2024-01-18
CMake: Force GCC/Clang color diagnostics for Ninja (#1596)

maxrdz:
## Issue description
Linux users are recommended to use Ninja when generating the build files through CMake. Ninja buffers output, so tools are writing to a pipe, not directly to the terminal, so auto-detection may turn ANSI colored output off conservatively.

## Solution description
This PR forces color diagnostics by setting `CMAKE_COLOR_DIAGNOSTICS` to `ON` on CMake versions 3.24 and above. For older CMake versions, compiler flags are invoked manually for GCC and Clang.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,25 @@ if(APPLE)
   endif()
 endif()
 
+# Enable 'FORCE_ANSI_COLORED_OUTPUT' option when using Ninja.
+# Forces GCC/Clang compilers to enable color diagnostics. Enabled by default.
+if(CMAKE_GENERATOR STREQUAL "Ninja")
+  option(FORCE_ANSI_COLORED_OUTPUT "Always produce ANSI-colored output. (GNU/Clang only)" TRUE)
+
+  if(FORCE_ANSI_COLORED_OUTPUT)
+    set(CMAKE_COLOR_DIAGNOSTICS ON)
+    # CMake versions 3.24 and below do not support this option, so we have
+    # to invoke the color diagnostics flags manually.
+    if(CMAKE_VERSION VERSION_LESS "3.24.0")
+      if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        add_compile_options(-fdiagnostics-color=always)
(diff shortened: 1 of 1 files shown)


maxrdz:
This resolves #1596.

CMakeLists.txt
```diff
@@ -41,6 +41,25 @@ if(APPLE)
   endif()
 endif()
 
+# Enable 'FORCE_ANSI_COLORED_OUTPUT' option when using Ninja.
+# Forces GCC/Clang compilers to enable color diagnostics. Enabled by default.
+if(CMAKE_GENERATOR STREQUAL "Ninja")
```
Moguri, line 46: Does Make support this as well?

Moguri (commented):
Maybe instead of a new flag and trying to support older versions of Cmake, should we just change the default for CMAKE_COLOR_DIAGNOSTICS for relevant generators and CMake versions?

CMakeLists.txt
maxrdz: Make works differently (not sure exactly about the details) but it doesn't need to force color diagnostics.
but yeah we could enable it for both generator options, although not necessary

maxrdz:
> Maybe instead of a new flag and trying to support older versions of Cmake, should we just change the default for CMAKE_COLOR_DIAGNOSTICS for relevant generators and CMake versions?

The `CMAKE_COLOR_DIAGNOSTICS` option is relatively new to CMake (version `3.24.x`) so I think it would be a safe route to take if we first verify the version of CMake being used.

maxrdz:
I do agree that we can just always apply it for Ninja gen builds and remove the new CMake option, since it should not affect anything with the build unless someone doesn't want the compiler to print ANSI colors.

rdb (changes requested):
I'm OK with this, but can we make it a cache variable, so that users can disable it?  And if people turn it off, we shouldn't add the compiler option.

maxrdz:
Sorry for the late update, was busy for a few weeks. :)

Moguri (approved):
I am okay with this as well.

PR #1599, opened 2024-01-19
build: Bring in panda3d-thirdparty repo as thirdparty-scripts

Moguri:
## Checklist


rdb:
Thanks so much for working on this!

We should consider whether we really want to include the Pmw sources, or whether we should fetch them from a separate repository.  I am leaning towards the latter, rather than include them with the Panda3D sources.

As discussed in Discord, I would prefer keeping the scripts in the thirdparty directory (or a subdirectory thereof), or to put them in cmake/thirdparty, or something like that, rather than a separate top-level directory.

dtool/src/prc/CMakeLists.txt
```diff
@@ -92,6 +92,10 @@ target_include_directories(p3prc PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY
 target_link_libraries(p3prc p3dtool PKG::OPENSSL)
 target_interrogate(p3prc ALL EXTENSIONS ${P3PRC_IGATEEXT})
 
+if(HAVE_OPENSSL AND WIN32)
+  target_link_libraries(p3prc ws2_32.lib crypt32.lib)
```
rdb, line 96: I think there's a better way to add indirect dependencies in CMake.

kamgha: The `.lib`-suffices for the system libraries won't work with MinGW.

Moguri: I copied this from another location (p3express maybe?); I'll look into fixing this properly and updating that other location while I'm at it.

Moguri: I believe this is what `OpenSSL:applink` (https://cmake.org/cmake/help/latest/module/FindOpenSSL.html) is for. However, it is only available on CMake 3.18+. Ubuntu 20.04 only has 3.16. We could probably do some guards to make use of this feature -- and thus require 3.18 -- only on Windows. We already do this for macOS, which requires 3.16 for ObjC (our current minimum is 3.13).

Issue #1605, opened 2024-01-23
GtkStatsMonitor crashes when closing session


el-dee:
## Description

When closing pstats or closing a session, the program crashes with the following stack trace :

<pre>
#0  0x00007ffff68a4198 in g_sequence_remove () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff6e40eb4 in  () at /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#2  0x00007ffff6978f03 in g_cclosure_marshal_VOID__OBJECTv () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#3  0x00007ffff6975a56 in  () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#4  0x00007ffff6994b48 in g_signal_emit_valist () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#5  0x00007ffff69950f3 in g_signal_emit () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x00007ffff6dc6f7c in gtk_container_remove () at /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#7  0x000055555556e9a2 in GtkStatsMonitor::close() ()
#8  0x000055555556eb36 in GtkStatsServer::close_session() ()
#9  0x000055555556fc20 in GtkStatsServer::create_window()::{lambda(_GtkWidget*, _GdkEvent*, void*)#1}::_FUN(_GtkWidget*, _GdkEvent*, void*) ()
#10 0x00007ffff703336f in  () at /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#11 0x00007ffff6975a56 in  () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#12 0x00007ffff6993df1 in g_signal_emit_valist () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#13 0x00007ffff69950f3 in g_signal_emit () at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#14 0x00007ffff6fdd9b3 in  () at /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#15 0x00007ffff6e9b13d in gtk_main_do_event () at /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
[...]

PR #1610, opened 2024-01-31
pstats: Workaround for Gtk bug that causes UI to crash when a session is closed

el-dee:
## Issue description

Workaround implementation to solve https://github.com/panda3d/panda3d/issues/1605

## Solution description

As workaround for https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/2029, we manually create a GtkFlowBoxChild instance and attach the label to it and increase its reference count so it it not prematurely destroyed when removed from the GtkFlowBox, causing the UI to segfault.

I am unsure if `GtkStatsMonitor::_status_bar_labels` and `GtkStatsServer::_status_bar_label` should have been renamed or not as they now store `GtkFlowBoxChild` instead of `GtkLabel`. I thought the renaming would be too invasive.

## Checklist

### DIFF
--- a/pandatool/src/gtk-stats/gtkStatsMonitor.cxx
+++ b/pandatool/src/gtk-stats/gtkStatsMonitor.cxx
@@ -60,8 +60,9 @@ close() {
 
   remove_all_graphs();
 
-  for (GtkWidget *label : _status_bar_labels) {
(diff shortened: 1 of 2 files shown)


rdb (approved):
Thank you!

PR #1611, opened 2024-02-04
support no audio for cmake

pmp-p:
because wasi openal will not happen with full exceptions+atomics c++ support.


rdb:
The intent is that you can just pass `-DHAVE_AUDIO=OFF` to suppress the error, does that not work?

pmp-p:
i'll check that but i've imagined that -DHAVE_AUDIO=YES +  no driver == null sink

PR #1612, opened 2024-02-04
fix build with some assimp distributions

pmp-p:
is an issue on pclinuxos x64 assimp 3.3.1-2, so maybe mandriva* too


pandatool/src/assimp/assimpLoader.cxx
```diff
@@ -1177,7 +1177,11 @@ load_node(const aiNode &node, PandaNode *parent, bool under_joint) {
       //case AI_BOOL:
       //  value = (*static_cast<bool *>(entry.mData)) ? "1" : "";
       //  break;
+#if defined(AI_INT32)
```
rdb, line 1180: This is an enum so this will always be false

pmp-p:
that happens on assimp 3.3.1, way too old anyway

PR #1620, opened 2024-02-27
Enable OpenAL-Soft extensions for non-framework builds and add config for disabling HRTF

### LINKED ISSUE #1613: Include alext.h header file to enable using OpenAL Extensions
## Description
If panda3d ever wants the ability to utilize [OpenAL-Soft extensions](https://openal-soft.org/openal-extensions/), then the alext.h file will need to be included in the audio source via `<AL/alext.h>`. 

## Blockers
On my macOS 14.x, there's no alext.h file in my Frameworks folders, so I'm assuming Apple stopped including them at some point, for some reason. Thus, trying to include `<OpenAL/alext.h>` fails with the frameworks linked. The solution would be to just use a custom build, however, it seems that the makepanda build relies solely on the OpenAL framework, as when I tried to use the `openal-disabled` libs in the latest thirdparty zip, it failed to link, likely because the libs are outdated.

I'm not sure what the best path forward for this is, either to start building openal again in the prebuilt thirdparty zips, or some more flexibility in the CLI of makepanda, perhaps.

## Use Case
I would love to add support for controlling HRTF via the [ALC_SOFT_HRTF](https://openal-soft.org/openal-extensions/SOFT_HRTF.txt) extension. This would allow greater control of audio playback fidelity for headphone users in games.

gabeklavans:
## Issue description
There was no include for the `alext.h` header, meaning the engine could not use any [OpenAL-Soft extensions](https://openal-soft.org/openal-extensions). Additionally, HRTF was always enabled, meaning when OpenAL detected headphones as the output device, it would apply the default HRTF to the audio.

## Solution description
Added the `alext.h` include for non-framework builds (the OpenAL framework does not include alext.h) and added a config variable `audio-want-hrtf` to determine how the `ALC_SOFT_HRTF` extension should be configured in the openal audio manager.

## Checklist

I've tested a wheel build of this branch using python3.9 on Windows, macOS, and Ubuntu 22.04

resolves #1613


makepanda/makepanda.py
```diff
@@ -2601,7 +2601,8 @@ def WriteConfigSettings():
         dtool_config["PYTHON_FRAMEWORK"] = 'Python'
         dtool_config["PHAVE_MALLOC_H"] = 'UNDEF'
         dtool_config["PHAVE_SYS_MALLOC_H"] = '1'
-        dtool_config["HAVE_OPENAL_FRAMEWORK"] = '1'
+        if not os.path.isdir(GetThirdpartyDir() + "openal"):
+            dtool_config["HAVE_OPENAL_FRAMEWORK"] = '1'
```
rdb, line 2605: Hmm, this isn't perfect, but I can't see a better way right now

panda/src/audio/config_audio.cxx
```diff
@@ -75,6 +75,13 @@ ConfigVariableInt audio_preload_threshold
           "practical to stream multiple sound-files from disk at the same "
           "time - the hard drive seek time makes it stutter."));
 
+ConfigVariableBool audio_want_hrtf
+("audio-want-hrtf", true,
+ PRC_DESC("This determines whether OpenAL-Soft should activate HRTF in "
+          "certain hardware configurations. Set it to true to cause "
+          "OpenAL to automatically apply HRTF processing to all output "
+          "audio when headphones are used for playback."));
```
rdb, line 83: Should we make this an openal-specific config variable?

makepanda/makepanda.py
gabeklavans: I had the same thought. I couldn't find anything that specified whether openal was enabled specifically via the thirdparty directory. However, I think removing OpenAL Framework support entirely is going to be my next suggestion, so I didn't belabor the change too much.

panda/src/audio/config_audio.cxx
gabeklavans: There are almost no config variables that specify that they are for openal by name, most of the config variables that do in fact only apply to openal just start contain "audio" (I know this for a fact with `audio-preload-threshold` and I'm sure there are many others).

So my way making it an "openal-specific config" was just to place it along side the other ones. Let me know if there's more I should do for that.

rdb:
Emscripten build failed... maybe we need a separate dtool_config.h flag for alext presence.

rdb:
Or, we need to just vendor alext.h, just like we do with glext.h... it's feature-gated at runtime anyway.

gabeklavans:
> Emscripten build failed... maybe we need a separate dtool_config.h flag for alext presence.

[Emscripten does actually support extensions](https://emscripten.org/docs/porting/Audio.html), including the one I'm trying to use here ALC_SOFT_HRTF. Not sure how if they don't expose alext.h, since you kinda need the defines. Maybe they pack it into alc.h? I can experiment on my local.

Otherwise yeah I can look into feature gating

rdb:
I am leaning towards investigating shipping our own alext.h.

gabeklavans:
Poking around in the PR that added support for the ALC_HRTF_SOFT extension in Emscripten, specifically in the tests added for it, it seems they require you to [specify the enum's value directly](https://github.com/emscripten-core/emscripten/pull/5367/files#diff-a559d1bf5a6775a4b2b1f9247010df33761da17621b2ad8e5e8740a17f02ba37R105), which is not quite ideal for our case.

That is to say, I agree that we could try to ship our own alext.h file for Emscripten builds. I'll try to look into that soon, likely by looking at glext.h as you mentioned for an example.

gabeklavans:
The missing `alext.h` file is actually [already on the radar for emscripten](https://github.com/emscripten-core/emscripten/pull/9852#issuecomment-556949790), but no one's made progress on it in a while. I think a better solution to this would be to just fix this upstream, since it seems to be a pretty simple fix there (just adding the alext.h header to their includes).

gabeklavans:
alext.h landed in emscripten https://github.com/emscripten-core/emscripten/pull/21857 so whenever the next version releases, if we bump to that, this PR should hopefully successfully build

gabeklavans:
@rdb I've updated the emscripten version in the CI workflow in this PR to the minimum that contains the upstreamed alext.h file. Hopefully it successfully builds now. Let me know if there are any other places that the emscripten version should be updated.

gabeklavans:
bumped the version to avoid the licensing issue addressed in https://github.com/emscripten-core/emscripten/pull/22349

This PR is likely all good to go now

rdb (approved):
LGTM

gabeklavans:
> The default is to fall back to user config, as can be seen here: [kcat/openal-soft@`c0cbe60`/alc/alc.cpp#L1746-L1754](https://github.com/kcat/openal-soft/blob/c0cbe602ce66481a822ba16856052a47154f5cc5/alc/alc.cpp#L1746-L1754) [kcat/openal-soft@`c0cbe60`/alc/alc.cpp#L1912-L1931](https://github.com/kcat/openal-soft/blob/c0cbe602ce66481a822ba16856052a47154f5cc5/alc/alc.cpp#L1912-L1931)
> 
> I'm actually inclined that we shouldn't be specifying the hrtf setting at all by default, since what we set overrides user configuration.

User config is actually not the preferred mechanism for configuring OpenAL-Soft. It’s less stable and less idiomatic. Using AL Extensions better, which is what this PR implemented. See my convo here for context https://github.com/kcat/openal-soft/issues/825#issuecomment-1488362668

gabeklavans:
I see what you’re saying about overriding user config even if a particular Panda3D game hasn’t specified an HRTF setting, though. Maybe if it’s not set we can have Panda3D just not pass any hrtf value so that it can fall back to whatever underlying mechanism OpenAL has at the moment.

PR #1639, opened 2024-03-26
3.13a5 draft

pmp-p:
wip 3.13git build ( with wasi support in mind )


dtool/src/interrogatedb/py_compat.h
```diff
@@ -275,6 +275,15 @@ INLINE bool PyLong_IsNonNegative(PyObject *value) {
 }
 #endif
 
+/* Python 3.13 */
+
+#if PY_VERSION_HEX >= 0x030d0000  // 3.13
+#define Py_BUILD_CORE
+#include "internal/pycore_unicodeobject.h"
```
rdb, line 282: This can't be the right way to fix this... we need to lobby for them to put these back until they've created a proper API before they remove anything.  I understand that these functions are underscored and therefore unstable, but there's no other way to properly deal with longs right now, so it's preposterous to remove them before adding a replacement!

rdb: I spoke too soon, it looks like they do have some new functions in 3.13:
https://docs.python.org/3.13/c-api/long.html

We should look into using those instead where we are currently using the unstable functions.

Issue #1410, opened 2022-12-19
Persistent Python wrapper objects


WMOkiishi:
## General Idea
It would be convenient if Panda's C++ objects were consistently exposed to Python with the same instance of a Python wrapper. This would require the C++ object to hold onto a reference to said Python wrapper, though, creating a reference cycle that won't automatically be taken care of.

## First Steps
For cases where the Python wrapper loses all other references (i.e., reaches a reference count of 1) before the C++ object does, the reference cycle can be detected and broken by an overridden `unref` method. It could check if the C++ object and its associated Python wrapper both have reference counts of 1, meaning they must only be referred to by each other. If so, the reference cycle can be broken on the C++ end, allowing both reference counts to reach 0, causing the objects to be cleaned up.

## The Problem
The procedure described above cannot do anything for cases where the C++ object reaches a reference count of 1 before its Python wrapper does. To handle those cases, we will have to hook into one of the following (or possibly some other overlooked area):

- Panda3D's `ReferenceCount::ref`
- Python's [`tp_traverse`](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse)
- Python's [`gc.callbacks`](https://docs.python.org/3/library/gc.html#gc.callbacks)
- Python destructors

## Analysis
Overriding `ref`: If both `ref` and `unref` were overridden, we could decrement the Python wrapper's reference count whenever the C++ reference count reaches 1 and increment it whenever the C++ reference count goes back above 1. This would allow Python to break the reference cycle by destroying the wrapper once it and the C++ object are only referenced by each other. This would require making `ref` virtual, though, incurring a performance penalty. While this could theoretically be worth the cost, other solutions seem more promising for less overhead.

`tp_traverse`: Classes holding onto references to their Python wrappers could implement a custom `tp_traverse` that only reports the wrapper's self-reference while the underlying C++ object's reference count is 1. This essentially allows (and relies on) Python's GC to clean up the wrapper (and thus the reference cycle), but only once we say it's okay to do so. Note that we could still implement the `unref` override as well to clean up the cases it can.

`gc.callbacks`: Python allows for custom callbacks to be called before or after its own garbage collector runs. We could register some function there that checks for any isolated reference cycles of Panda objects and their Python wrappers and cleans them up. Again, this could be done in addition to the `unref` override.

[...]

PR #1640, opened 2024-03-29
Allow subclassing C++ objects from Python, GC support

rdb:
## Issue description
The issue is that Python subclasses of C++ types are currently not preserved when they are handed off to C++.  See #1410 for more information.  Furthermore, reference cycles involving Panda objects cannot be found by Python's GC.  Both problems are rooted in the same limitation, namely that the Python wrapper object is not persisted, but rather, the bindings create new wrapper objects on demand.  Solving this issue further opens up a broad range of future possibilities.

## Solution description
This change enables persistent Python wrapper objects for Python subclasses of typed, reference counted C++ objects.  That means that these objects will store a reference to `self` on the C++ object, and interrogate will always return that instead of making a new Python wrapper every time it is returned from C++.
    
Practically, this means that you could subclass eg. PandaNode, Event, Fog, what have you, and store these in the scene graph - the Python data in the subclass will be retained and Panda will return your subclass when you ask for the object back rather than creating a new wrapper object without your original data.
    
To do this, Interrogate generates a proxy class inheriting from the C++ type with additional room to store a `self` pointer and a TypeHandle (which is returned by an overridden `get_type()`).  This TypeHandle is automatically created by registering the Python subclass with the typing system.  The proxy class is only used when the constructor detects that it's constructing for a subtype, so that regular uses of the C++ type are not affected by this mechanism.
    
(The registration with the typing system could use some improvement.  There's no regard for namespacing right now, in particular.  Furthermore, we could move the registration to an `__init_subclass__()` method, with parameters to specify an existing TypeHandle or to customize the type name.)
    
This creates a reference cycle, which must be cleared somehow.  One way this happens is by overriding `unref()` in the proxy, which checks that if the Python and C++ reference counts are both 1, this must be the circular reference, and then it breaks the cycle.  Note that this will _only_ work if all other Python references have been cleared before the last C++ reference goes away.  For the other case, we need to rely on Python's garbage collector, so these classes also implement tp_traverse and tp_clear.

This commit also therefore re-enables Python garbage collector support (ie. defining `__traverse__()`), which was previously disabled due to the problems caused by multiple Python wrappers referring to the same C++ object.  To avoid these problems, Panda will only cooperate with Python's GC if the C++ reference count is 1, in which case we can trivially prove that the last reference must be from the Python wrapper.  Note that this explains why we need persistent wrappers for traversal to work--if there are multiple Python wrappers pointing to the C++ object, and they are all participating in a reference cycle, the reference count cannot be 1, and cycle detection does not work.  By default, the GC is only enabled for Python subclasses, but a class may define `__traverse__()` in order to opt-in, keeping the previous limitation in mind.

There is a second mechanism introduced by this commit: classes may define a public `__self__` member of type `PyObject *` if they want to use persistent objects even if they aren't being subclassed.  This allows these classes to participate in Python's GC and avoid the situation above.  This is utilized by this PR in order to add GC support to PythonTask, making it possible for reference cycles involving tags to be found and broken.

Note that this may be considered a regression, since the introduction of a reference cycle for all tasks may affect existing applications, especially if they have disabled the GC for some reason (we even have a `gc-save-all` config switch for this in ConnectionRepository, I genuinely hope nobody has enabled this).  However, for tasks, I think it is generally the case that the last reference is in C++, since tasks are usually created and then handed off to the C++ task manager, so they can be cleared without the GC.  And it is relatively easy to work around to ensure that this is the case.  If this turns out to be a problem, though, we can add a special garbage collection pass to the task manager, an explicit method, or we can disable the behaviour using a Config.prc switch.

The cycle detection is also implemented for PandaNode's Python tags, but this is very incomplete, since the traversal doesn't recurse into children and it won't work if there is more than one wrapper object that is part of a cycle, since PandaNode by default doesn't use persistent wrappers.  This problem may need more attention later.

## Future
This change paves the way for supporting virtual method overrides from Python.  We probably need to think about whether we want to support this selectively, for some methods, or for all methods, and it also needs consideration about how to prevent infinite recursion when upcalling to the superclass.

I'm sure there are other cool uses we can think of, such as a Python-based importer subclassing Material to store custom properties, or as a nicer way to pass Python objects around the C++ event queue.

Serialization support is still a TODO.  However, since you can already subclass TypedWritable from Python, we can probably port this over to reference-counted TypedWritables without too much work.

It does not solve the problems involving subclassing NodePath.  It is probably possible, but there are more challenges to overcome if we want to do that, and I'm not convinced we should--leaving this to be considered at another time.

## Checklist

### DIFF
--- a/dtool/src/cppparser/cppPreprocessor.h
+++ b/dtool/src/cppparser/cppPreprocessor.h
@@ -25,7 +25,6 @@
 #include "vector_string.h"
 
 #include <map>
-#include <list>
 #include <vector>
 #include <unordered_map>
 
--- a/dtool/src/interrogate/functionRemap.cxx
+++ b/dtool/src/interrogate/functionRemap.cxx
@@ -450,8 +450,6 @@ get_call_str(const string &container, const vector_string &pexprs) const {
     _parameters[pn]._remap->pass_parameter(call, get_parameter_expr(pn, pexprs));
 
   } else {
-    const char *separator = "";
-
     // If this function is marked as having an extension function, call that
     // instead.
     if (_extension) {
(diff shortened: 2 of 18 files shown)


rdb:
Accidentally pushed this to master early, if there are problems let me know and I can address them or revert them

PR #1654, opened 2024-05-12
interrogate: avoid GC exceptions processing objects mid-construction

johnnyapol:
## Issue description
After https://github.com/panda3d/panda3d/commit/38692dd525525e20cd06204a646b35dd7e39902a, GC pauses occuring before
the `__init__` function of the C++ backed object is called in a inheritance chain will
cause an exception during garbage collection and cause the python runtime to exit.

This can be observed by say, forcing a GC-pause in `MetaInterval.__init__` before it calls `CMetaInterval::__init__`.

## Solution description

This change replaces the call to Dtool_Call_ExtractThisPointer with a direct check for whether DtoolInstance_VOID_PTR is nullptr and exit with status 0 if that's the case, rather than raising an exception.

## Checklist

### DIFF
--- a/dtool/src/interrogate/interfaceMakerPythonNative.cxx
+++ b/dtool/src/interrogate/interfaceMakerPythonNative.cxx
@@ -3057,9 +3057,10 @@ write_module_class(ostream &out, Object *obj) {
       out << "static int Dtool_Traverse_" << ClassName << "(PyObject *self, visitproc visit, void *arg) {\n";
       out << "  // If the only reference remaining is the one held by the Python wrapper,\n";
       out << "  // report the circular reference to Python's GC, so that it can break it.\n";
-      out << "  " << cClassName  << " *local_this = nullptr;\n";
-      out << "  if (!Dtool_Call_ExtractThisPointer(self, Dtool_" << ClassName << ", (void **)&local_this)) {\n";
-      out << "    return -1;\n";
+      out << "  " << cClassName  << " *local_this;\n";
+      out << "  DTOOL_Call_ExtractThisPointerForType(self, &Dtool_" << ClassName << ", (void**)(&local_this));\n";
+      out << "  if (local_this == nullptr) {\n";
+      out << "    return 0;\n";
       out << "  }\n";
       out << "  if (local_this->get_ref_count() == (int)((Dtool_PyInstDef *)self)->_memory_rules) {\n";
       if (py_subclassable) {


johnnyapol:
I am also open to alternative methods to fixing this like for instance creating an argument or version of DTool_Call_ExtractThisPointer that doesn't set a PyErr but figured I'd just start with the simplest patch and go from there.

rdb:
Thanks. I would prefer a solution that didn't set an exception at all, if it is indeed a valid state, we can just do the `DtoolInstance_Check` and `DtoolInstance_UPCAST` calls directly in the tp_traverse handler.

I want to understand the problem a bit better, though. In what situations is tp_traverse called with an exception already set? And in what situations is tp_traverse called without the `_signature` being initialized properly, which is done by `Dtool_new_<type>`? Do you have a test case?

johnnyapol:
Thanks for the response.

This is the simplest test case I can reproduce the problem at

```
import gc

from panda3d.direct import CMetaInterval

class MetaGCTestInterval(CMetaInterval):
    def __init__(self):
        gc.collect()

        super().__init__('MetaIntervalGCTest')

if __name__ == "__main__":
    MetaGCTestInterval()⏎   
```

Running this against code built at https://github.com/panda3d/panda3d/commit/38692dd525525e20cd06204a646b35dd7e39902a using 
1. python makepanda/makepanda.py --everything --threads 8 --wheel
2. Python 3.12.3 (main, Apr 23 2024, 09:16:07) [GCC 13.2.1 20240417] (Arch Linux's python package)
3. gcc (GCC) 14.1.1 20240507

Gives the following output:
```
Exception ignored in garbage collection:
  File "test.py", line 7, in __init__
    gc.collect()
TypeError: C++ object is not yet constructed, or already destructed.
```

In my experience this was prone to happening particularly with MetaIntervals since they have a bit of a window between calling `CMetaInterval::__init__` compared to some of the other native objects. Doing it in a real game context by say patching MetaInterval's construct to call gc.collect would lead to the game crashing.

As far as the question about tp_traverse exceptions, I was looking at the source of gcmodule.c in Python 3.8 -- I haven't checked all the in-between ones since this appears to hold for panda's minimum supported python and for 3.12 so I assume it's similar -- and I've trimmed the collect method here for brevity but:

```
/* This is the main function.  Read this to understand how the
 * collection process works. */
static Py_ssize_t
collect(struct _gc_runtime_state *state, int generation,
        Py_ssize_t *n_collected, Py_ssize_t *n_uncollectable, int nofail)
{
... initialization logic ommitted
    /* Using ob_refcnt and gc_refs, calculate which objects in the
     * container set are reachable from outside the set (i.e., have a
     * refcount greater than 0 when all the references within the
     * set are taken into account).
     */
    update_refs(young);  // gc_prev is used for gc_refs
    subtract_refs(young);

... more logic for GC generations, weakref processing, etc

    if (PyErr_Occurred()) {
        if (nofail) {
            PyErr_Clear();
        }
        else {
            if (gc_str == NULL)
                gc_str = PyUnicode_FromString("garbage collection");
            PyErr_WriteUnraisable(gc_str);
            Py_FatalError("unexpected exception during garbage collection");
        }
    }

   ...
}```

It doesn't seem any of the intermediate logic clears any PyErrors (unless the GC is invoked with nofail > 0), so in this case it seems like anything the GC is capable of tracking is capable of setting a PyErr when traverse is called.

rdb:
Okay, yes, gotcha.  So the main problem is that it's valid to call tp_traverse on a valid object, and it shouldn't raise an exception for that.  So, we should just replace the call to Dtool_Call_ExtractThisPointer with a direct check for whether `DtoolInstance_VOID_PTR` is nullptr and exit with status 0 if that's the case, rather than raising an exception.

johnnyapol:
Sounds good. I've redone the patch to just check the output of DTOOL_Call_ExtractThisPointerForType which seems to just be implemented in terms of DtoolInstance_VOID_PTR.

rdb (approved):
LGTM, thanks

PR #1656, opened 2024-05-14
Ability to read comments/metadata from sound files

Wizzerinus:
## Issue description
This PR allows loading metadata/comments/tags from sound files and exposes those to the programmer. Example code:

```python
from panda3d.core import loadPrcFileData

loadPrcFileData("", "audio-library-name p3openal_audio")
b = ShowBase()
f = b.loader.loadMusic("file.ogg")
print(list(f.raw_comments))  # something like ['ARTIST=artist', 'TITLE=my music track\n']
print(f.comments['TITLE'])  # something like 'my music traсk'
```

## Solution description
Metadata is exposed through two virtual methods, one on the AudioCursor, and another one on the AudioSound, linked through some binding code in AudioManager.

We have 3 sound managers, FMod, Null and OpenAL. FMod is way too encapsulated to do anything, and Null is Null, so those 2 managers return an empty string when the comment is read. For OpenAL, the method depends on the underlying AudioCursor class.

There are seven types of AudioCursor: FFmpeg, Opus, Flac, Wav, Vorbis, Microphone and Userdata. Only three of these currently have an implementation:
- Implementation for FFmpeg is based on https://ffmpeg.org/doxygen/0.7/metadata-example_8c-source.html. Note that FFmpeg uses a different data structure from ogg, which had to be converted to `vector<string>`.
- Implementation for Opus and Vorbis is more or less the same with minor differences based on how the Tag struct is accessed.
- There is no implementation for Flac. The reason being the Panda3D's dr_flac implementation being extremely old and that version does not have the metadata iterator. This can be revisited in the future when/if the bundled dr_flac is updated.
- Wav does not have an implementation seeing as Panda uses bytestring interaction to parse those files, and the standard does not seem to have a Metadata field. Therefore using `getComment()` returns an empty string here also.
- Microphone and Userdata don't have an implementation for an obvious reason.

~~Due to an interrogate limitation, the published C++ functions return `string` which is obtained by joining the vector of comments with newlines. Not sure if there's a better way.~~ no longer the case, see below

I've tested the implementation on Libvorbis which seems to work correctly, did not test FFmpeg and Opus (besides ensuring the engine compiles and can be imported without ABI errors) since there is no documentation on how to force Panda to use a specific cursor loader.

## Checklist

### DIFF
--- a/panda/src/audio/audioSound.cxx
+++ b/panda/src/audio/audioSound.cxx
@@ -198,12 +198,14 @@ operator << (ostream &out, AudioSound::SoundStatus status) {
   return out << "**invalid AudioSound::SoundStatus(" << (int)status << ")**";
 }
 
+static const vector_string empty;
+
 /**
  * Get the comment attached to this AudioSound as a list of strings.
  */
-vector_string AudioSound::
+const vector_string& AudioSound::
 get_raw_comment() const {
-  return vector_string();
+  return empty;
 }
 
 /**
--- a/panda/src/audio/audioSound.h
+++ b/panda/src/audio/audioSound.h
@@ -146,7 +146,7 @@ PUBLISHED:
   std::string get_comment(const std::string &key) const;
   MAKE_MAP_PROPERTY(comments, has_comment, get_comment);
 
-  virtual vector_string get_raw_comment() const;
+  virtual const vector_string& get_raw_comment() const;
 
   int get_num_raw_comments() const;
   std::string get_raw_comment(int index) const;
--- a/panda/src/audiotraits/openalAudioSound.cxx
+++ b/panda/src/audiotraits/openalAudioSound.cxx
@@ -1099,7 +1099,7 @@ status() const {
 /**
  * Returns the comments attached to this audio file.
  */
-vector_string OpenALAudioSound::
+const vector_string& OpenALAudioSound::
 get_raw_comment() const {
   return _comment;
 }
--- a/panda/src/audiotraits/openalAudioSound.h
+++ b/panda/src/audiotraits/openalAudioSound.h
@@ -118,7 +118,7 @@ public:
 
   void finished();
 
-  vector_string get_raw_comment() const;
+  const vector_string& get_raw_comment() const;
 
 private:
   OpenALAudioSound(OpenALAudioManager* manager,


rdb:
Great work, thank you!  Some comments.

We're trying to phase out Python extensions added in direct. They're only available when direct is loaded. Prefer a C++ interface if at all possible, or a C++-implemented extension method (see _ext.h and _ext.cxx files) if nothing else is possible.

Doing it in C++ would mean making sequential access methods (which means you could parse it into a SimpleHashMap, that supports both key-based and indexed access) and exposing those to Python with MAKE_MAP_PROPERTY - do a grep for that to search for examples.  We can keep the C++ methods making up the map-property as unpublished, so that the property is the only way to access it from Python.

An alternative is making an AudioTags class which exposes a mapping interface (overriding `operator []`), and so behaves like a dictionary.  Then we just need a getter property on the AudioSound.  This would be the preferred method if the interface is bigger (like an AudioMetadata containing more than just tags), so that the AudioSound interface isn't cluttered up.  If it's just a dict of tags, I think MAKE_MAP_PROPERTY might be easier.

I don't think that we should have this interface on both AudioSound and MovieAudioCursor.  Intuitively, MovieAudio should be the place for this, but it seems some implementations don't actually open the sound file until the cursor is created, so MovieAudioCursor sounds like the place to get it.  Is there a reason to also have it on AudioSound?

Would it be interesting to give thought to standardizing the keys between different formats, so that you can use the same code to access the title of an .mp3 file as for an .ogg file, via a named enum or named getters, or do you think the original keys should be kept?

Feel free to ping me on Discord for more discussion.

Wizzerinus:
> We're trying to phase out Python extensions added in direct. They're only available when direct is loaded. Prefer a C++ interface if at all possible, or a C++-implemented extension method (see _ext.h and _ext.cxx files) if nothing else is possible.
> Doing it in C++ would mean making sequential access methods (which means you could parse it into a SimpleHashMap, that supports both key-based and indexed access) and exposing those to Python with MAKE_MAP_PROPERTY - do a grep for that to search for examples. We can keep the C++ methods making up the map-property as unpublished, so that the property is the only way to access it from Python.

Implemented this. An interesting difference now is that you can no longer "parse all comments", only "comments with a given named tag". Not sure whether this is a significant issue or not. I could still expose the string as something like `getCommentString()` (since `getRawComment()` is now a vector).

> An alternative is making an AudioTags class which exposes a mapping interface (overriding operator []), and so behaves like a dictionary. Then we just need a getter property on the AudioSound. This would be the preferred method if the interface is bigger (like an AudioMetadata containing more than just tags), so that the AudioSound interface isn't cluttered up. If it's just a dict of tags, I think MAKE_MAP_PROPERTY might be easier.

Decided against it since it felt overkill.

> I don't think that we should have this interface on both AudioSound and MovieAudioCursor. Intuitively, MovieAudio should be the place for this, but it seems some implementations don't actually open the sound file until the cursor is created, so MovieAudioCursor sounds like the place to get it. Is there a reason to also have it on AudioSound?
AudioSound is returned from loader, so I think it only makes sense. Other 2 having it is implementation detail because Panda3D actually has two levels of abstraction you have to go through.

> Would it be interesting to give thought to standardizing the keys between different formats, so that you can use the same code to access the title of an .mp3 file as for an .ogg file, via a named enum or named getters, or do you think the original keys should be kept?
I could implement a series of getters like `getAuthor()`, `getTitle()`, etc. but this seems overkill having on AudioSound and putting them into a separate class isn't much better I think.

Wizzerinus:
Wanted to ask if the maintainers of the engine were still interested in this 🥺 everything here is complete but (since nothing was done in here) I wanted to wait until flac stuff was merged but it seems cancelled so.

panda/src/audio/audioSound.cxx
```diff
@@ -196,3 +196,58 @@ operator << (ostream &out, AudioSound::SoundStatus status) {
...
+  return false;
+}
+
+/*
+ * Returns the value for a given key in the comment. If the key is not present,
+ * returns an empty string.
+ */
+std::string AudioSound::
+get_comment(std::string key) const {
+  for (std::string st : get_raw_comment()) {
```
rdb: Same as above

panda/src/audio/audioSound.h
```diff
@@ -174,6 +182,8 @@ class EXPCL_PANDA_AUDIO AudioSound : public TypedReferenceCount {
   }
   virtual TypeHandle force_init_type() {init_type(); return get_class_type();}
 
+  virtual std::vector<std::string> get_raw_comment() const;
```
rdb: Put this further up, don't put this in the middle of the standard type boilerplate

panda/src/audiotraits/openalAudioManager.h
```diff
@@ -172,6 +172,8 @@ class EXPCL_OPENAL_AUDIO OpenALAudioManager final : public AudioManager {
     int                  _channels;
     int                  _client_count;
     ExpirationQueue::iterator _expire;
+
+    std::vector<std::string> _comment;
```
rdb: This is on the AudioManager, why exactly? Was this a mistake?

panda/src/ffmpeg/ffmpegAudioCursor.cxx
```diff
@@ -490,3 +490,16 @@ read_samples(int n, int16_t *data) {
   _samples_read += n;
   return n;
 }
+
+std::vector<std::string> FfmpegAudioCursor::
```
rdb: Needs docstring

panda/src/ffmpeg/ffmpegAudioCursor.cxx
```diff
@@ -490,3 +490,16 @@ read_samples(int n, int16_t *data) {
...
+get_raw_comment() const {
+  AVDictionaryEntry *tag = NULL;
+  std::vector<std::string> commentStrings;
```
rdb: Coding style nit: `comment_strings`

panda/src/audio/audioSound.cxx
```diff
@@ -196,3 +196,58 @@ operator << (ostream &out, AudioSound::SoundStatus status) {
 
   return out << "**invalid AudioSound::SoundStatus(" << (int)status << ")**";
 }
+
+/**
+ * Get the comment attached to this AudioSound as a list of strings.
+ */
+std::vector<std::string> AudioSound::
```
rdb: Let's use the `vector_string` class everywhere, this is an alias defined in `vector_string.h` and uses Panda allocators

rdb (changes requested):
Yes I am, sorry for not getting to this sooner. I just have some nits to pick and then we can merge this as far as I'm concerned.

panda/src/audio/audioSound.cxx
Wizzerinus: Fixed. Thank you for the assistance 😊 *(no I am not an ai network if it reads like that)*

panda/src/audio/audioSound.cxx
```diff
@@ -196,3 +197,58 @@ operator << (ostream &out, AudioSound::SoundStatus status) {
...
+get_raw_comment() const {
+  return vector_string();
+}
+
+/**
+ * Returns true if this AudioSound has a comment with the given key,
+ * i.e. "author". Case-sensitive.
+ */
+bool AudioSound::
+has_comment(std::string &key) const {
```
rdb: This should be `const std::string &`, requiring a mutable reference is silly and probably wouldn't be wrapped by interrogate.

panda/src/movies/vorbisAudioCursor.cxx
```diff
@@ -17,13 +17,24 @@
...
 using std::istream;
 
 TypeHandle VorbisAudioCursor::_type_handle;
 
+vector_string
+parse_vorbis_comments(vorbis_comment *com) {
+  vector_string comments;
+  for (int cnum = 0; cnum < com->comments; ++cnum) {
+    std::string tag(com->user_comments[cnum], com->comment_lengths[cnum]);
+    comments.push_back(tag);
```
rdb: `std::move(tag)`

rdb (changes requested):
Almost there! :)

A unit test might be a nice extra to have as well, if you find time for that

Wizzerinus:
I feel like this is difficult to avoid. We cannot return a reference to a vector, because in many cases, there's no vector to take a reference to (i.e. userdata sound, NULL manager, etc). Plus, I feel, most sound tracks don't have huge numbers of comments for this to become a performance bottleneck (especially since this API is not likely to be used every frame, or worse, many times every frame).

I should recheck the code flow again, but sound data should only be fetched from the cursors once if they're loaded by OpenAL, and saved into the AudioSound (this is necessary because cursors are thrown away afterwards), so the FFMPEG call should only be happening once per loaded sound.

Wizzerinus:
Would this cause lifetime issues if we call getComment() in Python, then have the sound garbage collected? Or does igate convert things into Python lists anyway?

rdb:
Igate immediately converts it because it doesn't wrap vector_string.

rdb (approved):
Thanks!

Issue #1659, opened 2024-05-21
Implement Khronos' reference PBR tone mapping operator


rdb:
Khronos released a reference PBR tone mapper that gives decent results without excessive need for tweaking, such as is the case with Filmic tone mappers. It might be a good default for Panda3D.

The current ACES tone mapper is fine, but it looks rather desaturated by default and requires application of custom LUTs in order to get good results.

We can add a parameter to the appropriate CommonFilters method to select the tone mapping operator.

I would be inclined to include this in the next minor release, it seems fairly harmless.

See:
https://www.khronos.org/news/press/khronos-pbr-neutral-tone-mapper-released-for-true-to-life-color-rendering-of-3d-products

Community contribution welcome.

raytopianprojects:
This should be easy to implement once #1643 is merged (I need to finish up my changes though). 

The example GLSL is only 13 lines so super easy to add. https://github.com/KhronosGroup/ToneMapping/blob/main/PBR_Neutral/pbrNeutral.glsl

PR #1661, opened 2024-06-04
Add Khronos Tonemap to Common Filters

raytopianprojects:
## Issue description
From #1659
Khronos released a reference PBR tone mapper that gives decent results without excessive need for tweaking, such as is the case with Filmic tone mappers. It might be a good default for Panda3D.

The current ACES tone mapper is fine, but it looks rather desaturated by default and requires application of custom LUTs in order to get good results.

We can add a parameter to the appropriate CommonFilters method to select the tone mapping operator.

## Solution description
Add the Khronos Tonemap to common filters. 
Allows for selecting the tone mapper in  the setHighDynamicRange function using the tonemap parameter (it's set to ACES by default). Add a ToneMap Enum so it is easier to select a tonemap.

## Checklist

### DIFF
--- a/direct/src/filter/CommonFilters.py
+++ b/direct/src/filter/CommonFilters.py
@@ -24,14 +24,14 @@ from panda3d.core import getDefaultCoordinateSystem, CS_zup_right, CS_zup_left
 
 from direct.task.TaskManagerGlobal import taskMgr
 
-from direct.filter.FilterManager import FilterManager
-from direct.filter.filterBloomI import BLOOM_I
-from direct.filter.filterBloomX import BLOOM_X
-from direct.filter.filterBloomY import BLOOM_Y
-from direct.filter.filterBlurX import BLUR_X
-from direct.filter.filterBlurY import BLUR_Y
-from direct.filter.filterCopy import COPY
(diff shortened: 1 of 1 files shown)


rdb:
Thanks, committed with some fixes

direct/src/filter/CommonFilters.py
```diff
@@ -383,8 +392,32 @@ def reconfigure(self, fullrebuild, changed):
...
+                        text += "if (peak < start_compression){ "
+                        text += "o_color.rgb =  neutral_color;"
+                        text += "}else{ "
+
+                        text += "const float d = 1.0 - start_compression;"
+                        text += "float new_peak = 1.0 - d * d / (peak + d - start_compression);"
+                        text += "neutral_color *= new_peak / peak;"
+                        text += "float g = 1.0 - 1.0 / (desaturation * (peak - new_peak) + 1.0);"
+
+                        text += "o_color.rgb = mix(neutral_color, new_peak * float3(1, 1, 1), g);}"
```
rdb, line 420: For future reference, mix() is not a Cg function, it's called lerp() in Cg

Issue #1666, opened 2024-06-19
TextNode 'set_small_caps' does not work on 1.11.0


Benjamin8693:
## Description
The 'set_small_caps' method of TextNode does not work. It does set the 'small_caps' field successfully (evident if you call 'get_small_caps'), but it does not have any visual impact on text once generated. This works as intended on 1.10.14.

Functional example:
![2024-06-18_21-27](https://github.com/panda3d/panda3d/assets/11221724/b8a81fda-1f32-4bcc-b217-cd8865b27701)

Non-functional example:
![2024-06-18_21-28](https://github.com/panda3d/panda3d/assets/11221724/5e766e82-8421-426c-804b-3630b4aeb023)

## Steps to Reproduce

1. Create this file:

```py
from direct.showbase.ShowBase import ShowBase
from panda3d.core import NodePath, TextNode

class Demo(ShowBase):

    def __init__(self):
        ShowBase.__init__(self)

        styling = TextNode('status')
        styling.set_align(TextNode.ACenter)
        styling.set_text_color(1, 1, 1, 1)
        styling.set_shadow(0.075, 0.075)
        styling.set_small_caps(True)
        styling.set_text("Testy McTest!")

        letter = NodePath(styling.generate())
        letter.reparent_to(aspect2d)
        letter.set_scale(0.25)

demo = Demo()
demo.run()
```
2. Run it
3. Note that the lowercase letters have not changed as expected ([see documentation](https://docs.panda3d.org/1.11/python/reference/panda3d.core.TextNode#panda3d.core.TextNode.set_small_caps))

## Environment
 * Operating system: Linux Mint 21.3
 * System architecture: x64
 * Panda3D version: 1.11.0 latest master
 * Installation method: pip
 * Python version (if using Python): 3.10.12
 * Compiler (if using C++): N/A

rdb:
Feel free to look into it, thank you!

PR #1670, opened 2024-07-02
text: add 'small_caps' functionality to harfbuzz

derpydraggo:
## Issue description

This commit attempts to resolve issue #1666 (which was introduced when building with harfbuzz after commit 0dad38d), by adding code in TextAssembler.shape_buffer(), to handle the case where:
1. harfbuzz is built with the engine
2. is wanted via the config variable `text-use-harfbuzz` being set to true, and
3. the TextProperties associated with the to-be-drawn text also has the small_caps field set.

## Solution description

If the above cases are all true, then the buffer is first remade with all harfbuzz 'codepoints' switched to the uppercase variant (if needed and possible), while keeping the harfbuzz 'cluster' value the same for tracking what characters need their scaling changed. 
Later when the glyphs are actually in the process of being generated, I check to see if those earlier 'cluster' values are lowercase values, and go ahead and use individual scale values for each character, using `get_small_caps_scale()` for these characters instead of the usual `get_glyph_scale()` in every other case.

## Checklist

### DIFF
--- a/panda/src/text/textAssembler.cxx
+++ b/panda/src/text/textAssembler.cxx
@@ -1781,6 +1781,76 @@ shape_buffer(hb_buffer_t *buf, PlacedGlyphs &placed_glyphs, PN_stdfloat &xpos,
              const TextProperties &properties) {
 
 #ifdef HAVE_HARFBUZZ
+  // before we do *anything*, check to see if small_caps == true, and change the characters before
+  // setting up everything else; check to see if small_caps == true, and change "glyph_index" if so
+  // (glyph_index *should* be the character, since we've not told harfbuff what to do with it yet)
+  // but first, we need some of these defined sooner rather than later:
+  DynamicTextFont *font = DCAST(DynamicTextFont, properties.get_font());
+
+  if (properties.has_small_caps()) {
+    // create a temporary buffer to iterate through
+    hb_buffer_t *tmp_harfbuff = nullptr;
+    tmp_harfbuff = hb_buffer_create();
+    // copy buf into our temp one, and reset it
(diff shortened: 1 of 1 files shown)


panda/src/text/textAssembler.cxx
```diff
@@ -1781,6 +1781,76 @@ shape_buffer(hb_buffer_t *buf, PlacedGlyphs &placed_glyphs, PN_stdfloat &xpos,
              const TextProperties &properties) {
 
 #ifdef HAVE_HARFBUZZ
+  // before we do *anything*, check to see if small_caps == true, and change the characters before
+  // setting up everything else; check to see if small_caps == true, and change "glyph_index" if so
+  // (glyph_index *should* be the character, since we've not told harfbuff what to do with it yet)
+  // but first, we need some of these defined sooner rather than later:
+  DynamicTextFont *font = DCAST(DynamicTextFont, properties.get_font());
+
+  if (properties.has_small_caps()) {
```
rdb, line 1790: This checks whether the property has been set, not whether it's actually set to true (use `get_small_caps`).

rdb (commented):
Thanks for working on this!

Should we not try and pass the `smcp` feature to `hb_shape` to use the native small caps support if provided by the font?

rdb:
No worries.  I'll finish the work and push it shortly.  Will be part of 1.10.15.  Thanks for looking into all this!

I looked into smcp and only some fonts supported it.  The first font I tried (Inter) actually only supported it for some glyphs, creating a strange result.  So, I think we're just going to do our own scaling, and maybe revisit smcp in the future.

PR #1677, opened 2024-08-09
pythonTask: fix refcount leak of non-panda Future::done

johnnyapol:
## Issue description
A missing Py_DECREF on the future's "done" method
caused both the bound method and the underlying
self instance of the future to be leaked when
awaiting non-panda futures (such as _asyncio.Future).

This change includes a simple new test addition
to catch this in the future.

## Solution description
Add the missing Py_DECREF call and a new unit-test to cover this area.

## Checklist


rdb (approved):
Thank you!

PR #1679, opened 2024-09-08
event: fix result refcount leak when awaiting AsyncFuture

johnnyapol:
## Issue description
Raising the StopIteration exception left the underlying result with an elevated refcount which prevented it from being cleaned up.

Running the new test (test_await_future_result_cleanup) on current master results in the result object of an AsyncFuture (or associated friends like AsyncGatheringFuture and PythonTask) being left with a permanently elevated refcount introduced when `await`ing the future.

## Solution description
Properly account for PyErr_SetObject raising the refcount of `result` from get_done_result and add a test to help cover this path for the future.

## Checklist


rdb (approved):
Thanks!

PR #1681, opened 2024-09-12
event: use the persistent python wrapper when appending task object

### LINKED ISSUE #1680: Debugging code inside a task often crashes the program
## Description
Using the extension Python Debugger in Visual Studio Code, debugging code running in a Python Task is likely to crash the program with the AssertionError `AssertionError: _function != nullptr at line 589 of panda/src/event/pythonTask.cxx`. While this is likely to happen, the crash does not consistently happen, suggesting the presence of some sort of race condition.

## Steps to Reproduce
Using Visual Studio Code and the Python Debugger extension (which can be installed/managed using the Ctrl+Shift+X hotkey), the following code can be run using the "Python Debugger: Python File" run and debug command:

```py
from direct.task.TaskManagerGlobal import taskMgr, Task

class GameBase(ShowBase):
    def run(self):
        taskMgr.doMethodLater(1.0, self.update, "update_task")
        super().run()

    def update(self, task):
        print("Hello World")
        return Task.again

game = GameBase()
game.run()
```

By placing a breakpoint on a line in update (ex: `print("Hello World")`) and pressing F5 to continue whenever the program pauses, the program will likely crash within 1-20 continues with either the AssertionError above, or just completely silently. This error is far more likely to happen with more complicated code than the provided snippet, making debugging an actual project impossible.

## Environment
 * Operating system: Windows 10 Professional
 * System architecture: x86 Intel 9th gen
 * Panda3D version: 1.11.0.dev3444
 * Installation method: pip
 * Python version (if using Python): Python 3.12
 * Compiler (if using C++): N/A

johnnyapol:
"event: use the persistent python wrapper when appending task object to args"

## Issue description

Seems to fix #1680 but also this is a nice optimization regardless since we no longer need to create a new python wrapper object every time we call into python.

Also fixes subclassing of PythonTask not passing in the subclass instance to the task function.

## Solution description
Don't create a new wrapper object if one exists when appending the task object to the function arguments, instead pass in a new reference to the existing `__self__` object

## Checklist

### DIFF
--- a/panda/src/event/pythonTask.cxx
+++ b/panda/src/event/pythonTask.cxx
@@ -1018,7 +1018,7 @@ call_function(PyObject *function) {
       __self__ = DTool_CreatePyInstance(this, Dtool_PythonTask, true, false);
     }
 
-    PyObject *result = PyObject_CallOneArg(function, Py_NewRef(__self__));
+    PyObject *result = PyObject_CallOneArg(function, __self__);
     Py_XDECREF(result);
   }
 }
--- a/panda/src/event/pythonTask.cxx
+++ b/panda/src/event/pythonTask.cxx
@@ -175,8 +175,16 @@ get_args() {
     }
 
     this->ref();
-    PyObject *self = DTool_CreatePyInstance(this, Dtool_PythonTask, true, false);
-    PyTuple_SET_ITEM(with_task, num_args, self);
+
+    // Check whether we have a Python wrapper.  This is not the case if the
+    // object has been created by C++ and never been exposed to Python code.
+    if (__self__ == nullptr)
+    {
+      // A __self__ instance does not exist, let's create one now.
+      __self__ = DTool_CreatePyInstance(this, Dtool_PythonTask, true, false);
+    }
+
+    PyTuple_SET_ITEM(with_task, num_args, Py_XNewRef(__self__));
     return with_task;
   }
   else {
--- a/panda/src/event/pythonTask.cxx
+++ b/panda/src/event/pythonTask.cxx
@@ -178,13 +178,12 @@ get_args() {
 
     // Check whether we have a Python wrapper.  This is not the case if the
     // object has been created by C++ and never been exposed to Python code.
-    if (__self__ == nullptr)
-    {
+    if (__self__ == nullptr) {
       // A __self__ instance does not exist, let's create one now.
       __self__ = DTool_CreatePyInstance(this, Dtool_PythonTask, true, false);
     }
 
-    PyTuple_SET_ITEM(with_task, num_args, Py_XNewRef(__self__));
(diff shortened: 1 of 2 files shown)


johnnyapol:
Worth noting: without this change, this means subclasses of PythonTask were not being passed into the function but rather a new PythonTask instance, so this change also fixes that.

johnnyapol:
Pushed up some new changes:
1. In the event a PythonTask was created directly from C++ (e.g. future done callbacks), we now create the `__self__` wrapper object.
2. Added a unit-test to verify subclassing, reusing `__self__` work as expected.

panda/src/event/pythonTask.cxx
```diff
@@ -175,8 +175,16 @@ get_args() {
     }
 
     this->ref();
-    PyObject *self = DTool_CreatePyInstance(this, Dtool_PythonTask, true, false);
-    PyTuple_SET_ITEM(with_task, num_args, self);
+
+    // Check whether we have a Python wrapper.  This is not the case if the
+    // object has been created by C++ and never been exposed to Python code.
+    if (__self__ == nullptr)
+    {
```
rdb, line 182: Nit: `{` should be on same line

panda/src/event/pythonTask.cxx
```diff
@@ -175,8 +175,16 @@ get_args() {
...
+      // A __self__ instance does not exist, let's create one now.
+      __self__ = DTool_CreatePyInstance(this, Dtool_PythonTask, true, false);
+    }
+
+    PyTuple_SET_ITEM(with_task, num_args, Py_XNewRef(__self__));
```
rdb, line 187: This increments the refcount a second time if `__self__` was created. Or was that intended?

You can also use Py_NewRef, since it will never be nullptr

rdb (changes requested):
Thank you!

panda/src/event/pythonTask.cxx
johnnyapol: Good question, I think we do need `__self__` to keep its initial reference and not have it stolen by the PyTuple_SET_ITEM call. The task could run multiple times via task.again and having the underlying `__self__` object being destructed before the task finishes could cause a premature clearing of everything cleared via tp_clear / `__clear__`.

No contesting on using Py_NewRef though, I'll fix that and the above style suggestion very soon.

johnnyapol: I've made some additional fixes -- I found that this was going to break the reference counting on the C++ side since my change was calling ref() even with we weren't making an additional DTool object instantiation. That has been fixed so we will only call `ref()` prior to creating  `__self__`. 

I also found that owner methods (e.g. _addTask/_clearTask) and the upon_death callbacks were not using the persistent python wrappers. That has been fixed and I've amended my test to check for correctness with both of those. I think this is good to go now unless there's any other issues

rdb (approved):
Thanks!

PR #1684, opened 2024-09-20
when pipelining off `from._cycler` may not exist

pmp-p:
found when building with em++/3.13t


rdb:
That should be `from._pointer` of course, I'll check that in shortly.

PR #1693, opened 2024-10-16
add rgba support using xrender draft

Yonnji:
Adds support for transparent windows on X11 when using compositing like this:
https://www.youtube.com/watch?v=3zWK6owqPe0
It works for makepanda and cmake builds.
It's a draft because I don't know how to make the Xrender dependency optional yet.


rdb:
Great!  I also have some code to add transparent windows on macOS lying around somewhere, would be a cool feature to have.  We might have to think about how alpha blending works, when I tested it on macOS there was some funny stuff going on with the window background color.

One way to deal with not creating a dependency is by loading Xrender the same way we load Xrandr dynamically, with dlopen--see x11GraphicsPipe.cxx for that.  Then we just need the headers to be present during build, which is okay.

PR #1699, opened 2024-10-30
draft: add steam audio support

lifelandman:
## Checklist


lifelandman:
Hmm... Steam Audio allows for effect settings to be set per-source through a different interface than the global effect options. That said, the options provided are 1-1 with the global effect options, so I'm wondering if I should write a "match_source_parameters_to_effect" function, which would take a SteamAudioProcessor (which handles all of its sources privately), the name of the source, and an effect to copy the parameters from. It's not exactly intuitive, but it means that any extra copies of SteamAudioEffect classes needed for per-source configuration could be destroyed.

Issue #1689, opened 2024-10-12
Prevent window size from being set to zero


rdb:
Reported on behalf of a forum user:
https://discourse.panda3d.org/t/window-size-singularity-in-windows-7/30351

## Description
Prevent the window from being resized to a height or width of 0.

## Use Case
This prevents the occurrence division-by-zero errors, creation of invalid textures, etc.

kekwlboy12469:
Hey can i contribute to this issue  ?

burningbuffer:
I would like to contribute.

This issue also occurs on Windows 10. I solved something very similar in one of my applications. Could you assign it to me?

rdb:
Feel free to contribute, of course. See the [Contributing Guidelines](https://github.com/panda3d/panda3d/blob/master/CONTRIBUTING.md). Maybe suggest your implementation strategy here first or discuss it in the [Discord server](https://discord.gg/UyepRMm) in #engine-dev, before working on a pull request. That is especially helpful to coordinate if you both want to work on it so you don't do duplicate work. :-)

rdb:
I tested that macOS indeed already prevents a window from being zero-sized. The only remaining case is Linux; we need to test the current behaviour for an X11 window and possibly adjust the size hints we already give the WM to set a particular min size.

burningbuffer:
> I tested that macOS indeed already prevents a window from being zero-sized. The only remaining case is Linux; we need to test the current behaviour for an X11 window and possibly adjust the size hints we already give the WM to set a particular min size.

I tested with a default x11 application and a panda3d sample in linux. Looks like the default is already width  = 1 x height = 1.
![Image](https://github.com/user-attachments/assets/4da676d4-b6a6-4744-9ba0-54e491f247d2)

PR #1704, opened 2024-12-16
FIXED ISSUE ON MINIMUM WIDTH

kekwlboy12469:
## Issue description

Prevent window size from being set to zero #1689

## Solution description

Added a function to make sure the window size is never too small, specifically not smaller than 1x1 pixels. Created a new method to set the window size, ensuring that the width and height always meet the minimum size requirement. Additionally, updated the setters for width and height to ensure they are never set to zero or negative values. This prevents errors and ensures valid window sizes.

## Checklist

### DIFF
--- a/panda/src/display/displaySearchParameters.cxx
+++ b/panda/src/display/displaySearchParameters.cxx
@@ -26,12 +26,18 @@ DisplaySearchParameters::
  */
 DisplaySearchParameters::
 DisplaySearchParameters() {
-  _minimum_width = 640;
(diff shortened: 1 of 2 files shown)


rdb (changes requested):
DisplaySearchParameters is an obscure class used in the DirectX code, this isn't really related to window sizes, so this does not address the problem at all.

Furthermore there are issues with the code that make it unable to compile (typo's, missing method from header).

kekwlboy12469:
okay will work on that

PR #1709, opened 2024-12-21
wasi sdk 25 cpp build test

pmp-p:
first attempt to build with https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-25


rdb:
What's the status of this PR?

I think it would be better to use PHAVE_TEMPNAM etc. kind of macros to eliminate the use of these functions rather than stubbing them out.

PR #1720, opened 2025-02-01
showbase: Annotate callback methods

WMOkiishi:
## Change description
These changes annotate the callback methods (and methods called by those methods) on `ShowBase` that are used by code that is already type-hinted. Of note is the fact that many of these take a second argument that is unused. I have elected to annotate these arguments with `object` to signify that any object can be passed here, but something like `AsyncTask` or `AsyncTask | None` is also an option.

## Checklist


rdb (approved):
Thanks. Doesn't really matter since they are just internal callbacks that are not part of the public interface, though they really are tasks so AsyncTask would also be acceptable.

PR #1723, opened 2025-02-16
showbase: Annotate basic `Loader` methods

WMOkiishi:
## Change description
These changes annotate methods of `Loader` that are used in the test suite or other type-annotated code. There is a way to use overloads to describe the relationship between parameter and return types for `loadModel`, but it's [quite verbose](https://github.com/WMOkiishi/types-panda3d/blob/cfad0b9aa00607c16cad95e0afd9a2fd5d454f73/src/direct-stubs/showbase/Loader.pyi#L69-L207), so I've just annotated the return type as `Any`.

There is also a slight change to the tests so that `Loader.destroy()` is also tested.

## Checklist

### DIFF
--- a/direct/src/showbase/Loader.py
+++ b/direct/src/showbase/Loader.py
@@ -2,9 +2,13 @@
 sound, music, shaders and fonts from disk.
 """
 
+from __future__ import annotations
+
 __all__ = ['Loader']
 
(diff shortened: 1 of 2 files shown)


rdb:
Thanks!

Issue #1691, opened 2024-10-14
Dropping Windows 7 and/or Python 3.8?


rdb:
Python 3.8 just went EOL.  That was the last Python version to support Windows 7. Just following upstream, like we do normally, therefore would mean we are dropping Windows 7.

I'm opening this thread to enquire whether there are still users who need to target Windows 7, or whether we can safely drop it.

Continuing to support Windows 7 would require us to either:
1. Continuing to support Python 3.8, even though it is EOL.
2. Using a build of [Python that is patched to maintain Windows 7 support](https://github.com/adang1345/PythonWin7). This would allow users to target Windows 7 even with the latest versions of Python.

At the moment, we do not have a pressing need to drop support for either Windows 7 or Python 3.8, although the Python 3.8 builds take up space and build time, of course, so I'm not really happy about having those wheels around indefinitely.

If you need support for Windows 7 or Python 3.8 to be maintained, please state your case.

Maxwell175:
I think that once the time does come to drop 3.8, it would be reasonable to go for option 3, which is let the end user figure it out. If you could just not remove the workarounds you currently have for Windows 7 APIs, it would be the responsibility of people still interested to potentially contribute any workarounds in the future and it would be on them to compile/patch newer versions of Python.

ArsThaumaturgis:
Looking at the [Steam Hardware Survey](https://store.steampowered.com/hwsurvey) for this year, I see that Windows 7 users (as surveyed) account for 0.31% of the total number, and that this number has declined from last year.

I'm thus comfortable with support for Windows 7 and Python 3.8 to be dropped.

(I don't require or request it, to be clear--I'm just not opposed to it.)

DylanCheetah:
Would it be possible to keep things so the engine could still be able to be built on older platforms but just reduce the number of pre-built wheels available?

For reference, I still have a Windows Vista laptop that I use as a spare and the newest version of Python that works on it is Python 3.6. It would be nice if it would at minimum remain possible to build newer versions of Panda3D myself.

It seems like a reasonable compromise.

DylanCheetah:
> Would it be possible to keep things so the engine could still be able to be built on older platforms but just reduce the number of pre-built wheels available?
> 
> For reference, I still have a Windows Vista laptop that I use as a spare and the newest version of Python that works on it is Python 3.6. It would be nice if it would at minimum remain possible to build newer versions of Panda3D myself.
> 
> It seems like a reasonable compromise.

[...]

PR #1726, opened 2025-02-20
showbase: Annotate basic `Messenger` methods

WMOkiishi:
## Change description
These changes annotate `Messenger` and `DirectObject` methods that are called from tests or other type-annotated code. I created type aliases in the `Messenger` file, which I named based on how they're referred to elsewhere in the code, because the types are otherwise very obtuse.

## Checklist


rdb:
We haven't decided to drop Python 3.8 support yet, that discussion is still open at #1691.

PR #1727, opened 2025-02-20
direct: Annotate functions called directly by top-level code

WMOkiishi:
## Change description
These changes annotate functions and methods that are called directly by top-level code elsewhere in the library, found by running mypy with the `--disallow-untyped-calls` option. This means that they are run by the test suite (since they run when the file is loaded) and their usage is covered by a type checker (since mypy checks top-level code and annotated functions).

## Checklist

### DIFF
--- a/direct/src/showbase/PythonUtil.py
+++ b/direct/src/showbase/PythonUtil.py
@@ -41,13 +41,17 @@ import time
 import builtins
 import importlib
 import functools
-from typing import Callable
+from collections.abc import Callable, Container, Iterable, Mapping
+from typing import Any, Generic, TypeVar
 
 __report_indent = 3
 
 from panda3d.core import ConfigVariableBool, ConfigVariableString, ConfigFlags
 from panda3d.core import ClockObject
 
+_T = TypeVar('_T')
+_KT = TypeVar('_KT')
+_VT = TypeVar('_VT')
 
 ## with one integer positional arg, this uses about 4/5 of the memory of the Functor class below
 #def Functor(function, *args, **kArgs):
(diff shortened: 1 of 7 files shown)


direct/src/showbase/ShowBase.py
```diff
@@ -382,7 +382,7 @@ def __init__(self, fStartDirect: bool = True, windowType: str | None = None) ->
         #: yourself every frame.
         self.cTrav: CollisionTraverser | Literal[0] = 0
         self.shadowTrav: CollisionTraverser | Literal[0] = 0
-        self.cTravStack = Stack()
+        self.cTravStack = Stack[CollisionTraverser]()
```
rdb, line 385: How does this work? Does Stack implement `__getitem__`? Should this not just be an annotation?

direct/src/showbase/PythonUtil.py
```diff
@@ -1972,7 +1976,13 @@ def wrap(*args, **kw):
...
 
 
-def report(types = [], prefix = '', xform = None, notifyFunc = None, dConfigParam = []):
+def report(
+    types: Container[str] = [],
+    prefix: str = '',
+    xform: Callable[[Any], object] | None = None,
+    notifyFunc: Callable[[str], object] | None = None,
+    dConfigParam: str | list[str] | tuple[str, ...] = [],
+) -> Callable[[_T], _T]:
```
rdb, line 1985: This makes the code a lot less readable, but this is PythonUtil so I don't really care too much

WMOkiishi: Interesting! I find parameter annotations almost universally a boost to readability. Definitely let me know if this comes up elsewhere in the future—I wouldn't want to impose something that I find helpful if it's distinctly unhelpful for others.

Issue #1478, opened 2023-03-16
Oscillating `button_down` state on key repeat


Schwarzbaer:
## Description
The code below should turn the screen green when `q` is pressed, and red when it isn't, and on 1.10.13, it does so. On 1.11.0.dev3152 however, once key repeat sets in, the apparent key state seems to oscillate between `True` and `False` frame by frame.

## Steps to Reproduce
```
import sys

from panda3d.core import Vec4
from panda3d.core import ButtonRegistry
from direct.showbase.ShowBase import ShowBase


def polling_task(task):
    if base.mouseWatcherNode.is_button_down(ButtonRegistry.ptr().find_button('q')):
        base.win.setClearColor(Vec4(0, 1, 0, 1))
    else:
        base.win.setClearColor(Vec4(1, 0, 0, 1))
    return task.cont


ShowBase()
base.task_mgr.add(polling_task)
base.accept('escape', sys.exit)
base.run()
```

## Environment
 * Operating system: Debian Linux
 * System architecture: x86_64
 * Panda3D version: 1.11.0.dev3152
 * Installation method: pip
 * Python version: Python 3.9.9

johnnyapol:
I'm able to reproduce this as well. I had to install an ubuntu 22.04 VM (with GNOME) to observe this (for some reason it doesn't happen on my preferred distributions).

I performed a git bisect which indicates https://github.com/panda3d/panda3d/commit/4cd579d0ba86dc952fb0acdef107a1d52944fa8b is the culprit commit.

The parameters for git bisect was GOOD=v1.10.13, BAD=53f32e440e3d21ff8148119770b9016a64f4d1b8 (HEAD of master as of time time of posting)

rdb:
Please tell me more about your X11 configuration and which input methods you're using, I can't reproduce it with X11 on ArchLinux using uim.

johnnyapol:
When I reproduced it and bisected it, I used a stock Ubuntu 22.04 VM with the default GNOME on Wayland desktop environment (panda running in XWayland). I only type English (US) using a regular US keyboard layout so I don't have any special input methods installed or configured. 

It does not happen for me on my regular installs which are Arch Linux using GNOME on Wayland nor Xfce4 (X11).

Schwarzbaer:
"rdb: But it looks like you're using Wayland with GNOME and Mutter, which is critical info! Please add it to the issue."

```
$ ps ax | grep -i wayland
[...]

PR #1735, opened 2025-03-11
Add function for detecting auto repeate key events from i.e accessibi…

omn14:
…lity settings

## Issue description

Attempts to solve [#1478 ](https://github.com/panda3d/panda3d/issues/1478)
Use [XkbSetDetectableAutoRepeat](https://linux.die.net/man/3/xkbsetdetectableautorepeat) to avoid auto-repeat generated by OS accessibility settings, you need to call this function to enable detectable auto-repeat. This will allow to distinguish between key press events generated by actual key presses and those generated by auto-repeat.

## Solution description

By calling [XkbSetDetectableAutoRepeat](https://linux.die.net/man/3/xkbsetdetectableautorepeat) with True, we enable detectable auto-repeat, which allows to distinguish between actual key presses and auto-repeat events. The supported variable will indicate whether the X server supports detectable auto-repeat. If it is not supported, we may need to handle auto-repeat events manually.
## Checklist

### DIFF
--- a/panda/src/x11display/config_x11display.cxx
+++ b/panda/src/x11display/config_x11display.cxx
@@ -91,6 +91,10 @@ ConfigVariableBool x_send_startup_notification
           "lets the window manager know that an application has launched, so "
           "that it no longer needs to display a spinning mouse cursor."));
 
+ConfigVariableBool x_detectable_auto_repeat
+("x-detectable-auto-repeat", false,
+ PRC_DESC("Set this true to enable detectable auto-repeat for keyboard input."));
+
 /**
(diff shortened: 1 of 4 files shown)


rdb (changes requested):
I approve in principle, but would like this gated by a config.prc variable check. See `config_x11display.cxx` and add something like `x-detectable-auto-repeat` or something to that effect, and check it before setting it.

Thanks!

omn14:
Thank you rdb for your review. I have added the requested config.prc variable check. I have the default set to not enable the detectable auto repeat. I also have the PRC description very minimal. Let me know if I should change any thing. O.

Issue #1468, opened 2023-02-23
Expose Native Buffers ID for OpenGL-CUDA Interoperability 


QuanyiLi:
## Description
As discussed in https://discourse.panda3d.org/t/allowing-torch-tensorflow-to-directly-access-rendered-image-in-gpu/29119/7, it is possible to do interoperation between OpenGL and CUDA with native buffer ID. However, the related Python API lacks in Panda. Could you please provide such an interface?

## Use Case

I am testing OpenGL and CUDA interoperability with pure Python interfaces provided by Panda3D and CuPY, respectively. Such a feature allows nowadays deep learning frameworks, like Tensorflow/PyTorch to manipulate the rendered content on GPU, which could improve the efficiency of visual data collection and model training by a large margin.

rdb:
We just need to add a virtual get_native_id method to BufferContext, similar to how we have it in TextureContext, and implement it in the various buffer-related GL subclasses.

This could then easily be supported for vertex buffers, index buffers, and SSBOs.

Spartan-X1:
Hi @rdb, I'm interested in working on this issue. I'm an absolute beginner to open source, but I have experience with in C++ & Python and am eager to learn more about graphics programming. Since this issue is from 2023, I wanted to check if it's still available for contribution. I'd appreciate any guidance on getting started, particularly about the BufferContext implementation you mentioned.

rdb:
Yes, feel free to work on this. You can look at the existing get_native_id code and implement it for the buffer-related classes.

Spartan-X1:
Thank you for the opportunity, I'll study the existing get_native_id implementation in TextureContext and implement it for the buffer-related GL subclasses as suggested.

Spartan-X1:
Hello @rdb , I am sorry to bother you again but I just need to expose the OpenGL native ID in just these files right :
glIndexBufferContext_src ,  glBufferContext_src & glVertexBufferContext_src  & a base file code for BufferContext ?

rdb:
Yup, that sounds about right!

Spartan-X1:
I made a PR , please review it & if something is wrong please tell me

rdb:
I pushed support for `get_native_id` for all relevant resources (sampler objects, IBOs, VBOs, SSBOs)

PR #1736, opened 2025-03-13
Exposing the native OpenGL ID (issue ##1468)

Spartan-X1:
## Issue description
This change exposes the native OpenGL buffer ID via a virtual method in BufferContext. It allows external tools and libraries to interoperate with GPU buffers without copying data between system memory and GPU memory
( Issue #1468 )

## Solution description
This PR adds a new method virtual uint64_t get_native_id() const override; to the BufferContext class which returns the contents of the _index member cast to a uint64_t. The method provides direct access to the native OpenGL buffer ID, enabling efficient interoperability with external libraries like CUDA, TensorFlow, and PyTorch. The implementation adheres to Panda3D's existing design patterns and coding style.

## Checklist

### DIFF
--- a/panda/src/glstuff/glBufferContext_src.cxx
+++ b/panda/src/glstuff/glBufferContext_src.cxx
@@ -47,3 +47,10 @@ evict_lru() {
   update_data_size_bytes(0);
   set_resident(false);
 }
+
+/**
+ * Returns the native OpenGL buffer ID for this buffer.
+ */
+uint64_t CLP(BufferContext)::get_native_id() const {
+  return static_cast<uint64_t>(_index);
+}
--- a/panda/src/gobj/bufferContext.cxx
+++ b/panda/src/gobj/bufferContext.cxx
@@ -17,6 +17,27 @@
 TypeHandle BufferContext::_type_handle;
 
 /**
+ * Returns an implementation-defined handle or pointer that can be used
+ * to interface directly with the underlying API.
+ * Returns 0 if the underlying implementation does not support this.
+ */
+uint64_t BufferContext::
+get_native_id() const {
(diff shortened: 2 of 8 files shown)


panda/src/glstuff/glBufferContext_src.cxx
```diff
@@ -47,3 +47,10 @@ evict_lru() {
   update_data_size_bytes(0);
   set_resident(false);
 }
+
+/**
+ * Returns the native OpenGL buffer ID for this buffer.
+ */
+uint64_t CLP(BufferContext)::get_native_id() const {
```
rdb, line 54: Nitpick: Panda style is to put the part starting with `get_native_id` on a new line. See the other definitions in this file.

panda/src/gobj/bufferContext.cxx
```diff
@@ -16,6 +16,27 @@
...
+}
+
+/**
+ * Similar to get_native_id, but some implementations use a separate
+ * identifier for the buffer object associated with buffer textures.
+ * Returns 0 if the underlying implementation does not support this, or
+ * if this is not a buffer texture.
+ */
+uint64_t BufferContext::
+get_native_buffer_id() const {
```
rdb, line 36: There's no need to have a separate `get_native_buffer_id` (which isn't implemented anyway) for buffers since the `get_native_id` already returns the buffer id. The distinction only makes sense for textures.

rdb (changes requested):
Thanks for working on this! Sorry for the delay in responding.

Looks great. I just have two nitpicks. If you fix that and the CI passes this is good to go.

rdb:
Finished this myself

Issue #1747, opened 2025-05-11
package_option does not take library search dirs into account


mucki-at:
## Description

Homebrew is a common way to install 3rd party packages on macOS. It is essentially a package manager, similar to apt, etc ... on linux. A big difference is that homebrew does not usually install libraries and headers in /usr/include and /usr/lib, but in /opt/homebrew/packagename/lib. Therefore linking often requires a library search directory specified.

Unfortunately package_option does not take the library search dirs found by the find_package() make functionality into account when creating the interface library. This leads to link errors during the build process for all libraries which are not in the default library search path (/usr/lib, /usr/local/lib).

## Steps to Reproduce
 1. Checkout sorce on macOS
 2. install homebrew (https://brew.sh)
 3. install gtk+3 using homebrew: "brew install gtk+3"
 4. configure and build panda3d as normal
 5. notice build errors when trying to link any app requiring gtk (pview, pstats-gtk, ...)

## Environment
 * Operating system: macOS
 * System architecture: arm64
 * Panda3D version: latest master
 * Installation method: built from source
 * Python version (if using Python): 3.13.x
 * Compiler (if using C++): clang 17

PR #1748, opened 2025-05-11
Support 3d party dependencies on Mac using homebrew

mucki-at:
## Issue description

Issue #1747

Homebrew is a common way to install 3rd party packages on macOS. It is essentially a package manager, similar to apt, etc ... on linux. A big difference is that homebrew does not usually install libraries and headers in /usr/include and /usr/lib, but in /opt/homebrew/packagename/lib. Therefore linking often requires a library search directory specified.

## Solution description
This patch fixes the package_option make function to add the library search path to the interface library it creates, so that linking works with libraries from homebrew on macOS.

## Checklist

### DIFF
--- a/cmake/macros/PackageConfig.cmake
+++ b/cmake/macros/PackageConfig.cmake
@@ -223,6 +223,14 @@ function(package_option name)
 
       target_link_libraries(PKG::${name} INTERFACE ${libs})
 
+      if(DEFINED ${found_as}_LIBRARY_DIRS)
+        set(libdir ${${found_as}_LIBRARY_DIRS})
+      elseif(DEFINED ${found_as}_LIBDIR)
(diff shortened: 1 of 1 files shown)


rdb:
In this code, what happens if neither is set?

mucki-at:
Then either the library does not need a special library search path, or cmake failed to find the library dependency properly (most likely it is missing), or the library is using variables incompatible with cmake norms.

The first case is benign, the other two lead to build errors, but are most likely a user specific problem (libraries not installed correctly or other system config issue). They will need to use the prepackaged libraries method.

mucki-at:
Sorry, to be more specific to your question: if the variable is not found then nothing happens (the behavior before the patch is unchanged).

PR #1759, opened 2025-07-12
Add a Default Release Sound for DirectButtons

DTM1218:
## Issue description
While DirectButtons have an audio state in mind for hovering over and pressing a button, there isn't a native option for release sounds.  If someone wanted to develop an interface that has a final sound for when the button is let go, they would have to inherit from DirectButton only to include a binding for release sounds.  I feel like including the option for default release sounds would be consistent with rollover and press.

## Solution description
Simply put, I set up a binding just like `setClickSound()` called `setReleaseSound()` which handles the release events for the 3 mouse button bindings.  I also added a global to DirectGUIGlobals that mirrors the other default sounds.  (As a small bit of "tidying", I also arranged the sound functions in matching order.)

## Checklist


rdb (approved):
LGTM

PR #1766, opened 2025-08-05
Fix 3D audio direction in y-up-left coordinate system

Tanay-ErrorCode:
## Issue description

The 3D audio positioning was wrong when using the `y-up-left` coordinate system. That’s because the code assumed a right-handed system, so sounds ended up flipped on the wrong side in left-handed mode.

## Solution description

This fixes it by checking which coordinate system is in use. If it’s left-handed, we flip the X-axis when sending positions to OpenAL so the audio ends up in the right place. It works the same as before for right-handed setups.

## Checklist

### DIFF
--- a/panda/src/audiotraits/openalAudioSound.cxx
+++ b/panda/src/audiotraits/openalAudioSound.cxx
@@ -17,6 +17,9 @@
 #include "openalAudioSound.h"
 #include "openalAudioManager.h"
 
+
+#include "coordinateSystem.h"
+
 TypeHandle OpenALAudioSound::_type_handle;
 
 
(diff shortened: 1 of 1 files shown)


rdb:
Would you be interested in having it handle the other coordinate system cases as well?

Tanay-ErrorCode:
@rdb PTAL. 
Added coordinate system detection to the OpenAL audio code. When using left-handed coordinate systems (like y-up-left), we now negate the X-axis when positioning sounds and the listener, which fixes 3D audio positioning that was previously incorrect.

Coordinate systems now supported:

CS_yup_right (Y-up, right-handed) - Default Panda3D
CS_yup_left (Y-up, left-handed) - Fixed with X-axis negation
CS_zup_right (Z-up, right-handed) - Direct OpenAL mapping
CS_zup_left (Z-up, left-handed) - Fixed with X-axis negation

rdb:
Continued by #1786

Issue #1535, opened 2023-09-18
CI needed for static build


rdb:
It's come to my attention that the static build of Panda is currently broken.

We need a CI job that builds and tests a static build of Panda on different platforms.

Would be grateful if someone is willing to work on this.

JuliaLWang8:
I'd like to work on this issue if no one else is currently on it!

rdb:
Feel free!

Oculus049:
Looks like this hasn't been touched in a while. I haven't done much work with devops before but since it's marked with good first issue so I presume won't be too crazy difficult so I'll give a shot. To be done with github actions yeah? Haven't done before but seems to be plenty of guide content on it.

https://github.blog/enterprise-software/ci-cd/build-ci-cd-pipeline-github-actions-four-steps/

Oculus049:
https://github.com/Oculus049/panda3d/blob/master/.github/workflows/main.yml
I have this as a base but am iffy on how to actually make the static build.

Oculus049:
Sorry I realized I wasn't properly following the contribution guidelines. (My first time doing any proper open source work) https://github.com/panda3d/panda3d/pull/1755

Issue #1080, opened 2020-12-21
Static linking results in a missing panda3d.core


Clockwork-Muse:
## Description
I'm trying to create a minimal panda build for inside a docker container working (I have it working in general, just seeing what I can remove/tweak), and when specifying `makepanda/makepanda.py --static <all other options>` I can no longer run projects from python:

```
$ python headless.py
Traceback (most recent call last):
  File "headless.py", line 1, in <module>
    from panda3d.core import *
ModuleNotFoundError: No module named 'panda3d.core'
```

Note that my `<build_directory>/lib` contains a `libpy.panda3d.core.cpython-38-x86_64-linux-gnu.a` file, which I am assuming is the module it's looking for.

Note that leaving `--static` off (using dynamic linking) results in a working build, at least for the features I'm using so far.

## Steps to Reproduce
I'm building python with the following command:
```
 makepanda/makepanda.py --nothing --wheel --optimize 4 --threads 6 --static --use-python --use-direct --use-gl --use-egl --no-x11 --use-eigen  --use-pandaparticlesystem --use-pandaphysics --use-zlib --use-png --use-jpeg --use-openexr --use-squish --use-gtk2 --use-pandatool --use-pview --use-deploytools --use-sse2 --use-egg --verbose --outputdir /some_build_directory
 ```
[Build Output](https://github.com/panda3d/panda3d/files/5726205/build_output.txt)

## Environment
 * Operating system: Ubuntu 20.04 docker container
 * System architecture: 
 * Panda3D version: 1.11, commit 54b93116e8910dc19a68780d6325ea5e8f8636f9
 * Installation method: pip-installed built wheel into venv
 * Python version (if using Python): 3.8

rdb:
What are you expecting it to generate, if not a static library?

Clockwork-Muse:
... I guess I was expecting it to still be usable from python, given it was generating a wheel (since there are ways to use static libraries from python). 

Otherwise, "not supported" is an acceptable answer to me.

rdb:
The reason I asked what you were expecting is that I think there are two kinds of people who might want to make a static build of Panda with Python:
1. People who want to produce a fully self-contained executable, probably including frozen bytecode (as produced by pfreeze), that entirely contains all the Python bindings and C++ code.  These users will expect static builds of the Python modules, as makepanda is producing.
2. People who still want to get "normal" Python extension modules like panda3d/core.so, but simply expect all the C++ code to be linked statically into it.

[...]

PR #1779, opened 2025-11-02
Added static build configurations to ci.yml

rawmeat21:
## Issue description

Hello, I saw an open issue <a href="https://github.com/panda3d/panda3d/issues/1535">#1535</a> for adding CI static build configurations for different platforms. I have added 3 jobs for testing static builds on Ubuntu, Windows and macOS based platforms to .github/workflows/ci.yml.

While testing the jobs on Github Actions, I noticed the following error:

`ImportError while loading conftest '/Users/runner/work/panda3d/panda3d/tests/conftest.py'.
../tests/conftest.py:3: in <module>
    from panda3d import core
E   ImportError: cannot import name 'core' from 'panda3d' (/Users/runner/work/panda3d/panda3d/build/panda3d/__init__.py)`

## Checklist

I have done my best to ensure that:

* [ ] I have familiarized myself with the CONTRIBUTING.md file
* [ ] this change follows the coding style and design patterns of the codebase
* [ ] I own the intellectual property rights to this code
* [ ] the intent of this change is clearly explained
* [ ] existing uses of the Panda3D API are not broken
* [ ] The changed code is adequately covered by the test suite, where possible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,10 @@ jobs:
         - macos-nometa-standard-makefile
         - windows-standard-unity-msvc
         - windows-nopython-nometa-standard-msvc
-
+        # static build profiles
+        - ubuntu-static-standard-makefile
+        - windows-static-standard-msvc
+        - macos-static-standard-makefile
         include:
         - profile: ubuntu-bionic-double-standard-unity-makefile
           os: ubuntu-22.04
@@ -29,7 +32,8 @@ jobs:
           python: YES
           eigen: NO
           double: YES
-
+          link_type: SHARED
+          
         - profile: ubuntu-bionic-coverage-ninja
           os: ubuntu-22.04
           config: Coverage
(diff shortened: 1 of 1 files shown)


Wizzerinus:
That import issue actually somewhat makes sense, see also: https://github.com/panda3d/panda3d/issues/1080

rawmeat21:
@Wizzerinus Thanks, I read the @rdb 's comment under the issue. I'd like to work on the panda3d/_all.so library. I'm quite new to open source and would like to make my first contribution. What tools should I look into so that I can do this?

rdb:
For testing I don't think we need to go with the `_all.so` approach (which I think is more complicated to do, perhaps too much so for a first contribution).

Rather we just need to compile a runner program that interfaces with the Python C API and links in all the statically built Python modules. Fortunately I've already made such a script, it's `tests/main.c`, which builds a `run_tests` executable, currently used by the emscripten test suite, which is always built statically. This is currently only built with makepanda so you would have to adapt the CMake scripts to build it as well when Panda is built statically.

PR #1781, opened 2025-11-23
Allow Windows to search and load Icons and Cursors from VFS.

RegDogg:
## Issue description
This change will allow Windows to search Multifiles mounted with VFS for icons and cursors

## Solution description
By utilizing functions `LookupIconIdFromDirectoryEx` and `CreateIconFromResourceEx` we can read filedata in the VFS from `icon-filename/cursor-filename` filepaths

## Checklist


panda/src/windisplay/winGraphicsWindow.cxx
```diff
@@ -3072,39 +3074,54 @@ get_icon(const Filename &filename) {
...
   // If it wasn't found, resolve the filename and search for that.
-
-  // Since we have to use a Windows call to load the image from a filename, we
-  // can't load a virtual file and we can't use the virtual file system.
   Filename resolved = filename;
-  if (!resolved.resolve_filename(get_model_path())) {
-    // The filename doesn't exist along the search path.
-    if (resolved.is_fully_qualified() && resolved.exists()) {
-      // But it does exist locally, so accept it.
+  if ((resolved.resolve_filename(get_model_path())) || (resolved.is_fully_qualified() && resolved.exists())) {
```
rdb: It may still be in the VFS if `exists()` returns true if there is an overlapping mount.

Here's how to get the system filename from a VFS filename:
```python
      PT(VirtualFile) vfile = ...;
      // check vfile
      SubfileInfo info;
      if (vfile->get_system_info(info) && info.get_start() == 0) {
        Filename fn = info.get_filename();
        // do something with fn.to_os_specific()
```

panda/src/windisplay/winGraphicsWindow.cxx
```diff
@@ -3072,39 +3074,54 @@ get_icon(const Filename &filename) {
...
+  }
+  VirtualFileSystem *vfs = VirtualFileSystem::get_global_ptr();
+  PT(VirtualFile) vfile = vfs->get_file(filename);
+  if (vfile != nullptr) {
+    std::string vfile_data = vfile->read_file(0);
+    long fsize = vfile_data.length();
+
+    int offset = LookupIconIdFromDirectoryEx((PBYTE)vfile_data.data(), TRUE, 0, 0, LR_DEFAULTCOLOR);
+    if (offset != 0) {
+      HICON hIcon = CreateIconFromResourceEx((PBYTE)vfile_data.data() + offset, fsize - offset, TRUE, 0x30000, 0, 0, LR_DEFAULTCOLOR);
```
rdb: I'm wondering if there is a lifetime issue here. I haven't read the documentation, but does the icon created by CreateIconFromResourceEx copy the data passed in, or does it continue to refer to the same data in memory? If so, that might be an issue since the `std::string` will be deallocated at the end of the function.

Secondly, you're passing in the remaining size of the whole data structure, not the size of the icon in it; is that okay?

rdb (commented):
This sounds like a useful feature to have, thanks for working on it! Comments inline.

PR #1800, opened 2026-01-12
makepanda: fix for python >= 3.13 on android

thunder-coding:
android is now a supported platform Python since Python 3.13

Found initially during Python bump for Termux (https://github.com/termux/termux-packages/pull/27739)

Also seems like Pyright too hasn't caught up with this yet. So it needs
to be made aware as well

## Checklist


rdb (approved):
Thanks!

Issue #1815, opened 2026-04-04
AudioSound: A Copy Method


ArsThaumaturgis:
## Description
In short, I'm requesting that a method be added to AudioSound that allows one to make a copy of the AudioSound in question.

(i.e. Without loading the sound by filename.)

Much as a model can be copied via NodePath's "copyTo" method.

## Use Case
The main use-case that I have to hand is that of a class that accepts a sound-object, and that is intended to be able to copy itself--and thus which calls for copying the sound-object.

Now, the same sound-object could be shared between copies of the class that holds them... but this wouldn't allow the copies to apply different settings to their sound-objects (e.g. play-rate, volume, 3D positioning, etc.).

gossfunkel:
Hiya! I'm having a look at the source for the AudioSound classes, and it seems that yes, this should be quite achievable! I'm working on a `make_copy` function as we see on many classes such as Actor, Geoms, Lenses etc, as the `copy_to` on NodePath is more of a hybrid of `make_copy` and `reparent_to` (thanks to @rdb for this tip and other guidance to help me put this together). There's some detailed discussion on the Discord if you want to keep up with the nitty-gritty, but I've taken your wording as the brief to work to, and will implement this in-fitting with Panda3D style standards! I may add the same functionality for FMOD if I can, which might add a little longer to the time before I can get that in a PR, but I'd prefer to avoid having a temporary error thrown for FMOD users to get a quicker update out, since this seems like a small enough change to just implement cross-library support all at once. I'll try to update you if there's anything I need to know for what you'd need in your use-case, but so far it looks fairly straightforward.

ArsThaumaturgis:
Thank you very much, gossfunkel! And please let me know if I can help with the efforts in some way! ^_^

>I'm working on a make_copy function as we see on many classes such as Actor, Geoms, Lenses etc, as the copy_to on NodePath is more of a hybrid of make_copy and reparent_to ...

That is fair!

Although I'd suggest that having a similarly-hybrid approach be available for audio-copying wouldn't be entirely unfounded, as it would allow the resultant AudioSound to be associated with the AudioManager of the original AudioSound. Which would be convenient in at least some cases!

rdb:
I think that should be the default; I'm not sure I see a good use case for copying sounds to other managers? That won't always be possible - like copying an OpenAL sound to an FMOD manager

gossfunkel:
[...]

PR #1822, opened 2026-04-08
Feature request in issue #1815 - copying an OpenALAudioSound

gossfunkel:
## Issue description
In issue #1815 there was a request for a simple method to be added to the OpenALAudioSound class that allows a copy to be made of an AudioSound object at runtime,  when using OpenAL.  

## Solution description
I have implemented a basic copy constructor for the OpenALAudioSound class.  It performs a soft copy  (reference pointers are copied;  the objects pointed to are not).  I have also implemented a virtual `make_copy()` method in the parent AudioSound class,  and implemented it in the OpenALAudioSound class and the NullAudioSound class.  The OpenALAudioSound implementation constructs a new OpenALAudioSound on the heap with the copy constructor,  using the object the method was called on,  and returns a pointer to it.  The NullAudioSound implementation simply returns a pointer to a new NullAudioSound constructed on the heap.  

I have included tests for both implementations,  and I have added a new `wav_test.wav` sound file to `tests/audio` for these and future tests  (it is two samples long; the first sample has an amplitude value of 0 and the second a value of 1, and the sample rate is 48kHz).

## Checklist


.gitignore
```diff
@@ -64,6 +64,9 @@ ehthumbs.db
 __pycache__/
 *.pyc
 *.pyo
+python-1.11.0*
+pyproject.toml
+uv.lock
```
rdb: I would prefer no unrelated changes in this commit and pyproject.toml is in fact something we might want to add in the future :-)

tests/audio/test_openalAudioSound.py
```diff
@@ -0,0 +1,29 @@
...
+
+    test_sound.setActive(1)
+    test_sound.set3dMaxDistance(20.0)
+    test_copy = test_sound.make_copy()
+
+    assert(test_copy.getActive() == test_sound.getActive())
+    assert(test_copy.getName() == test_sound.getName())
+    assert(test_copy.get3dMaxDistance() == test_sound.get3dMaxDistance())
+    # check that make_copy has copied manually-modified parameters
+    assert(test_copy.get3dMaxDistance() == 20.0)
```
rdb: Nitpicks: `assert expr` instead of `assert(expr)` and we prefer `snake_case` conventions in the unit test suite

panda/src/audiotraits/openalAudioSound.cxx
```diff
@@ -123,6 +182,22 @@ cleanup() {
   _manager = nullptr;
 }
 
+/**
+ * Copies an OpenALAudioSound into a new OpenALAudioSound.
+ */
+AudioSound *OpenALAudioSound::
+make_copy() const {
+  ReMutexHolder holder(OpenALAudioManager::_lock);
```
rdb: The copy ctor already holds the lock, so we don't need to hold it here - it doesn't do anything in addition that needs to be covered by the lock

panda/src/audiotraits/openalAudioSound.cxx
```diff
@@ -94,6 +94,65 @@ OpenALAudioSound(OpenALAudioManager *manager,
...
+
+  if (copy_sound.is_positional()) {
+    if (_sd->_channels != 1) {
+      audio_warning("copied stereo sound " << _movie->get_filename() << " will not be spatialized");
+    }
+  }
+
+  _comment = std::move(_sd->_comment);
+  release_sound_data(false);
+}
```
rdb, line 154: Copy ctor looks good! :-) The warning probably isn't necessary (since it was already issued for the original sound but doesn't really do any harm either.

panda/src/audio/audioSound.cxx
```diff
@@ -12,6 +12,7 @@
  * Prior system by: cary
  */
 
+#include "throw_event.h"
```
rdb: Was this necessary for something? We're not actually using `throw_event` in this file

tests/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ if(NOT INTERROGATE_PYTHON_INTERFACE)
   return()
 endif()
 
+cmake_minimum_required(VERSION 3.2...4.3)
```
rdb: Can you explain the need for this line?

rdb (changes requested):
Thanks a lot for working on this! Looks good generally but I have left some comments inline.

tests/CMakeLists.txt
gossfunkel: I included this line to get CI to complete successfully- when trying to build the test suite first time, CMake threw an error due to lacking this line. This is because, from CMake 4.0, [it now throws an error rather than a deprecation warning](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) when this line is not included. I tried removing it, but found that some of the CI builds failed when I submitted the PR, and when I committed the line in with this formatting, all the builds went green!

I understand it's not related to this particular PR- I just wanted to submit this in the best working order I could, in line with the practice recommended by the build system.

.gitignore
gossfunkel: No problem, I'll take these out for now.

panda/src/audio/audioSound.cxx
gossfunkel: Oops, that's my mistake, sorry about that. Removed in the next commit.

rdb (approved):
Great! Thanks so much!

PR #1853, opened 2026-06-01
Implement EBR to solve scenegraph threading.

Maxwell175:
## Issue description

Panda3D's threaded `PipelineCycler` has not been reliably thread-safe, which has
come to the forefront as we see more multithreading efforts from Python No-GIL 
and possible C# bindings.

Two design facts come together to cause the issue:

- `CycleDataReader` is lock-free by design - it raw-loads the published
  `CycleData` pointer with no lock, no ref-count, and no atomic.
- A writer may mutate the published `CycleData` in place - `write_stage()`
  only copied when `get_node_ref_count() != 1`, otherwise it mutated the very
  object a reader might already be holding by raw pointer.

## Solution description

Replace the cycler's "lock-free reader + in-place writer" model with a combination of
lock-free readers, copy-on-write writers, and epoch-based reclamation (EBR). A reader
loads the published pointer (no shared write) and dereferences it for the
lifetime of an EBR critical section; a writer copies, mutates the copy, then
atomically publishes it and "retires" the old pointer, which is freed only once
no thread can still be observing it. The pointer a reader holds is therefore
never mutated under it.

Performance regression was a legitimate concern, especially since single-threaded 
load is the only real supported existing use case. I went down a few paths to 
really smooth out the performance penalty.

- A **per-stage in-place fast path**: a write still mutates in place when
  provably safe (the writer is the sole occupant of its own, unshared stage),
  gated by a stage-occupancy interlock. This recovers single-threaded write speed.
- Amortizing EBR critical-section entry to **once per frame/task** via framework
  hooks (`GraphicsEngine::render_frame`, `AsyncTaskManager::poll`, each
  `AsyncTaskChain` task), so inner `CycleDataReader`s cost only a thread-local
  depth bump — no atomic per node read.
- The EBR participant living on the `Thread` for bound threads (a plain member
  load on the cull hot path, not a slow shared-library `thread_local` lookup).

I added a `pipeline-always-cow` config var that bypasses the per-stage fast path
optimization to force COW everywhere. It was useful for benchmarking COW itself and
it could be removed if needed.

### What's in each commit

| Commit | What                                                                                                                                                |
|---|-----------------------------------------------------------------------------------------------------------------------------------------------------|
| `dtool: make dlmalloc mmap-only…` | dlmalloc shared `sbrk`/the program break with glibc's allocator; concurrent `sbrk()` corrupted the heap. Force `HAVE_MORECORE 0` (mmap-only)        |
| `putil: fix CopyOnWritePointer lock lifetime…` | Pin the COW object in a strong ref so a holder can't unlock freed memory; defer EBR reclaim past the locked region to avoid a self-deadlock.        |
| `pipeline: thread-safe PipelineCycler via EBR` | The core change: `EpochManager`, lock-free readers, COW writers, per-stage in-place fast path, stage-occupancy interlock acquired at thread start.  |
| `pipeline: drive EBR epoch frames…` | Framework hooks that open/close one critical section per frame/task.                                                                                |
| `pgraph: make PandaNode concurrency-safe under EBR` | `update_cached`/cull read-path safety, a lock-order-inversion fix, dead `node_ref_count` removal, and the `~PandaNode` parent-walk crash fix.       |
| `gobj: honor cycler write lock in multi-writer GeomVertexData paths` | Reuse an in-flight writer instead of starting a second COW on a stage this thread already writes (fixes an issue when writing data like in a Card). |
| `tests: scene-graph stress, EBR correctness, and cycler benchmarks` | Various test tools: `scenegraph_stress`, `ebr_cycler`, `cycler_bench`, `scenegraph_bench`.                                                          |
| `ci: run scene-graph stress and EBR tests on all platforms` | `ctest` step across the Ubuntu/macOS/Windows matrix (Ubuntu under ASan).                                                                            |

### Notes

- `DO_PIPELINING=OFF` / `THREADED_PIPELINE=OFF` builds are unaffected — the EBR
  API stubs to no-ops there.
- dlmalloc mmap-only forfeits the `sbrk` fast path and rounds small-heap growth
  to mmap granularity maybe there is a better solution, but only applies in configs 
  that actually use dlmalloc.

## Checklist


panda/src/putil/copyOnWritePointer.cxx
```diff
@@ -60,36 +110,46 @@ get_read_pointer(Thread *current_thread) const {
  */
 PT(CopyOnWriteObject) CopyOnWritePointer::
 get_write_pointer() {
-  if (_cow_object == nullptr) {
+  // Pin _cow_object before locking — see get_read_pointer().  Without the pin,
+  // another thread can swap and free the old object before our explicit
+  // unlock, leaving us unlocking freed memory.
+  PT(CopyOnWriteObject) old_object = _cow_object;
```
rdb: If this is indeed a risk as you say, the pointer swap must be atomic - but you can't grab the pointer and incref in one atomic operation.

But we already have the right tool to deal with this risk properly: EBR, so I don't quite understand this.

panda/src/putil/copyOnWritePointer.cxx
```diff
@@ -14,6 +14,51 @@
...
+   * (which cache_unref()s its sub-objects) can re-acquire that very _lock_mutex
+   * on this same thread, a self-deadlock.
+   *
+   * Keeping an outer critical section open for the whole COW makes the inner
+   * leave non-outermost, so reclamation is deferred until this guard is
+   * destroyed at function exit -- after every _lock_mutex has been released.
+   * (Deferring reclamation is always safe under EBR.)  No-op when the pipeline
+   * is not threaded, where there is no EBR and thus no such reclamation.
+   */
+  class CowReclaimDeferral {
```
rdb: This is a good RAII wrapper to have, and it should be used more generally in this PR, but the name and stated use case is confusingly specific.

dtool/src/dtoolbase/memoryHook.cxx
```diff
@@ -78,6 +78,13 @@ static_assert((MEMORY_HOOK_ALIGNMENT & (MEMORY_HOOK_ALIGNMENT - 1)) == 0,
 // dlmalloc can do the alignment we ask for.
 #define MALLOC_ALIGNMENT MEMORY_HOOK_ALIGNMENT
 
+// Use mmap exclusively: dlmalloc's default MORECORE is sbrk, which moves the
+// same program break as the system (glibc) allocator.  The two allocators'
+// locks are independent, so concurrent sbrk() from a dlmalloc thread and a
+// glibc thread (e.g. libstdc++ operator new) corrupts the heap ("malloc():
+// corrupted top size").  mmap never touches the program break.
+#define HAVE_MORECORE 0
```
rdb, line 86: I did not know about this - we should probably be considering using mimalloc more aggressively in any case

panda/src/display/graphicsEngine.cxx
```diff
@@ -2874,8 +2882,14 @@ thread_main() {
       break;
 
     case TS_do_frame:
+#ifdef THREADED_PIPELINE
+      current_thread->epoch_enter();
```
rdb: 1. The ifdefs aren't necessary, epoch_enter are already noops, they're just noise
2. This is where a RAII block (which you already have, see below) would be useful

panda/src/pgraph/pandaNode.cxx
```diff
@@ -114,7 +114,16 @@ PandaNode::
...
+  // Use do_remove_all_children, not the public remove_all_children: the latter
+  // also runs force_bounds_stale + children_changed + mark_bam_modified, which
+  // the destructor must not.  force_bounds_stale walks self->_up, whose
+  // UpConnections hold raw, un-refcounted back-pointers to parents (kept raw to
+  // avoid parent<->child cycles); under concurrent EBR a parent may be
+  // mid-teardown, so the walk could read freed memory in
+  // CDStageReader<PandaNode::CData>::ctor.  The notifications are also moot
+  // here: any parent that held us already dropped its Down entry (else our
+  // refcount could not be 0) and staled its bounds then.
+  do_remove_all_children(Thread::get_current_thread());
```
rdb, line 117: The change is fine, but comments like these are unnecessarily wordy (having to parse a block of English slows down code comprehension dramatically) and hard to maintain. It outlines a very specific corner case whereas it is obviously correct to skip force_bounds_stale etc. in the destructor on a much more general level.

panda/src/pgraph/pandaNode.cxx
```diff
@@ -3270,25 +3310,36 @@ update_cached(bool update_bounds, int pipeline_stage, PandaNode::CDLockedStageRe
...
+    pending_cache.resize(num_children);
 
     // We need to keep references to the bounding volumes, since in a threaded
     // environment the pointers might go away while we're working (since we're
     // not holding a lock on our set of children right now).  But we also need
     // the regular pointers, to pass to BoundingVolume::around().
+    // Heap-backed, not alloca: the do/while above retries under concurrent
+    // mutation and alloca isn't freed until the function returns, so repeated
+    // allocas would overflow the stack; a pvector frees each iteration.
+    pvector<const BoundingVolume *> child_volumes_storage;
```
rdb, line 3327: fair point

panda/src/pgraph/pandaNode.cxx
```diff
@@ -3238,7 +3269,16 @@ update_cached(bool update_bounds, int pipeline_stage, PandaNode::CDLockedStageRe
   }
   Thread *current_thread = cdata.get_current_thread();
 
+  // Under sustained concurrent mutation the optimistic snapshot-then-commit
+  // below can fail its staleness check every iteration and livelock.  We
+  // prioritise the cull thread never hitching over bound freshness: accept the
+  // consistent snapshot computed on the first contended commit (see below)
+  // rather than retrying the expensive recursive recompute.  Hence 1.
+  const int max_update_attempts = 1;
+  int attempt = 0;
```
rdb: Isn't this functionally the same as taking out the whole loop?

It's becoming harder and harder to reason through the logic in this function, unfortunately, would be good to simplify/refactor this a bit.

In principle I think it's fine that cull accepts any change so long as it was made in the current frame

panda/src/pgraph/pandaNode.h
```diff
@@ -405,6 +405,12 @@ class EXPCL_PANDA_PGRAPH PandaNode : public TypedWritableReferenceCount,
   class CData;
 
   INLINE int do_find_parent(PandaNode *node, const CData *cdata) const;
+  // Shared body of remove_all_children: severs every child connection but
+  // does NOT call force_bounds_stale / children_changed / mark_bam_modified.
+  // Used directly by ~PandaNode where the trailing notifications would walk
+  // self->_up via raw back-pointers into potentially-dead parents (see the
+  // comment on the destructor for the race details).
+  void do_remove_all_children(Thread *current_thread);
```
rdb, line 408: doesn't need a comment, especially not in the header, it's self-explanatory

panda/src/pgraph/pandaNode.h
```diff
@@ -579,9 +585,11 @@ class EXPCL_PANDA_PGRAPH PandaNode : public TypedWritableReferenceCount,
     // This section contains the lightweight parts of the node that are likely
     // to change fairly often: transform and state.
 
-    NCPT(RenderState) _state;
-    NCPT(TransformState) _transform;
-    NCPT(TransformState) _prev_transform;
+    // CPT, not NCPT: nothing reads these states' node_ref_count (cache GC uses
+    // cache_ref_count), so NCPT's extra atomic per copy/destroy was dead weight.
```
rdb: Pet peeve: this is an annoying LLM quirk that they will make comments in the code explaining why a change was made; comments should describe the current state of the code, someone reading the current code doesn't have to concern themselves with what the code used to do.

panda/src/pgraph/pandaNode.I
```diff
@@ -1181,8 +1181,13 @@ PandaNodePipelineReader(const PandaNode *node, Thread *current_thread) :
 #ifdef DO_PIPELINING
   // We node_ref the CData pointer, so that if anyone makes changes to the
   // PandaNode while we hold this pointer, it will force a copy--so that this
-  // object will remain unchanged (if out-of-date).
-  _cdata->node_ref();
+  // object will remain unchanged (if out-of-date).  The null check tolerates a
+  // null _cdata, which a cull reader can rarely observe under heavy concurrent
+  // structural mutation; compare_draw_mask then skips the node instead of
+  // dereferencing null.
```
rdb: This raises alarm bells for me; this is a thread-local object and other threads can't concurrently modify its `_cdata` pointer (and if they could, this would be woefully insufficient to address it).

The only places that can set it to null are the destructor (so you may have observed use-after-free) and the outer `update_cached()`. You should NEVER see concurrent access to the same PandaNodePipelineReader object.

panda/src/pipeline/cycleDataLockedReader.I
```diff
@@ -27,6 +27,7 @@ CycleDataLockedReader(const PipelineCycler<CycleDataType> &cycler,
   _cycler(&cycler),
   _current_thread(current_thread)
 {
+  Thread::epoch_enter(_current_thread);
```
rdb: A RAII member would earn its keep here, but I have thoughts about entering on every CDReader/CDWriter

panda/src/pipeline/epochManager.cxx
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// The shared ExternalThread's per-OS-thread participant.  Function-local
+// (not an exported global) because MSVC forbids a dll interface on a
+// thread_local; bound threads carry their own participant on the Thread.
+// Default-constructed lazily; its destructor runs at thread exit and
+// unregisters it.
+EpochParticipant &EpochManager::
+external_participant() {
+  static thread_local EpochParticipant tl;
```
rdb: Should probably just live on Thread; we generally already have a Thread pointer handy that is probably already hot most of the time

panda/src/pipeline/epochManager.cxx
```diff
@@ -0,0 +1,248 @@
...
+patomic<int> EpochManager::_active_cs_count{0};
+patomic<size_t> EpochManager::_retired_count{0};
+
+// Never-destroyed participant registry (see epochManager.h).  Heap-allocated
+// once and intentionally leaked, so a thread_local participant's destructor
+// can unregister safely at any point -- including during static teardown at
+// program exit, when an ordinary file-scope std::mutex might already be gone.
+namespace {
+  struct EbrRegistry {
+    std::mutex lock;
```
rdb: If we use std::mutex and std::lock_guard that should be an across-the-board migration, for now, better use Panda classes (which come with a bunch of useful debug/pstats features and deadlock detection with DEBUG_THREADS on)

panda/src/pipeline/epochManager.h
```diff
@@ -0,0 +1,276 @@
...
+// patomic members; keep it opaque.
+struct EpochParticipant {};
+#endif  // CPPPARSER
+
+class EXPCL_PANDA_PIPELINE EpochManager {
+public:
+  // Enter/leave a critical section.  The participant is passed in (rather than
+  // looked up) to keep the hot path off the slow shared-library thread_local
+  // access for bound threads; see Thread::epoch_participant.  Depth-counted --
+  // only the outermost transition touches the atomic slot.
```
rdb: The comments belong in the .cxx file in Panda style

panda/src/pipeline/epochManager.h
```diff
@@ -0,0 +1,276 @@
...
+  // mid-write: try_begin_inplace_write claims a slot and re-checks the count;
+  // stage_occupancy_inc, on raising a stage to 2+, spin-drains in-flight
+  // in-place writers first.  The count is a live gate, not sticky.
+  ALWAYS_INLINE static bool try_begin_inplace_write(int stage);
+  ALWAYS_INLINE static void end_inplace_write(int stage);
+  ALWAYS_INLINE static bool inplace_allowed_hint(int stage);
+
+  // Pipelines rarely exceed 3 stages; a writer at a stage >= this bound never
+  // takes the in-place path (conservative).
+  static constexpr int MAX_STAGES = 16;
```
rdb: I'm fine with this (probably belongs on Pipeline as a constant), but 4 is enough

panda/src/pipeline/epochManager.h
```diff
@@ -0,0 +1,276 @@
...
+    // order a store ahead of a later load to a different location, so without
+    // this fence (even on x86, whose store buffer reorders store->load) the
+    // load could hoist above the slot store, letting a reclaimer see us
+    // quiescent and free a pointer we are about to dereference.  Pairs with the
+    // fence in observed_min_epoch_locked().  Outermost enter only (~per frame).
+    // StoreLoad fence is provided for free by the seq_cst RMW below: on x86
+    // LOCK XADD drains the store buffer and is a full barrier; on ARM,
+    // seq_cst adds a DMB ISH after STLXR.  This subsumes the previous
+    // patomic_thread_fence(seq_cst) which was redundant on x86.
+    _active_cs_count.fetch_add(1, std::memory_order_seq_cst);
```
rdb: This is probably my biggest concern with the current approach; in a model where we're doing an epoch entry/exit on every reader/writer, a seq-cst on a hotly contested shared variable is a significant pessimization. I also need some convincing that this mechanism is the right approach, because I can't imagine this is guaranteed to ever reach 0.

panda/src/pipeline/pipeline.cxx
```diff
@@ -154,7 +157,7 @@ cycle() {
           // We save the result of cycle(), so that we can defer the side-
           // effects that might occur when CycleDatas destruct, at least until
           // the end of this loop.
-          saved_cdatas.push_back(cycler->cycle_2());
+          cycler->cycle_2();
```
rdb, line 163: This is one of the areas where the system really pays off! But we need to be inside an active EBR block for this to be safe. So needs an assert or an enter/exit around the method.

panda/src/pipeline/pipeline.cxx
```diff
@@ -262,11 +265,10 @@ cycle() {
...
 
-  // And now it's safe to let the CycleData pointers in saved_cdatas destruct,
-  // which may cause cascading deletes, and which will in turn cause
-  // PipelineCyclers to remove themselves from (or add themselves to) the
-  // _dirty list.
-  saved_cdatas.clear();
+  // Drain a bounded chunk of the EBR retire queue (bounded so heavy churn
+  // can't make cycle() pay an unbounded reclamation cost; the rest waits for
+  // later cycles).
+  EpochManager::try_reclaim(256);
```
rdb: This is okay (the placement of this is pretty much arbitrary), but someone might be misled to think that this will free up the stuff we pushed out in `cycle()`. It could never, because we're inside an EBR block.

There *is* a major issue here that you never do an unbounded `try_reclaim()` call, so on large applications the reclaim list will currently grow indefinitely.

panda/src/pipeline/pipelineCyclerTrueImpl.cxx
```diff
@@ -277,34 +268,44 @@ write_stage_upstream(int pipeline_stage, bool force_to_0, Thread *current_thread
  * effects that cascade through the system, so the caller may choose to hold
  * the pointer until it can safely be released later.
  */
-PT(CycleData) PipelineCyclerTrueImpl::
+void PipelineCyclerTrueImpl::
 cycle() {
```
rdb, line 385: Need to assert we're inside an EBR block.

panda/src/pipeline/thread.cxx
```diff
@@ -108,6 +115,50 @@ bind_thread(const std::string &name, const std::string &sync_name) {
   return thread;
 }
 
+#ifdef THREADED_PIPELINE
+/**
+ * Explicit EBR reclamation point for long-running user code: drains fully
+ * (unlike the throttled hook in the yield family).  Call only with no
+ * CycleDataReader outstanding on this thread.
```
rdb, line 122: We can trivially verify this, right, since CDReader enters?

panda/src/pipeline/thread.I
```diff
@@ -362,3 +373,32 @@ operator << (std::ostream &out, const Thread &thread) {
   thread.output(out);
   return out;
 }
+
+#ifdef THREADED_PIPELINE
+
+ALWAYS_INLINE EpochParticipant &Thread::
+epoch_participant() {
+  return _epoch_use_tls ? EpochManager::external_participant() : _epoch_participant;
```
rdb: Oh, I think I see now - this is a workaround for the fact that there is only one ExternalThread pointer, so otherwise you can't guarantee thread safety among multiple external threads.

I'm not 100% sure about this, we should either ensure that there's a unique Thread pointer registered per thread (in your use case it can be handled in the bindings) or we should use a different TLS block in EpochManager in general. I'm not sure this optimization buys us much. I'm leaning towards the former.

panda/src/gobj/geomVertexData.cxx
```diff
@@ -2537,6 +2537,11 @@ modify_array(size_t i) {
   PT(GeomVertexArrayData) new_data;
   if (_got_array_writers) {
     new_data = _array_writers[i]->get_object();
+  } else if (_cdata->_arrays[i].get_unsafe_pointer()->is_write_in_progress_by(_current_thread)) {
+    // Already being written by this thread; reuse rather than copy-on-write so
+    // the caller's handle nests on the same in-flight generation (see
+    // make_array_writers()).
+    new_data = _cdata->_arrays[i].get_unsafe_pointer();
```
rdb, line 2544: It's not immediately obvious to me why this lives here and is not generalized in get_write_pointer.

panda/src/pipeline/epochManager.cxx
```diff
@@ -0,0 +1,248 @@
...
+  std::lock_guard<std::mutex> hold(r->lock);
+  EpochParticipant **pp = &r->head;
+  while (*pp != nullptr && *pp != p) {
+    pp = &(*pp)->next;
+  }
+  if (*pp == p) {
+    *pp = p->next;
+  }
+  p->next = nullptr;
+  p->registered = false;
```
rdb, line 107: This is fine I guess, registration is pretty rare anyway

panda/src/pipeline/epochManager.h
```diff
@@ -0,0 +1,276 @@
...
+  // Participant registry head/lock live behind a never-destroyed singleton
+  // (see .cxx) so a thread_local participant dtor can unregister safely even
+  // during static teardown.
+#endif  // CPPPARSER
+};
+
+#ifndef CPPPARSER
+ALWAYS_INLINE void EpochManager::
+enter(EpochParticipant &p) {
+  if (p.depth++ == 0) {
```
rdb: This is the only line that should be inlined - everything else needs to explicitly NOT be inlined, otherwise it will actively hurt code size and performance.

panda/src/pipeline/epochManager.cxx
```diff
@@ -0,0 +1,248 @@
...
+      if (read->epoch < min_observed && to_free.size() < budget) {
+        to_free.push_back(read->cd);
+      } else {
+        if (write != read) {
+          *write = *read;
+        }
+        ++write;
+      }
+    }
+    _retired.erase(write, _retired.end());
```
rdb: I tried for a while to reason through why this resorting approach is necessary, why not use a pdeque and erase from the beginning... but it doesn't really hurt much in any case, so probably fine

tests/scenegraph_bench/scenegraph_bench.cxx
```diff
@@ -0,0 +1,343 @@
...
+#include <chrono>
+#include <cstdio>
+#include <cstring>
+#include <deque>
+#include <random>
+#include <string>
+#include <thread>
+#include <vector>
+
+using Clock = std::chrono::steady_clock;
```
rdb: Not using Panda clock but for a test I'll accept it

panda/src/putil/copyOnWritePointer.cxx
```diff
@@ -14,6 +14,51 @@
 #include "copyOnWritePointer.h"
 #include "config_putil.h"
 #include "config_pipeline.h"
+#include "thread.h"
+
+#ifdef COW_THREADED
+namespace {
```
rdb: This seems like a style departure; we don't do this elsewhere in the Panda code base.

rdb (commented):
Cool! I didn't realise you were working on this, I have been mucking around a bit with QSBR in the meantime so all the considerations are fairly fresh in my head; I'm putting some comments in here for now but haven't reasoned through everything yet. I'm curious why you went with EBR in the end? You're provided a means to identify quiescent states and entering/exiting around tasks, but then you're *also* entering/exiting around every CDReader/CDWriter, so it feels a bit like you couldn't quite decide and implemented both at the same time, which is a bit redundant. The enter/exit for every CDReader/CDWriter has a real cost, one which feels unnecessary; you're guarding it with a depth check at least, which brings the cost down significantly, but now we have a performance pitfall.

I'm also not convinced that the reclaim list doesn't grow indefinitely when more than 256 objects are retired per epoch under the current design.

I'm excited about this but I think we need to discuss a few more things before proceeding on this, hit me up on Discord when you have time.

panda/src/putil/copyOnWritePointer.cxx
Maxwell175: This is a leftover from before I actually did EBR and was still trying to poke holes using the stress tester. Reverted.

panda/src/putil/copyOnWritePointer.cxx
Maxwell175: Added EpochHolder class and used it across the PR.

panda/src/display/graphicsEngine.cxx
Maxwell175: Added EpochHolder block

panda/src/pgraph/pandaNode.cxx
Maxwell175: Cleaned up

panda/src/pgraph/pandaNode.cxx
Maxwell175: Cut the while loop.

panda/src/pgraph/pandaNode.h
Maxwell175: Done

panda/src/pgraph/pandaNode.h
Maxwell175: Fixed.

panda/src/pgraph/pandaNode.I
Maxwell175: Also a remnant of pre-EBR flailing. Reverted and cannot reproduce any issues.

panda/src/pipeline/cycleDataLockedReader.I
Maxwell175: Added a RAII member.

panda/src/pipeline/epochManager.cxx
Maxwell175: Was a workaround for ExternalThread singleton. For now will assume all threads are bound to a Thread object, as discussed.

panda/src/pipeline/epochManager.cxx
Maxwell175: Fixed

panda/src/pipeline/epochManager.h
Maxwell175: Fixed

panda/src/pipeline/epochManager.h
Maxwell175: I can't put it in the Pipeline header as is because it causes an include cycle. Updated to 4. Let me know if there is a better place to put it.

panda/src/pipeline/epochManager.h
Maxwell175: Totally eliminated _active_cs_count in favor of patomic_thread_fence, similar to the reclaim side..

panda/src/pipeline/pipeline.cxx
Maxwell175: Added RAII guard

panda/src/pipeline/pipeline.cxx
Maxwell175: Updated to do a full reclamation at the frame or task manager boundary, while keeping the 256 number for reclamations that happen in between.

panda/src/pipeline/pipelineCyclerTrueImpl.cxx
Maxwell175: Done

panda/src/pipeline/thread.cxx
Maxwell175: Added an assert.

panda/src/pipeline/thread.I
Maxwell175: Workaround Eliminated.

panda/src/gobj/geomVertexData.cxx
Maxwell175: If I start trying to generalize it, I end up with issues around managing the lock_mutex. Pushing this for all COW objects then has performance implications.

panda/src/pipeline/epochManager.h
Maxwell175: Did some benchmarking and measurements show it made sense to uninline consider_reclaim, but not enter or leave, which caused a ~3ns performance regression.

panda/src/pipeline/epochManager.cxx
Maxwell175: I overcomplicated this. Switched to pdeque and just took the lock right away. Cleaner now.

panda/src/putil/copyOnWritePointer.cxx
Maxwell175: Do you mean the namespace {} ? This one was removed, but there is still one in EpochManager. I do also see one in audioManager.cxx and nullAudioSound.cxx.

tests/scenegraph_bench/scenegraph_bench.cxx
Maxwell175: Switched to TrueClock, but kept chrono for sleeping since I don't want to contaminate the test with panda's Thread::sleep.

PR #1858, opened 2026-06-24
Fix Vulkan build and SSBO memory mapping on macOS/UMA

DarthMDev:
## Issue description
This pull request addresses two separate issues encountered when building and verifying the Vulkan renderer (`p3vulkandisplay`) on macOS:
1. **Unix/macOS Build Failures (`glslang`)**: The `makepanda.py` script expected a legacy Windows-specific static library layout for `glslang` on Unix/macOS targets, preventing compilation with modern Homebrew dynamic libraries.
2. **SSBO Memory Mapping Crash on Unified Memory Systems (UMA)**: On UMA systems (such as Apple Silicon Macs), staging buffers are bypassed (`use_staging_buffer = false`) and the engine directly maps memory buffers. However, `UH_dynamic` shader storage buffers (SSBOs) were being allocated as device-local without `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT`, causing `vkMapMemory` to crash with a segmentation fault during shader tests (e.g., `test_glsl_ssbo[vk-spirv]`).

## Solution description
1. **Dynamic `glslang` Library Resolution**:
   - In `makepanda/makepanda.py`, separated the `GLSLANG` dependency checks by platform target. On Unix/macOS, we now verify only the essential dynamic libraries (`glslang`, `SPIRV`, and `glslang-default-resource-limits`) instead of requiring legacy static libs like `MachineIndependent` and `GenericCodeGen`.
2. **Unified Memory Property Alignment**:
   - In `panda/src/vulkandisplay/vulkanGraphicsStateGuardian.cxx`, added `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT` to the memory property flags when `use_staging_buffer` is false. This ensures any buffer designed to be mapped directly on the host is allocated with the appropriate host-visible flags, resolving the mapping crash on UMA platforms.
   - Tested and verified against the full `tests/display/` test suite, with all 53 Vulkan shader/SSBO unit tests and 77 GLSL type tests passing successfully.

## Checklist

### DIFF
--- a/makepanda/makepanda.py
+++ b/makepanda/makepanda.py
@@ -899,7 +899,10 @@ if (COMPILER=="GCC"):
     SmartPkgEnable("VRPN",      "",          ("vrpn", "quat"), ("vrpn", "quat.h", "vrpn/vrpn_Types.h"))
     SmartPkgEnable("OPUS",      "opusfile",  ("opusfile", "opus", "ogg"), ("ogg/ogg.h", "opus/opusfile.h", "opus"))
     SmartPkgEnable("JPEG",      "",          ("jpeg"), "jpeglib.h")
-    SmartPkgEnable("GLSLANG",   "",          ("glslang", "MachineIndependent", "GenericCodeGen", "SPIRV", "OSDependent", "glslang-default-resource-limits"), "glslang/Public/ShaderLang.h", optional_libs=("OGLCompiler", "HLSL"))
+    if GetTarget() == 'windows':
+        SmartPkgEnable("GLSLANG",   "",          ("glslang", "MachineIndependent", "GenericCodeGen", "SPIRV", "OSDependent", "glslang-default-resource-limits"), "glslang/Public/ShaderLang.h", optional_libs=("OGLCompiler", "HLSL"))
+    else:
+        SmartPkgEnable("GLSLANG",   "",          ("glslang", "SPIRV", "glslang-default-resource-limits"), "glslang/Public/ShaderLang.h", optional_libs=("MachineIndependent", "GenericCodeGen", "OSDependent", "OGLCompiler", "HLSL"))
     SmartPkgEnable("SPIRV-TOOLS", "",        ("SPIRV-Tools-opt", "SPIRV-Tools"), "spirv-tools/optimizer.hpp")
     SmartPkgEnable("SPIRV-CROSS-GLSL", "",   ("spirv-cross-core", "spirv-cross-glsl"), "spirv_cross/spirv_cross.hpp", thirdparty_dir="spirv-cross")
     SmartPkgEnable("SPIRV-CROSS-HLSL", "",   ("spirv-cross-core", "spirv-cross-hlsl"), "spirv_cross/spirv_cross.hpp", thirdparty_dir="spirv-cross")
--- a/panda/src/vulkandisplay/vulkanGraphicsStateGuardian.cxx
+++ b/panda/src/vulkandisplay/vulkanGraphicsStateGuardian.cxx
@@ -2682,11 +2682,11 @@ prepare_shader_buffer(ShaderBuffer *data) {
   if (use_staging_buffer) {
     usage_flags |= VK_BUFFER_USAGE_TRANSFER_DST_BIT | VK_BUFFER_USAGE_TRANSFER_SRC_BIT;
   }
-  else if (data->get_usage_hint() == GeomEnums::UH_dynamic) {
-    usage_flags |= VK_BUFFER_USAGE_TRANSFER_DST_BIT;
-  }
   else {
     mem_flags = (VkMemoryPropertyFlagBits)(mem_flags | VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT);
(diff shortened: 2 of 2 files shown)


DarthMDev:
these failures are pre-existing on the upstream vulkan branch and do not represent regressions introduced by my changes.

rdb:
>   * In `makepanda/makepanda.py`, separated the `GLSLANG` dependency checks by platform target. On Unix/macOS, we now verify only the essential dynamic libraries (`glslang`, `SPIRV`, and `glslang-default-resource-limits`) instead of requiring legacy static libs like `MachineIndependent` and `GenericCodeGen`.

I need you to report the actual issue you were running into with this because the premise is wrong; I have been compiling this on Linux and macOS all the time.

>   * In `panda/src/vulkandisplay/vulkanGraphicsStateGuardian.cxx`, added `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT` to the memory property flags when `use_staging_buffer` is false. This ensures any buffer designed to be mapped directly on the host is allocated with the appropriate host-visible flags, resolving the mapping crash on UMA platforms.

The deeper issue is that the logic is fragile here, but yes, the fix is correct, thank you. Since you mixed this in with an unrelated change I'll just reproduce and commit this one locally on my end since it's trivial. There's also a related bug that the use of the staging buffer should actually be gated on `use_staging_buffer` and not `_has_unified_memory` or the `UH_client` path on a discrete GPU won't follow the correct branch.

DarthMDev:
> > * In `makepanda/makepanda.py`, separated the `GLSLANG` dependency checks by platform target. On Unix/macOS, we now verify only the essential dynamic libraries (`glslang`, `SPIRV`, and `glslang-default-resource-limits`) instead of requiring legacy static libs like `MachineIndependent` and `GenericCodeGen`.
> 
> I need you to report the actual issue you were running into with this because the premise is wrong; I have been compiling this on Linux and macOS all the time.
> 
> > * In `panda/src/vulkandisplay/vulkanGraphicsStateGuardian.cxx`, added `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT` to the memory property flags when `use_staging_buffer` is false. This ensures any buffer designed to be mapped directly on the host is allocated with the appropriate host-visible flags, resolving the mapping crash on UMA platforms.
> 
> The deeper issue is that the logic is fragile here, but yes, the fix is correct, thank you. Since you mixed this in with an unrelated change I'll just reproduce and commit this one locally on my end since it's trivial. There's also a related bug that the use of the staging buffer should actually be gated on `use_staging_buffer` and not `_has_unified_memory` or the `UH_client` path on a discrete GPU won't follow the correct branch.

On the makepanda.py change; you're right that the premise as I wrote it is too broad, sorry about that. The real issue is narrower: I'm building against Homebrew's glslang (16.3.0), which installs only libglslang, libSPIRV, and libglslang-default-resource-limits as shared libraries. In that build, MachineIndependent, GenericCodeGen, and OSDependent are merged into libglslang and don't exist as separate libs. Because makepanda lists them as required, SmartPkgEnable fails to locate them, sets have_pkg = False, and ends up calling PkgDisable("GLSLANG"); so GLSLANG gets excluded and the Vulkan/shaderpipeline build fails.

You don't see this because the thirdparty bundle (and static source builds) still ship those as separate archives; it only bites when linking against a system glslang built as shared libs. So the right fix is just to move those three to optional_libs rather than gating by platform; happy to redo it that way if you'd prefer.

On the Vulkan host-visible fix; thanks, and agreed it got tangled up with the unrelated change, that's on me. Since you're reproducing and committing that one locally, would you mind adding me as a co-author on the commit? Something like:

Co-authored-by: DarthM <mgracer48@yahoo.com>

No worries if it's more hassle than it's worth; just appreciated if it's easy.

rdb:
I grabbed glslang 16.3.0, built it using the [homebrew recipe](https://github.com/Homebrew/homebrew-core/blob/60536ad7cbed28bdbab1fb9908b687be8f2cd7c3/Formula/g/glslang.rb) and found that it *does* build the .a files, but they're not installed by `cmake --install`. Because libglslang is built as a shared library and linked with the .a files, they do include their symbols. This is a bit of a strange setup on their end, because it means the set of libraries to link with is different between a dynamic and a static build, but we need to deal with this.

This shouldn't be gated on whether we're on Windows or not because the Windows build actually never hits this branch. It should ideally be gated on whether glslang is built as a shared library or not, but that's not trivial to reliably determine here. The pragmatic fix is would be to mark the ones that aren't always installed as optional. That fix should go into the shaderpipeline branch and then be merged onto vulkan. I will do that.

> On the Vulkan host-visible fix; thanks, and agreed it got tangled up with the unrelated change, that's on me. Since you're reproducing and committing that one locally, would you mind adding me as a co-author on the commit? 

Sorry, I appreciate you taking the time to find and submit this change, but I can't rewrite commit history to change the commit message of a published commit; force pushes should only be done in exceptional situations since they disrupt everyone's clones. Going forward, please submit unrelated changes in separate pull requests so that authorship remains properly tracked on merges.

DarthMDev:
Sounds good , I'm going to close this for now.

PR #1868, opened 2026-07-02
Update Bullet example code to more accurately reflect mesh contents

TV4Fun:
Many meshes include multiple submeshes, as well as built-in transforms. The existing example code only takes the untransformed vertices of the first submesh, which creates some very confusing and time-consuming behavioral inconsistencies if someone is following an example with a model that is more complex than this. This fixes the example code to be more robust. The examples in the docs under Bullet collision shapes for Convex Hull and Triangle Mesh should also be updated.

## Checklist

### DIFF
--- a/samples/bullet-physics/20_bowl_and_eggs.py
+++ b/samples/bullet-physics/20_bowl_and_eggs.py
@@ -131,13 +131,14 @@ class Game(ShowBase):
 
         # Bowl
         visNP = loader.load_model('models/bowl.egg')
+        # This is necessary to incorporate internal transforms and submeshes into the geometry imported into Bullet. If
+        # you don't want to alter your model's internal structure, you can also follow the BulletConvexHullShape example
+        # below.
+        visNP.flatten_strong()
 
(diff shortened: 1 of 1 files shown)


rdb (changes requested):
This is a nice change in principle, but the getTransform should be taken relative to the root node of the model, currently it just takes it relative to whatever its parent might be, which won't work if the model is a nested subtree with transforms on the various intervening nodes.

TV4Fun:
> This is a nice change in principle, but the getTransform should be taken relative to the root node of the model, currently it just takes it relative to whatever its parent might be, which won't work if the model is a nested subtree with transforms on the various intervening nodes.

Done. I also realized after I wrote this that I could have just flattened the geometry as well. So I have now included examples of doing it both way along with an explanatory comment. Thank you, BTW, for responding to all of my issues so quickly. Side question, is there some magic that `ShowBase` does to put `render`, `base`, `globalClock`, etc. into the global namespace? I would expect to have to access these members through `self`, and PyCharm thinks the global references are errors, but the code still seems to work fine.

PR #1874, opened 2026-07-04
bullet: Expose btConvexHullShape::optimizeConvexHull in BulletConvexH…

### LINKED ISSUE #1866: Expose `btConvexHullShape::optimizeConvexHull` in `BulletConvexHullShape`
## Description
Bullet provides the `btConvexHullShape::optimizeConvexHull` to automatically remove redundant points from a convex hull. I don't know if you do this elsewhere in your code. If you do, I couldn't find it. It would be really nice to expose this method or otherwise provide code to optimize points in a convex hull.

## Use Case
When creating a convex hull programmatically from a provided mesh, there are likely to be many redundant interior points. This can cause strange physics results and poor performance. There should be a way to automatically remove interior points.

Pooja250588:
…ullShape

## Issue description
This change resolves #1866 by exposing the `btConvexHullShape::optimizeConvexHull` method on the `BulletConvexHullShape` class. This allows automatically removing redundant/interior points from a convex hull, improving collision detection performance and physical accuracy when creating a convex hull programmatically from a mesh.
 fixes=#1866
## Solution description
We exposed the following methods/properties on the `BulletConvexHullShape` class:
* `void optimize_convex_hull()`: Calls the underlying `btConvexHullShape::optimizeConvexHull` under global lock protection.
* `int get_num_points() const`: Exposes a way to query the count of vertices currently defining the shape, matching the interface pattern of `BulletConvexPointCloudShape`.
* Python property `num_points` mapped to `get_num_points()`.

We also added a Python unit test `test_convex_hull_optimization` in `tests/bullet/test_bullet_bam.py` that verifies that adding a redundant interior point and then calling `optimize_convex_hull()` successfully reduces the point count by 1.

## Checklist
I have done my best to ensure that...
* [x] ...I have familiarized myself with the CONTRIBUTING.md file
* [x] ...this change follows the coding style and design patterns of the codebase
* [x] ...I own the intellectual property rights to this code
* [x] ...the intent of this change is clearly explained
* [x] ...existing uses of the Panda3D API are not broken
* [x] ...the changed code is adequately covered by the test suite, where possible.

### DIFF
--- a/panda/src/bullet/bulletConvexHullShape.cxx
+++ b/panda/src/bullet/bulletConvexHullShape.cxx
@@ -143,6 +143,26 @@ add_geom(const Geom *geom, const TransformState *ts) {
 }
 
 /**
+ *
+ */
+int BulletConvexHullShape::
+get_num_points() const {
+  LightMutexHolder holder(BulletWorld::get_global_lock());
+
+  return _shape->getNumPoints();
+}
+
+/**
+ *
+ */
+void BulletConvexHullShape::
+optimize_convex_hull() {
+  LightMutexHolder holder(BulletWorld::get_global_lock());
+
+  _shape->optimizeConvexHull();
+}
+
+/**
  * Tells the BamReader how to create objects of type BulletShape.
  */
 void BulletConvexHullShape::
--- a/panda/src/bullet/bulletConvexHullShape.h
+++ b/panda/src/bullet/bulletConvexHullShape.h
@@ -38,6 +38,11 @@ PUBLISHED:
   void add_geom(const Geom *geom,
                 const TransformState *ts=TransformState::make_identity());
 
+  int get_num_points() const;
+  void optimize_convex_hull();
+
+  MAKE_PROPERTY(num_points, get_num_points);
+
 public:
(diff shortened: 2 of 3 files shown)


rdb:
Might as well offer a `points` property (with MAKE_SEQ_PROPERTY) here as well, thoughts?

Pooja250588:
Thanks for the suggestion! That makes sense. I'll add a points property using MAKE_SEQ_PROPERTY and update this PR.

Pooja250588:
> **Done! I have removed the `num_points` property and moved the C++ `get_num_points` and `get_point` methods from `PUBLISHED` to `public` to avoid interface duplication.**
> 
> **Instead, the branch now exposes the single `points` sequence property:**
> `MAKE_SEQ_PROPERTY(points, get_num_points, get_point);`
> 
> **I have also updated the unit tests to test the new sequence property.**

rdb:
I don't see the changes, forgot to push?

rdb:
I see you probably meant to make that comment on #1876, I see new commits there and will review that.

Pooja250588:
please give me some more time.

PR #1876, opened 2026-07-04
bullet: Expose get_num_points and get_point in BulletConvexHullShape

### LINKED ISSUE #1875: Missing API to query points from BulletConvexHullShape
## Description
Currently, `BulletConvexHullShape` allows developers to add points individually using `add_point(p)` or as an array using `add_array(points)`. However, there is no corresponding API to retrieve the points from the shape once it is 

## Use Case
To make the `BulletConvexHullShape` API complete, we should expose a way to query:
1. The total number of points via `get_num_points()` (and its Python property `num_points`).
2. Individual points via `get_point(i)`.

### Solution Proposal
Expose the following methods in the `PUBLISHED` section of `BulletConvexHullShape`:
* `int get_num_points() const;`
* `LPoint3 get_point(int i) const;`

Pooja250588:
## Issue description
This change resolves #1875 by exposing methods to query point counts and individual points from a `BulletConvexHullShape`. Currently, developers can only add points but cannot retrieve them back.
 fixed=#1875
## Solution description
We exposed the following methods/properties on the `BulletConvexHullShape` class:
* `int get_num_points() const`: Returns the total number of points in the shape.
* `LPoint3 get_point(int i) const`: Returns the point at index `i` (includes boundary check `nassertr`).
* Python property `num_points` mapped to `get_num_points()`.

We also added a Python unit test `test_convex_hull_points` in `tests/bullet/test_bullet_bam.py` to verify that added points are retrieved accurately.

## Checklist
I have done my best to ensure that...
* [x] ...I have familiarized myself with the CONTRIBUTING.md file
* [x] ...this change follows the coding style and design patterns of the codebase
* [x] ...I own the intellectual property rights to this code
* [x] ...the intent of this change is clearly explained
* [x] ...existing uses of the Panda3D API are not broken
* [x] ...the changed code is adequately covered by the test suite, where possible.

### DIFF
--- a/panda/src/bullet/bulletConvexHullShape.cxx
+++ b/panda/src/bullet/bulletConvexHullShape.cxx
@@ -143,6 +143,27 @@ add_geom(const Geom *geom, const TransformState *ts) {
 }
 
 /**
+ *
+ */
+int BulletConvexHullShape::
+get_num_points() const {
+  LightMutexHolder holder(BulletWorld::get_global_lock());
+
+  return _shape->getNumPoints();
+}
+
+/**
+ *
+ */
+LPoint3 BulletConvexHullShape::
+get_point(int i) const {
+  LightMutexHolder holder(BulletWorld::get_global_lock());
+
+  nassertr(i >= 0 && i < _shape->getNumPoints(), LPoint3::zero());
+  return btVector3_to_LPoint3(_shape->getUnscaledPoints()[i]);
+}
+
+/**
  * Tells the BamReader how to create objects of type BulletShape.
  */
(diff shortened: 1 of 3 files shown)


Pooja250588:
Hi! The workflows are awaiting approval. Could someone please approve them so the checks can run? Thank you!

rdb (changes requested):
Rather than a num_points property we would rather expose a property like this:

```
MAKE_SEQ_PROPERTY(points, get_num_points, get_point);
```
and then mark the methods as public instead of PUBLISHED so we don't duplicate the interface.

PR #1879, opened 2026-07-22
Fix NameError in TexMemWatcher from missing config-variable imports

uttam12331:
## Issue description

`direct.showutil.TexMemWatcher` imports its config-variable classes explicitly from `panda3d.core`, but the import list only contains `ConfigVariableInt`. `TexMemWatcher.__init__` also uses two other classes that are never imported:

- `ConfigVariableString('tex-mem-pipe', '')`
- `ConfigVariableDouble("tex-mem-update-interval", 0.5)`

Both are on the unconditional `__init__` path, so constructing a `TexMemWatcher` — e.g. via `ShowBase.toggleTexMem()` — raises:

```
NameError: name 'ConfigVariableString' is not defined
```

before the window can open. This makes the built-in texture-memory viewer unusable. The names were dropped from the import list during the `import *` cleanup and haven't been referenced-but-imported since.

## Solution description

Add the two missing names, `ConfigVariableDouble` and `ConfigVariableString`, to the existing `from panda3d.core import (...)` list (keeping it alphabetical). No logic changes.

The sibling `ConfigVariableInt` is already imported and used the same way in this file, confirming these belong in the same import block.

Verified against `master`: with the current code, importing the module and referencing these names fails; loading `panda3d.core` confirms all three classes exist, so the import is valid. Before the fix `ConfigVariableString`/`ConfigVariableDouble` are absent from the module namespace; after it they resolve, so `TexMemWatcher.__init__` no longer raises `NameError`.

## Checklist


rdb (approved):
Good catch, thanks.

PR #1882, opened 2026-07-26
Fix AnimControlInterval.getCurrentFrame using numFrames instead of frameRate

uttam12331:
## Issue description

`AnimControlInterval.getCurrentFrame()` returns the wrong frame number, the same way `ActorInterval.getCurrentFrame` did (#1881). It computes:

```python
framesPlayed = self.numFrames * self.currT
retval = self.startFrame + framesPlayed
```

but `self.currT` is the elapsed time in **seconds** (set in `privStep`), so this multiplies a frame *count* by seconds. The frame the animation is actually posed to is computed in the same class's `privStep`:

```python
def privStep(self, t):
    frameCount = t * self.frameRate
    ...
```

So `getCurrentFrame` should use `frameRate` (frames/second × seconds = frames), not `numFrames`.

For a 48-frame animation at 24 fps (`startFrame=0`, `endFrame=47`, `numFrames=48`, `frameRate=24`):

| elapsed time | frame `privStep` poses | `getCurrentFrame()` returns |
|---|---|---|
| 0.5 s | 12 | **24** |
| 1.0 s | 24 | **48** |
| 1.5 s | 36 | **72** |

The reported frame is roughly double the real one and outside the documented `[startFrame, endFrame]` range for most of any animation longer than a second.

## Solution description

Use `self.frameRate` so `getCurrentFrame` matches the frame `privStep` poses:

```python
framesPlayed = self.frameRate * self.currT
```

This is the identical fix to the sibling `ActorInterval.getCurrentFrame` in #1881.

Added `tests/interval/test_AnimControlInterval.py`, which builds an `AnimControlInterval` from a real `AnimControl` (bound from an `AnimBundle`) and asserts `getCurrentFrame()` equals `startFrame + currT * frameRate`. It fails on the current code and passes with this change.

## Checklist


rdb:
Seems to be an unintentional duplicate of #1881, so I'm closing this in favour of the other one.

